PR #96, opened 2026-03-29
Patch/parse refinement

josephismikhail:
## Summary
Adds YAML and Dockerfile parsing support, fixes Go/Python parser regressions, resolves cross-batch edge resolution in large repos, and restores parse-commit pipeline overlap that was causing per-1000-file pauses during ix map.

## Type
- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Docs
- [x] Test
- [ ] CI

## Changes
-YAML parsing: entities, relationships, config_entry kind for scalar values
-Dockerfile parsing with singleton map regions
-Go parser: resolve aliased import CALLS edges, capture qualified type refs, fix file discovery canonicalization (Chain 4/5 fixes)
-Python parser: fix edge extraction regressions
-ix trace: disambiguate same-named functions with source path
-GlobalResolutionIndex: pre-scan all Go files before streaming parse so cross-batch IMPORTS/REFERENCES edges resolve correctly in large repos
-Perf: restore pendingFlush parse-commit overlap — backend commit now runs concurrently with next chunk parse, eliminating per-1000-file blocking pause
-Fix ix map ingest crash and suppress nested output

## Validation
ix map on repos with YAML and Dockerfile files
ix map on large Go repo (Kubernetes, ~15k files) — cross-batch edges resolve correctly, no per-1000-file pause

## Checklist
- [ ] Tests pass
- [ ] Smoke tests pass
- [ ] No raw errors introduced

### DIFF
--- a/core-ingestion/src/__tests__/resolveEdges.test.ts
+++ b/core-ingestion/src/__tests__/resolveEdges.test.ts
@@ -576,4 +576,53 @@ describe('resolveEdges', () => {
       confidence: 0.9,
     });
   });
+
+  it('Python: from X import ClassName resolves IMPORTS edge to the class node via Tier 2', () => {
+    // `from models import Column` produces two IMPORTS edges: one for the module
+    // (resolves to models.py as a file) and one for the symbol (dstName='Column',
+    // no file match). The PascalCase fallthrough should bind Column to the class node.
+    const consumer = fileResult(
+      '/repo/consumer.py',
+      SupportedLanguages.Python,
+      [entity('use_column', SupportedLanguages.Python)],
+      [
+        { srcName: 'consumer.py', dstName: 'models', predicate: 'IMPORTS' },
+        { srcName: 'consumer.py', dstName: 'Column', predicate: 'IMPORTS' },
+      ],
+    );
+    const models = fileResult(
+      '/repo/models.py',
+      SupportedLanguages.Python,
+      [entity('Column', SupportedLanguages.Python, 'class')],
+    );
+
+    expect(resolveEdges([consumer, models])).toContainEqual({
+      srcFilePath: '/repo/consumer.py',
+      srcName: 'consumer.py',
+      dstFilePath: '/repo/models.py',
+      dstName: 'Column',
+      dstQualifiedKey: 'Column',
+      predicate: 'IMPORTS',
+      confidence: 0.9,
+    });
+  });
+
+  it('Go: IMPORTS edge with PascalCase name and zero importMatches does not fall through to symbol resolution', () => {
+    // An unresolvable external package (no matching file) with a PascalCase name
+    // should not leak into Tier 2/3 and bind to an in-repo Go symbol of the same name.
+    const consumer = fileResult(
+      '/repo/main.go',
+      SupportedLanguages.Go,
+      [entity('main', SupportedLanguages.Go, 'function')],
+      [{ srcName: 'main.go', dstName: 'HttpClient', predicate: 'IMPORTS' }],
+    );
+    const lib = fileResult(
+      '/repo/lib/http.go',
+      SupportedLanguages.Go,
+      [entity('HttpClient', SupportedLanguages.Go, 'class')],
+    );
+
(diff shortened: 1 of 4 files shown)


TannerTorrey3 (changes requested):
## Review: Suggested changes before merge

Full spec with code snippets: `docs/superpowers/specs/2026-03-28-pr96-review-suggestions.md`

---

### 1. `ix trace` JSON — restore compact format, reduce token cost

Single-direction mode changed the `tree` field from `tree.map(compactTreeNode)` to a nested object with `id`/`path`/`children`, and added redundant `nodes`/`edges` arrays — three representations of the same data. The "both" direction still uses the old compact format, creating an inconsistency.

**Requested change:** Revert to `tree.map(compactTreeNode)` and remove `nodes`, `edges`, `collectJson`, and `toJsonTree`. The compact tree already encodes the full structure without UUIDs and is more token-efficient for LLM consumers. This also restores backward compatibility with the previous JSON shape.

`ix-cli/src/cli/commands/trace.ts` lines 472-509 → replace with:
```typescript
if (opts.format === "json") {
  const output: Record<string, unknown> = {
    mode: "directional",
    target: { name: target.name, kind: target.kind, path: relativePath(target.path) },
    direction,
    kind: relKind,
    depth: maxDepth,
    tree: tree.map(compactTreeNode),
    summary: { nodes_visited: nodesVisited, max_depth: maxDepthReached },
  };
```

---

### 2. Python IMPORTS fallthrough — tighten language guard + add tests

The fallthrough at line 1800 uses `!/^[A-Z]/.test(rel.dstName)` which is language-agnostic. If a Go IMPORTS edge has zero `importMatches` (unresolvable external package) and a PascalCase name, it leaks into Tier 2/3 and could bind to the wrong node.

**Requested change (2a):** Add a language check to the guard.

`core-ingestion/src/index.ts` line 1800 — change:
```typescript
// before
if (!/^[A-Z]/.test(rel.dstName)) continue;
// after
if (srcLanguage !== SupportedLanguages.Python || !/^[A-Z]/.test(rel.dstName)) continue;
```

**Requested change (2b):** Add two tests in `resolveEdges.test.ts`:
1. **Happy path** — Python `from X import Column` where `Column` is a class in another file → IMPORTS edge resolves to the class node
2. **Guard** — Go IMPORTS edge with PascalCase name and zero `importMatches` → does NOT fall through to Tier 2/3

Run test 1 first to validate the fallthrough actually works. Test 2 will fail against the current code (proving the guard is too loose), then pass after applying the language guard fix.

---

### 3. Extract `narrowGoImportCandidates` helper

`resolveImportTargets` and `resolveImportQualifierTargets` (lines 1522-1590) share ~90% identical code. The only difference is the `dirGroups.size === 1` branch.

**Requested change:** Extract shared suffix-scoring and dir-grouping into `narrowGoImportCandidates(srcFilePath, modName, pickFromSingleDir)`, parameterized by a callback. Each caller keeps its own early return for non-Go languages and initial `modNameToFiles` call, then delegates to the helper.

See spec for full code.

---

### 4. Memory optimization — `sources.clear()` after GlobalResolutionIndex build

Both Path A (line 906) and Path B (line 951) in `ingest.ts` load all Go source files into a `Map<string, string>`. After `buildGlobalResolutionIndex` returns, the source content is never referenced again — the index only stores file paths and symbol names. The map stays in memory until GC, overlapping with the parse loop.

**Requested change:** Add `sources.clear()` immediately after each `buildGlobalResolutionIndex` call. Zero-risk — the function is synchronous and the map is block-scoped.

---

### Summary

| # | Change | Risk | Action |
|---|--------|------|--------|
| 1 | `ix trace` JSON compact format | Breaking → Restored | Revert to `compactTreeNode`, drop redundant `nodes`/`edges` |
| 2 | Python IMPORTS fallthrough | Untested + Go leak-through | Add language guard + two tests |
| 3 | Extract `narrowGoImportCandidates` | Duplicate code | Refactor to shared helper |
| 4 | `sources.clear()` after index build | Memory optimization | Add two one-liners |

TannerTorrey3 (approved):

PR #161, opened 2026-04-08
fix: patch Vite security advisories (6 Dependabot alerts)

### LINKED ISSUE #4: ix entity
determine, do we actually want a change for every single part, right now, ix is storing to mcuh and with long term use, could be a problem, it is storing multiple claims and having multiple clauims for same time, so most likely just inddual line or name changes, not the once the entire file is change.  This needs to be looked into and only store as new claim when entire file saved or change completed.

### LINKED ISSUE #7: Simplify user experience - release version - packaging - etc
we need release version
update notifications
single command preferred
-curl, brew, whatever windows uses etc, and plugin

josephismikhail:
Adds npm overrides to force patched Vite versions:
- ix-cli: vite >=7.3.2 <8.0.0 (fixes #7, #8, #9 — CVEs in 7.x range)
- core-ingestion: vite >=8.0.5 (fixes #4, #5, #6 — CVEs in 8.x range)

## Summary
Patches 6 Dependabot security advisories (4 high, 2 moderate) affecting Vite, which is pulled in transitively via vitest. Uses npm `overrides` to force patched versions without changing vitest itself.

## Type
- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [x] CI

## Changes
- `ix-cli/package.json`: added `overrides` pinning vite to `>=7.3.2 <8.0.0`
- `core-ingestion/package.json`: added `overrides` pinning vite to `>=8.0.5`
- Both `package-lock.json` files regenerated with patched versions (7.3.2 and 8.0.7)

## Validation
- `npm install` reports `found 0 vulnerabilities` in both packages
- `npm test` in `core-ingestion` passes 129/135 tests; 6 failures are pre-existing and unrelated to this change (missing `memory-layer` fixtures, snapshot drift from recent main changes)

## Checklist
- [x] Tests pass
- [x] Smoke tests pass


TannerTorrey3 (approved):
No breaking changes. Necessary security patches for existing vulnerabilities

PR #191, opened 2026-04-12
fix: dereference symlinks in CLI tarballs

riley0227:
## Summary
- `cp -r` preserves npm workspace symlinks in `node_modules`, which end up in the release tarballs
- macOS `tar` refuses to extract files through symlinks (`Cannot extract through symlink` errors), breaking installs on Mac
- Changed to `cp -rL` across all 5 platform packaging steps to dereference symlinks into real files

## Test plan
- [ ] Merge and cut a new release
- [ ] Download the darwin-arm64 tarball and verify `tar xzf` extracts cleanly
- [ ] Run the install script on a clean Mac


josephismikhail:
Closing PR because this fix already landed on main as part of #193 (6ab8141, sub-commit "fix: dereference symlinks in CLI tarballs"). Same exact change, zero diff against main now. Thanks!

PR #200, opened 2026-04-21
security(cli): tighten ~/.ix perms + back off on /auth/token 429

jbmiv:
## Summary
- `saveConfig` writes `~/.ix/config.yaml` at `0600` and creates `~/.ix` at `0700`, with `chmod` on each save to tighten any pre-existing `0644` file from older clients (API keys + JWTs were world-readable on shared dev hosts otherwise).
- `refreshAuthIfNeeded` keeps the cached token on HTTP `429` from the tunnel's `tokenRateLimit` middleware instead of clearing auth and retrying — cached token is still valid for the `bufferMs` window.
- Sends `Authorization: Bearer <api_key>` on `/auth/token` in addition to the body field, matching the server's preferred `extractBearer` path (`ix-tunnel-server/internal/apikey/handler.go:67`).

## Stacked on #5
This depends on `InstanceConfig` / `InstanceAuth` and the auth helpers (`storeAuth`, `clearAuth`, `getAuthToken`, `refreshAuthIfNeeded`) introduced in #5. **Base is `feat/cloud-auth`, not `main`.** When #5 lands, GitHub will retarget this PR to `main` automatically.

## Test plan
- [ ] `cd ix-cli && npm test -- config-auth` — 11 vitest cases pass
- [ ] perms: first write, dir-create, tighten pre-existing 0644
- [ ] auth round-trip: storeAuth / clearAuth / expiry boundaries
- [ ] 429: cached token returned unchanged, stored auth not overwritten
- [ ] Bearer header: forwarded on `/auth/token` POST


jbmiv:
Rebased onto current `main` and added a fix-up commit. Force-pushed (`a918ee8 → 6e99dbe`).

## What changed

- **Rebased** onto `main` to absorb the 16 commits since divergence (Node 20 guard, view -p, symlink fix, HOME bind-mount removal, Riley's #193 client-agnostic backend, README/install churn). The rebase was conflict-free — our `config.ts` edits live in the imports/types/save-perm/auth-helper region of the file; Riley's `absoluteFromSourceUri` lives mid-file between `getActiveWorkspaceRoot` and `resolveWorkspaceRoot`. Different regions, clean 3-way merge.

- **Added fix-up commit** (`6e99dbe`) that repairs two bugs introduced by #193 that ship on `main` today:

  **Bug 1 (BLOCKER):** `ix-cli/src/cli/config.ts:185` calls `require("node:path").resolve(...)` inside an ESM module. `require` is undefined in ESM scope. Every `ix read` against a v0.6 graph with a relative `source_uri` crashes with `ReferenceError`. Existing test suite did not cover this path. Fixed: import `resolve` properly alongside `join`.

  **Bug 2 (HIGH):** `ix-cli/src/cli/commands/ingest.ts:409-411` derives `workspaceRoot` from the ingest target rather than the configured workspace, so `ix map ./src` and `ix map .` produce different `workspaceId`s and fragment the graph. Fixed: `workspaceRoot = resolveWorkspaceRoot(opts.root)`, reused for `resolvedPath`.

  Both fixes have unit-test coverage (new `src/cli/__tests__/source-uri.test.ts`, 7 tests). Real `ix read` round-trip validated against a running memory-layer (`api.ts:12-335` rendered correctly — would have crashed at module load before the fix).

## Validation

- `vitest run src/cli/__tests__/{config-auth,source-uri}.test.ts` → 22/22 pass
- `tsc --noEmit` → exit 0
- `ix read IxClient` against local memory-layer → source returned, no crash
- One pre-existing failure in `resolver-type-preference.test.ts` confirmed unrelated (same failure on `origin/main`)

## Upstream

Drafted issue text against `ix-infrastructure/Ix` for the two bugs and the process pattern that landed them (PR #193 self-merged 43 minutes after open with 0 reviews / 0 comments; "NEEDS HEAVY REVIEW" banners stripped 2 days before PR opened). Will file separately.

jbmiv:
Closing per PR policy — internal dev branch (james/cli-config-hardening), no cross-team review needed. Branch stays pushed; will continue iteration there. Will reopen / re-PR if/when it needs to retarget main for cross-team review.

Issue #11, opened 2026-03-15
context and what context a user wants to be loaded into their new session and kept from last, place that stores context etc


riley0227:
- so there are things that need to be consdiered
- context and what context a user wants to be loaded into their new session and kept from last, place that stores context etc

PR #204, opened 2026-04-30
Feat/new parser r

BrandonGarate177:
## R language support

Adds first-class R support to Ix ingestion: an optional `tree-sitter-r` grammar,
R-specific queries for definitions/calls/imports, R-aware cross-file resolution,
and external-package stubs for `pkg::func` calls. Validated end-to-end against
`r-lib/lintr`.

### core-ingestion

- Added `R` to `SupportedLanguages`, mapped the `.r` extension.
- Loaded `@davisvaughan/tree-sitter-r` via `tryLoadGrammar` (optional — graceful
  fallback if the grammar isn't installed).
- `R_QUERIES` covering: function definitions (`<-` and `=`, identifier **and**
  string LHS for S3/R5 method names), direct calls, namespace-qualified calls
  (`pkg::func` / `pkg:::func`), `$` method calls, and imports (`library()` /
  `require()` quoted + unquoted, `source()`).
- `R_BUILTINS` to suppress base-R noise from CALLS edges.
- **Cross-file global resolution index is parser-derived** — it reads the
  parser's `definition.function` entities rather than regex-scanning source, so it
  captures every definition form the parser does (`<-`, `=`, string-keyed S3 like
  `"print.myClass"`) and can't drift from the in-batch resolution path. (Same
  approach as the SAS branch.)
- **`pkg::func` namespace calls keep their `::`** through to the relationship
  `dstName`, so the patch builder externalizes genuine package calls under
  `external://<pkg>` — splitting on the `::` so multi-dot members (`utils::write.csv`)
  resolve correctly — while plain dotted names (base-R `is.null`, Go `external.Func`)
  are left untouched. `ix trace` / `ix callees` / `ix impact` can follow these edges.
- **Backtick identifiers preserved** on definition names: quote-stripping is scoped
  to string-keyed S3 names only, so `` `my fn` `` keeps its backticks and still
  matches its call sites.

### ix-cli

- Added `.r` to supported extensions.
- Generalized the Go-only global-index pre-read into a shared
  `INDEX_PRESCAN_EXTENSIONS` set (`{.go, .r}`) + `needsIndexPrescan()` helper —
  mirrors the SAS branch so the two compose cleanly on main.

### Tests

- `queries.r.test.ts` — 14 cases: language detection, function definitions
  (`<-` / `=` / string-keyed S3 / nested), backtick-name preservation, CALLS
  (direct + `pkg::`), IMPORTS (`library()`/`require()` quoted + unquoted, `source()`).
- `patchBuilder.resolution.test.ts` — `::`-gated external stubs, multi-dot split
  (`utils::write.csv`), and no-externalize for plain dotted names (R `is.null`, Go `fmt.Println`).
- `resolveEdges.test.ts` — R cross-batch index captures `<-`, `=`, and string-keyed defs.

### Validation

Ingested `r-lib/lintr` end-to-end (functions, direct + `pkg::` calls,
`library()`/`source()` imports, nested / `=` / string-keyed S3 defs all captured).
Screenshots below.

### Review follow-ups addressed

- External-stub dot-split + missing language gate (the `external://is` / `utils.write`
  noise) — fixed.
- Backtick-stripping regression (was breaking Scala/R backtick-named resolution) — fixed.
- Cross-batch index coverage (`= function`, string-keyed S3) + the "regex too loose"
  note — both resolved by going parser-derived.
- **Deferred:** `obj$method` receiver capture — needs a resolver change (bare-name
  fallback for `$` qualifiers), tracked as its own follow-up.

Full `core-ingestion` suite is back to the 6 pre-existing Scala/Rust/TS failures —
no R regressions.

---


josephismikhail:
Tested end-to-end against the paired memory-layer PR #11 with `r-lib/lintr` ingested (343 R files, 481 functions, 3908 CALLS edges). Most of the PR works as advertised: the 13 unit tests pass, the `inventory --path` fix correctly resolves `unnecessary_placeholder_linter` (the canary in your description), S3 string-keyed methods like `print.lint` and `format.lints` extract cleanly, and no regression on Java ingest when guava is added on top.

One real finding worth a follow-up though: namespace-qualified R calls (`pkg::func`) appear to create orphan CALLS edges that point to phantom node IDs.

### Repro

Pick any lintr function with namespace calls, e.g. `get_source_expressions` (calls `xml2::xml_missing()` 2x, `xml2::xml_add_parent`, `xml2::xml_new_root`, `utils::getParseData` etc).

```
curl -sX POST localhost:8090/v1/list -H 'Content-Type: application/json' \
  -d '{"kind":"function","limit":1000,"scope":"get_source_expressions"}' \
  | jq -r '.[0].id'
# 31000f8b-3e4e-9cb0-1b35-4ba2eec1cd34

curl -sX POST localhost:8090/v1/expand -H 'Content-Type: application/json' \
  -d '{"nodeId":"31000f8b-3e4e-9cb0-1b35-4ba2eec1cd34","direction":"out","predicates":["CALLS"],"hops":1}' \
  | jq '{edges: (.edges|length), nodes: (.nodes|length)}'
# {"edges": 20, "nodes": 15}
```

20 CALLS edges, 15 resolved nodes. The 5 missing nodes are the namespace-qualified calls. Their dst IDs (e.g. `c3adf394-6655-2c50-19e7-94098535561f`) are referenced by edges but:

- `GET /v1/entity/<id>` returns 404
- `POST /v1/search {"term":"xml_missing"}` returns 0 results
- `POST /v1/list {"kind":"function","scope":"xml2::"}` returns 0 results
- `ix trace` stops at the phantom

Only 1 function node across the whole lintr ingest has `::` in its name (`%:::%`, the meta-operator), so this isn't "they exist under a qualified name." They aren't materialized as nodes at all.

### Root cause hypothesis

Looking at the R query block:

```
; Package-qualified calls: pkg::func(args) and pkg:::func(args)
(call
  function: (namespace_operator
    lhs: (identifier) @_qualifier
    rhs: (identifier) @call.name)) @call
```

The `@_qualifier` is captured but the downstream parser logic may not be combining it with `@call.name` into a single resolved identifier the way the Go query handles `pkg.NewManager`. So we end up creating edges to `xml_missing` (just the rhs), no internal node matches it, no external/unresolved node gets materialized either, and we get orphan dst IDs.

### Impact

`ix trace` cannot follow these calls. `ix callees` undercounts external dependencies. `ix impact` misses cross-package blast radius. Subsystem mapping cannot cluster around external-package usage. None of these matter for R packages that are mostly self-contained, but lintr (and any R package depending on xml2 / dplyr / etc) loses cross-package signal.

### Not a blocker

The 95% case (same-language internal calls) works fine. This is worth a follow-up, not a hold on the PR. Possible fixes:

1. Combine `@_qualifier` + `@call.name` into a single name like `xml2::xml_missing` before resolution, so the edges land on namespace-qualified function nodes that get created on demand.
2. If the downstream parser already does this for Go, port the same path to R.
3. Drop the edge entirely when the call is namespace-qualified and unresolvable, instead of creating an orphan.

Otherwise LGTM. The inventory --path fix in particular is a real quality-of-life win and applies cleanly to all languages, not just R.

Also one cosmetic thing: commit message `OOps was meant to delete this` would benefit from being rewritten or squashed before merge.

BrandonGarate177:
### Follow-up for @josephismikhail 

Addressed the namespace-qualified call issue flagged in review. Two changes on top of the original commits:

**`Materialise external stubs for unresolved CALLS`** — instead of emitting a CALLS edge to a phantom node ID when `pkg::func()` can't be cross-file resolved, we now materialise a real `UpsertNode` under `external://pkg` (`kind: function`, `attrs: { package, external: true }`). `ix trace`, `ix callees`, and `ix impact` can now follow these edges. Verified against lintr: `get_source_expressions` went from 5 phantom edges → 0, `xml2::xml_missing` / `xml2::xml_add_parent` / `xml2::xml_new_root` all resolve to live stub nodes.

**`R: expand R_BUILTINS`** — added dotted-name base R functions (`on.exit`, `sys.call`, `proc.time`, `file.path`, `Sys.time`, etc.) and common unqualified builtins (`attr`, `nzchar`, `isFALSE`, `srcfile`, `tryCatch`, etc.) that were producing same-file phantom edges. These are now suppressed at parse time.

Also squashed the `OOps was meant to delete this` commit per review note.

One known residual: unqualified calls to package-imported functions (e.g. `xml_find_all` used without `xml2::` via `importFrom`) still produce same-file phantom edges. Fixing that requires NAMESPACE file parsing — tracked separately.

core-ingestion/src/patch-builder.ts
```diff
@@ -210,18 +215,43 @@ export function buildPatch(
...
+  const seenExternalNodes = new Set<string>();
   for (const r of relationships) {
     // For CONTAINS edges, srcName is the container of dstName — use that to disambiguate.
     const srcKey = resolveKey(r.srcName);
     const dstKey = r.predicate === 'CONTAINS'
       ? resolveKey(r.dstName, r.srcName)
       : resolveKey(r.dstName);
 
+    let dstNid: string;
+    if (r.predicate === 'CALLS' && dstKey.includes('.') && !allQKeys.has(dstKey)) {
```
josephismikhail: **Bug: R's dot-in-name convention breaks the `pkg.func` split, and this branch fires for all languages.**

This externalizes any CALLS target containing a `.`, splitting on the *last* dot into `pkg`/`func`. R uses dots pervasively in ordinary names, so it fabricates bogus packages. Ingesting `r-lib/lintr` produced these among the 146 `external://` stubs:

`external://is` (x8), `external://as` (x3), `external://Sys` (x3), `external://find`, plus garbage like `external://as.data.frame`, `external://utils.write` (from `utils::write.csv`), `external://data.table.as.data` (from `data.table::as.data.frame`).

Two root causes:
1. A dotted `dstKey` can be a `pkg::func` namespace call, a base-R name not in `R_BUILTINS` (`is.null`, `Sys.glob`, `find.package`), or a multi-dot namespaced call (`utils::write.csv` -> key `utils.write.csv` -> last-dot split gives the wrong pkg *and* func). `lastIndexOf('.')` cannot tell these apart.
2. Locally-defined S3 methods (`print.myClass <- function(...)`) are externalized when called **cross-file**, because `allQKeys` only holds *this* file's keys. S3 dispatch is idiomatic R, so this is a primary failure mode, not an edge case.

Separately, there is no language gate here, so the same branch fires for Go/Python/TS. Pre-PR, these unresolved dotted CALLS were dropped (`continue`), so this PR newly materializes external nodes and edges for every language, changing graph output for existing repos with no determinism/stats baseline update.

Suggested fix: carry the `namespace_operator` (`::`) signal through the relationship and only externalize when the qualifier was a real `::` namespace (and/or consult `symbolToFiles`/`fileHasSymbol` for the full dotted name before externalizing). The `buildPatchWithResolution` copy at L449 needs the same fix.

core-ingestion/src/patch-builder.ts
```diff
@@ -400,18 +432,39 @@ export function buildPatchWithResolution(
...
 
-    // For cross-file resolved edges (CALLS, EXTENDS), use the defining file's nodeId
     let dstNodeId: string;
     const resolutionKey = `${r.srcName}:${r.predicate}:${r.dstName}`;
+
     if (edgeResolution.has(resolutionKey)) {
+      // Cross-file resolved — use the defining file's nodeId.
       const { dstFilePath, dstQualifiedKey } = edgeResolution.get(resolutionKey)!;
       dstNodeId = nodeId(dstFilePath, dstQualifiedKey);
+    } else if (r.predicate === 'CALLS' && dstKey.includes('.') && !allQKeys2.has(dstKey)) {
```
josephismikhail: Second copy of the external-stub branch (the `buildPatchWithResolution` path used by `ix map`/`ix ingest`; the L227 copy is used by `ix watch`). Same two issues as flagged on L227: the dotted-name mis-split for R, and the missing language gate that materializes external nodes/edges for Go/Python/TS too. Both copies should get the fix together so the two ingestion paths stay consistent.

josephismikhail (commented):
Tested this branch end-to-end: built the memory-layer jar plus a `node:22` parser harness in Docker, ran the unit suite, and ingested `r-lib/lintr` (the repo in the description) into a local backend.

**Works as intended:** all 13 R query tests pass. The full `core-ingestion` suite is 150/156, and the 6 failures are identical on an `origin/main` baseline (pre-existing Scala/Rust/TS snapshot drift, unrelated to R), so **no regressions from this PR**. lintr ingested cleanly: 373 R files -> 3010 nodes / 6948 edges (CALLS 3212, IMPORTS 23). Function defs, direct and `pkg::` calls, `library()`/`source()` imports, nested/`=`/string-keyed S3 defs were all captured.

A few correctness issues below, in rough priority order. None break ingestion; they degrade graph quality (mostly for R) and change graph output for non-R repos. Happy to send a fix PR.

core-ingestion/src/index.ts
```diff
@@ -1315,8 +1356,10 @@ export function parseFile(filePath: string, source: string): FileParseResult | n
 
       if (defCapture) {
         const kind = DEFINITION_KIND_MAP[defCapture.name] ?? 'function';
-        const name = nameCapture?.node.text
+        const rawName = nameCapture?.node.text
           ?? (defCapture.name === 'definition.constructor' ? 'init' : '');
+        // Strip surrounding quotes for string-keyed definitions (e.g. R S3 method names)
+        const name = rawName.replace(/^(['"`])(.*)\1$/, '$2');
```
josephismikhail: **Regression: this strips backticks from Scala (and R) definition names, but only on the def path.**

The strip runs for every language, not just the R string-keyed S3 case the comment describes. Backtick identifiers are part of the token text in both grammars (verified against tree-sitter-scala and @davisvaughan/tree-sitter-r): Scala `` `type` ``, `` `my val` ``, and R `` `my fn` <- function() {} ``. The call and heritage paths do **not** strip, so a def stored as `type`/`my fn` no longer matches a call captured as `` `type` ``/`` `my fn` ``. Result: CALLS (and EXTENDS/IMPLEMENTS for backtick-named Scala types) silently fail to resolve.

Suggested fix: only strip when the captured `@name` node is a `string` (the actual S3 motivation), leaving backtick identifiers intact:

```ts
const name = nameCapture?.node.type === 'string'
  ? rawName.replace(/^(['"]).*\1$/, m => m.slice(1, -1))
  : rawName;
```

core-ingestion/src/index.ts
```diff
@@ -1921,6 +1964,25 @@ export function buildGlobalResolutionIndex(
       }
     }
 
+    // R: scan function definitions (name <- function(...))
+    // Enables cross-batch Tier-3 resolution for large R repos (>500 files).
+    const rFuncRe = /^([a-zA-Z_.][a-zA-Z0-9_.]*)\s*<-\s*function\b\s*\(/gm;
```
josephismikhail: **Coverage gap: this only matches `<- function(`, but `R_QUERIES` also captures `= function` and string-keyed S3 defs.**

`buildGlobalResolutionIndex` is the only cross-batch source of definitions (the parse-based `fileQKeys` covers just the current batch). This regex excludes `foo = function()` (the `=` operator) and `"print.myClass" <- function()` (string LHS, blocked by the `[a-zA-Z_.]` leading class), yet both are parsed as real entities. So on repos large enough to batch (the >500-file regime this block was added for), CALLS to those defs cannot resolve across batches, and string-keyed S3 methods are exactly the common case.

Suggested fix: mirror the query coverage (both operators + quoted LHS), e.g.

```js
/^\s*(?:"([^"]+)"|`([^`]+)`|([a-zA-Z_.][a-zA-Z0-9_.]*))\s*(?:<-|=)\s*function\b/gm
```

taking whichever group matched and stripping quotes. Guard the bare `=` against `==`/`<=`/`>=`/`!=`.

core-ingestion/src/queries.ts
```diff
@@ -1382,6 +1382,55 @@ export const SCALA_QUERIES = `
...
+
+; Package-qualified calls: pkg::func(args) and pkg:::func(args)
+(call
+  function: (namespace_operator
+    lhs: (identifier) @_qualifier
+    rhs: (identifier) @call.name)) @call
+
+; Method calls: obj$method(args)
+(call
+  function: (extract_operator
```
josephismikhail, line 1585: **Precision gap: `obj$method` captures only the bare method name, no receiver.**

This binds `rhs: (identifier) @call.name` and discards the `lhs` object, so `a$run(...)` and `b$run(...)` both emit `dstName: run`. With no qualifier they resolve name-only, and on any repo with multiple same-named `$` methods they hit the ambiguity guard and get dropped. Unlike the Python/Go method-call patterns elsewhere here, there's no qualifier path at all.

Suggested fix: add a qualifier-capturing variant alongside the existing bare one (the `extract_operator` node exposes `lhs`), so bare-name fallback still fires but same-file tie-breaking can work:

```
(call
  function: (extract_operator
    lhs: (identifier) @_qualifier
    rhs: (identifier) @call.name)) @call
```

core-ingestion/src/patch-builder.ts
BrandonGarate177: Nice catch, this was the big one. The root issue was that R `pkg::func` got flattened to `pkg.func` before it ever reached the patch builder, so real package calls looked exactly like base-R dotted names. That's where `external://is`, `external://as`, and the `utils.write` mis-split were all coming from. 

Fixed in `6324ae5`: we now carry the `::` through instead of collapsing it to a `.`, only externalize when that `::` is actually there, and split on it — so `utils::write.csv` lands as `utils` / `write.csv`.

core-ingestion/src/patch-builder.ts
BrandonGarate177: Same fix in `6324ae5` for this copy (the `ix map`/`ix ingest` path) so both ingestion paths behave identically — `::`-gated, split on the namespace, plain dotted names untouched.

core-ingestion/src/index.ts
BrandonGarate177: Yep, fixed in `6324ae5`. The strip only runs when the `@name` node is actually a `string` now, so backtick identifiers 
(Scala `` `type` ``, R `` `my fn` ``) 
keep their backticks and line up with their call captures again. Added a test for the R backtick case.

core-ingestion/src/index.ts
BrandonGarate177: The cross-batch index now just reads the parser's `definition.function` entities, so it picks up exactly what the in-batch path does: `foo = function()`, string-keyed S3 like `"print.myClass"`, all of it.

core-ingestion/src/queries.ts
BrandonGarate177: I'm going to punt this one to a follow-up rather than rush it in here. Capturing the receiver is the easy half — the catch is our resolver treats any dotted name as "qualified, don't fall back to bare" (`index.ts:~2600`), so turning `a$run()` into `a.run` would actually *hurt* resolution for the common single-`run` case unless I also add a bare-name fallback for `$` qualifiers. That's a resolver change that touches every language, so I'd rather do it on purpose in its own PR.

BrandonGarate177:
- **External-stub dot-split + gate**  — `pkg::func` was getting flattened to `pkg.func`, so real package calls looked like base-R dotted names (`external://is`/`external://as`, the `utils.write` mis-split). Now the `::` carries through, only genuine `::` calls externalize (split on `::` → `utils::write.csv` right), everything else falls through to main untouched.
- **Backtick regression** — quote-strip scoped to string-keyed S3 names only; backtick identifiers keep their backticks.
- **Cross-batch coverage** — `<- function` regex → parser-derived index, so `= function` + string-keyed S3 get indexed ("too loose" is moot). Mirrors the SAS branch.
- **Merge** `efba503` — resolved `inventory --path` toward the client-side filter (no backend dep) + aligned the `ingest.ts` prescan with SAS so they compose on main.
- **Deferred:** `obj$method` receiver — needs a resolver change (bare-name fallback for `$` qualifiers), so it's its own PR.


---
@josephismikhail This one as well should be polished up

josephismikhail (approved):
Verified all 4 addressed review items in code (commit 6324ae5) and empirically on r-lib/lintr (343 files, node:22): external:// noise eliminated (0 bogus stubs; all 30 packages are real CRAN pkgs; utils::write.csv splits correctly on ::), backtick identifiers preserved, S3/= cross-batch resolution works (parser-derived index), deterministic, 1142 CALLS resolved with only 4 ambiguous. Confirmed the ::-gated external branch is effectively R-only (no other language emits :: into call dstNames). Bug #5 (obj$method receiver) reasonably deferred. No regressions (the 4 failing test files are pre-existing on the merge-base). Merged current main and resolved the language-registry conflicts.

Issue #12, opened 2026-03-15
Determine what we should actually have using ix watch - right now proboboly does to much


riley0227:
Not sure if we should be storing ever single thing that an llm is doing, that may be to excessive for people working in large repos.  so maybe to based off push or saves, etc instead of every thing that the llm changes.

PR #208, opened 2026-05-14
feat/sas-parser

BrandonGarate177:
## Summary

Adds first-class SAS language support to Ix ingestion. This PR ships a new `tree-sitter-sas` grammar (published to npm as a standalone package), wires it into `core-ingestion`, and validates it against real-world SAS repositories on GitHub.

SAS programs don't use functions — they use `%macro` definitions. The graph reflects this: macros are first-class entities, DATA/PROC steps are captured as modules, and `%macro` calls produce `CALLS` edges across files.

## Type
- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Docs
- [x] Test
- [ ] CI

## Changes

**New grammar package**
- Published `tree-sitter-sas` to npm: https://github.com/ix-infrastructure/tree-sitter-sas / https://www.npmjs.com/package/tree-sitter-sas
- Grammar pinned at **v0.4.2**. Notable fixes across the 0.3.x → 0.4.x line:
  - `macro_options` rule: `%macro foo(x) / secure des='...'` now parses correctly
  - `&var.suffix` macro references consumed as one token (was splitting at the dot)
  - `%"` inside double-quoted strings consumed opaquely; removed the `%'` single-quote alternative (`''` is SAS's only escape)
  - `fileref_source` promoted to a first-class node (enables `%include FILEREF(member.sas)` capture)
  - parenthesized macro-expression bodies, `%DO/%END` inside expression generators, and macro bodies ending immediately before `%MEND` without a semicolon

**`core-ingestion` changes**
- `languages.ts` — `SAS` enum value + `.sas` extension registration
- `queries.ts` — `SAS_QUERIES`: macro definitions, macro call statements, inline macro calls, DATA step modules, PROC step modules, PROC SQL CREATE/FROM/JOIN, DATA step SET/MERGE/UPDATE inputs, `%INCLUDE` imports (string-literal **and** fileref forms), `LIBNAME` imports. Dataset captures take the whole `(dataset_name)` node so two-part names (`work.foo`) stay one entity
- `index.ts` — dynamic ESM `import()` for `tree-sitter-sas` (top-level `await` grammar); load failures other than missing-module surface to stderr; SAS module normalization (drops `_null_`/`_data_`/`_last_` sentinels and unexpanded `&macro` names; PROC steps re-keyed on `out=` output dataset); single-backslash Windows path handling in import normalization; cross-batch macro index derived from parser `definition.macro` entities; SAS-only case-insensitive `.sas` extension stripping in `resolveEdges`
- `package.json` — `tree-sitter-sas@^0.4.2` added as an optional dependency (`tree-sitter` remains at `^0.21.0`)

**`ix-cli` changes**
- `ingest.ts` — index-prescan generalized from Go-only to a `{.go, .sas}` set, so SAS files are read ahead of the streaming parse loop for cross-batch resolution

**Tests**
- `src/__tests__/queries.sas.test.ts` — 32 vitest unit tests covering language detection; macro/DATA/PROC entity capture; `CALLS`/`IMPORTS` extraction; nested macros; inline block comments in parameter lists; macro options clause; two-part dataset names; sentinel/macro-var filtering; PROC `out=` re-keying; backslash-path dropping; comment-between-`%macro`-and-name cross-batch indexing; string edge cases

## Review follow-up (graph quality)

Addressed the graph-quality findings from review — all additive on all-new SAS code, no regressions:
- Whole-`dataset_name` capture eliminates double-emitted two-part-name and `.sas`-filename module nodes
- `_null_`/`_data_`/`_last_` sentinel DATA steps and unexpanded `&macro` dataset names no longer materialize as nodes/edges
- PROC steps keyed on their output dataset rather than the (too-generic) procedure name; output-less PROC steps dropped
- Single-backslash Windows `LIBNAME` paths normalized; `&`-bearing import targets dropped
- Cross-batch macro index derived from the parser instead of a regex (no drift from in-batch extraction)
- Case-insensitive extension stripping scoped to `.sas` only (no cross-language behavior change)

## Validation

Tested `ix ingest` + `ix map` + `ix inventory` against open-source SAS repositories (AHRQ Quality Indicators, SAS Packages Framework, friendly/SAS-macros). Public SAS repos are sparse and mostly macro collections; these are the best available public proxies for internal enterprise SAS projects.

## Checklist
- [x] Tests pass (32/32 SAS; no new failures in the full core-ingestion suite vs main)
- [x] Smoke tests pass (real-world repos ingested cleanly)


josephismikhail:
Tested end-to-end against the paired memory-layer PR #12 with `AHRQ/AHRQ-Quality-Indicators` ingested (43 .sas files). The 20 unit tests pass, the new `macro` node kind works, DATA/PROC steps map to modules, and `tree-sitter-sas` installs cleanly via the optional dep path. Memory-layer PR #12 will hit a one-line conflict with PR #11 (both add to `SourceExtensions`) but that's trivial to resolve.

Found two real bugs worth fixing before merge though. Both are quantifiable on the AHRQ corpus and both produce phantom edges that inflate metrics without giving traceable targets.

### Bug 1: `%include FILEREF(name)` syntax silently dropped (82 percent miss)

Ground truth on AHRQ: 79 `%include` statements, **all** using fileref-identifier syntax:

```sas
%include PROGRMS(IQI_HOSP_FORMATS.SAS);
%include MacLib(MHI_MEASURES_macro.sas);
```

Graph: 14 IMPORTS edges total, **zero** of them from `%include`. The 14 are all LIBNAME-derived. Verified by counting via `grep -ri '^\s*%include'` (79) and string-literal includes (0) across the repo.

Root cause in `queries.ts`:

```
(include_statement
  source: (string_literal) @import.source) @import
```

The query only matches a string_literal source. SAS production code rarely uses string-literal `%include`; the fileref-identifier form is the dominant pattern. Result: most cross-file include relationships in real SAS code are invisible.

Fix sketch: add a second pattern matching the fileref-identifier form. Whatever tree-sitter-sas names that node type, point a second `@import` capture at it.

### Bug 2: Cross-file macro resolution misses indented `%MACRO` definitions (78 percent miss)

Ground truth on AHRQ: 230 `%MACRO` definitions across the corpus. Only 50 are at column 0; the other 180 are indented:

```sas
 %MACRO MDX(FMT);     <-- leading space
 %MACRO MDX1(FMT);
 %MACRO MDX2(FMT);
```

Root cause in `index.ts`:

```js
const sasMacroRe = /^%macro\s+([a-zA-Z_][a-zA-Z0-9_]*)\s*[(/;]/gim;
```

The `^` anchor combined with `/m` flag requires line-start, so indented `%MACRO` lines never match. The cross-file resolution index sees only 50 of 230 macros; calls to the other 180 become phantom edges with no traceable target.

Concrete repro: `IQI/Programs/IQI_HOSP_MEASURES.SAS` calls `%MDX`, `%MDX1`, `%MDX2`, `%MDX2Q2`, `%MDXAQ2`, `%MPR` (all defined in `IQI/Macros/IQI_HOSP_Dx_Pr_Macros_v2025.sas` with leading whitespace). The graph shows 14 outgoing CALLS edges from `IQI_HOSP_MEASURES.SAS`, only 3 resolved (`check_mdc`, `check_pay1_race`, `label_qis`). The 11 phantom edges include the MDX family that should have resolved.

Fix: change `/^%macro` to `/^\s*%macro`. One character.

### Smaller observation

The graph has 190 macro nodes but the source has 230 `%MACRO` definitions (17 percent gap). Less critical than the two bugs above and root cause is unclear (could be name deduplication, could be specific files failing to parse). Worth a quick check but not a blocker.

### Things that work well

`%macro` definitions get their own first-class `macro` node kind, which is the right call for SAS where macros aren't functions. The data_step query capturing both `identifier` and `macro_variable_ref` as dataset names is a thoughtful touch that preserves macro-driven name resolution. The SAS_BUILTINS list reasonably filters PROC/control-flow noise from CALLS edges. The dynamic ESM import for the grammar (avoiding the CJS `tryLoadGrammar` path) is the correct approach for grammars that use top-level await.

Memory-layer PR #12 LGTM as a one-liner; just be aware of the PR #11 conflict.

### Comparison note

Same overall shape as PR #204 (R): both PRs do language registration plus queries plus builtins plus a cross-file regex. Both have the same architectural issue, phantom edges from incomplete cross-file resolution, but with different root causes (R loses namespace-qualified calls, SAS loses indented macros). Both are sub-day fixes; both should land before merge.

Happy to pair on any of these. The one-character regex fix in bug 2 in particular is essentially free.

BrandonGarate177:
# Follow-up for @josephismikhail 

Addressed both blocking bugs flagged in review. Three additional changes on top of the original commits:

**`Fix: %INCLUDE fileref syntax not captured (Bug #1)`** — the original query only matched string-literal sources (`%include 'path/to/file.sas'`), missing the fileref-identifier forms dominant in production SAS code. Added a second tree-sitter pattern for `(fileref_source)` and extended `normalizeCapturedImport` to extract the member name from `FILEREF(member.sas)` syntax. Regression tests added for `FILEREF(member)`, bare `FILEREF`, and lowercase-extension variants.

**`Fix: indented %MACRO definitions missed in cross-file resolution (Bug #2)`** — the regex used to build the global symbol index required macros to start at column 0 (`/^%macro/`), silently dropping the ~78% of production definitions that are indented. Changed to `/^\s*%macro/` (one character). Regression tests confirm the old pattern misses indented definitions and the fixed one catches them.

**`SAS queries: add PROC SQL and DATA step data-flow patterns`** — added tree-sitter queries for `sql_create_statement` (CREATE TABLE AS output), `sql_select_statement` FROM/JOIN table references, and DATA step SET/MERGE/UPDATE inputs. These were present in the v0.3.x grammar but not yet wired into `SAS_QUERIES`.

**`Bump to tree-sitter-sas@0.4.2`** — v0.4.1 added `fileref_source` as a first-class node (required for Bug #1 fix). v0.4.2 resolved three grammar-level failures: parenthesized macro-expression bodies without trailing semicolons, `%DO/%END` inside parenthesized expression generators, and macro bodies ending immediately before `%MEND` without a semicolon. Tests for all three cases added.

**`.sas` added to extension strip regex in `resolveEdges`** — `IQI_HOSP_FORMATS.SAS` was being looked up verbatim instead of also trying the stem `IQI_HOSP_FORMATS`, causing IMPORTS edges from fileref includes to go unresolved. Now case-insensitive and includes `.sas`.

josephismikhail (commented):
Built this branch in Docker (a node:22 harness + the tree-sitter-sas@0.4.2 grammar), ran the unit suite, and ingested AHRQ/AHRQ-Quality-Indicators (79 .sas files) into a local backend.

**Solid:** 28/28 SAS tests pass; the full core-ingestion suite has no new failures vs origin/main (the 6 failures are pre-existing Scala/Rust/TS snapshot drift); the ingest is clean end to end (1613 nodes, 2975 edges, exit 0). The top-level-await grammar load works through the real parse-worker pool, not just vitest. The grammar itself parses correctly everywhere I tested, including the v0.3.5 to v0.4.2 edge cases.

The findings below are graph-quality, not correctness: none are regressions (SAS is all-new code) and nothing crashes. Empirically ~30% of the 571 module nodes from AHRQ are noise (69 are .sas filenames, 23 are _null_ sentinels, 85 contain unexpanded &macro vars, 7 are Windows path fragments), and all of it traces back to the dataset_name captures plus missing normalization in core-ingestion, not to the grammar. The single highest-leverage fix (capture the whole dataset_name node + filter sentinels/macro-vars) removes most of it. Happy to send a fix PR.

FYI: the PR description is stale. It mentions a tree-sitter 0.25 bump and grammar 0.3.6, but this head keeps tree-sitter at 0.21 and uses tree-sitter-sas 0.4.2.

core-ingestion/src/queries.ts
```diff
@@ -1382,6 +1382,79 @@ export const SCALA_QUERIES = `
...
+
+; Inline macro calls — %name(args) inside DATA/PROC/etc.
+(macro_call
+  name: (macro_name) @call.name) @call
+
+; DATA step — captures the dataset identifier or macro variable reference
+(data_step
+  (data_step_header
+    (dataset_name
+      [(identifier) (macro_variable_ref)] @name))) @definition.module
```
josephismikhail: **Biggest source of module-node noise: `[(identifier) (macro_variable_ref)] @name` is unanchored under `(dataset_name)`, so it matches once per child, and the names are never normalized.** From the AHRQ ingest (571 module nodes total):

- **Two-part names double-capture.** `dataset_name` for `work.foo` (or a bare `IQI_HOSP_FORMATS.SAS`, which lexes as `ident . ident`) has two `identifier` children, so this emits TWO `module` entities (`work` + `foo`, or qualifier + `SAS`). This is the root cause of the **69 of 571 module nodes that are `.sas` filenames**.
- **`_null_`/`_data_`/`_last_` sentinels** become modules (**23 of 571** were `_null_`). `data _null_;` produces no dataset; it is a side-effect-only step.
- **Unexpanded macro vars** become modules (`&OUTFILE_COMP.`, `&PATHNAME.\SASData`: 85 contain `&`, 7 are path fragments). These have no stable identity (the real name only exists after macro expansion).

The same unanchored pattern repeats at the CREATE-output capture (L1424) and the FROM/JOIN/SET/MERGE/UPDATE import captures (L1432, L1439, L1446, L1452), so the IMPORTS side has the same double-edge + macro-var-target problem.

Fix: capture the whole node, `(data_step (data_step_header (dataset_name) @name)) @definition.module`, so `work.foo` is one entity; then skip names matching `/^_(null|data|last)_$/i` and (for module defs) names containing `&`. Apply to all 8 capture sites.

Optional grammar ergonomics (not required, the CLI can fix everything above): a two-part `dataset_name` exposes `lib`/`member` as bare sibling identifiers with no `library:`/`member:` field labels, and `_null_` is aliased to a plain `identifier`. Field labels plus a distinct sentinel node would let these be handled with query predicates instead of post-filters. Filing that on tree-sitter-sas is optional, not blocking.

core-ingestion/src/queries.ts
```diff
@@ -1382,6 +1382,79 @@ export const SCALA_QUERIES = `
...
+
+; PROC step — captures the procedure name
+(proc_step
+  (proc_step_header
+    name: (identifier) @name)) @definition.module
```
josephismikhail, line 1511: **PROC names become duplicated `module` nodes, one per file, even though `SAS_BUILTINS` suppresses them as CALLS.** `proc means` across 40 files yields 40 nodes literally named `means`. The asymmetry is the issue: `SAS_BUILTINS` (index.ts:214) drops `print`/`means`/`freq`/`sort`/... from CALLS edges, but this query still materializes them as entities, and `module` is a structural kind that `ix overview`/`ix rank`/search boost. If a PROC name is too generic to be a call edge, it is too generic to be a graph node.

Fix: key the PROC module on its output/working dataset rather than the procedure name, or reuse `builtinsForLanguage` in the pass-1 definition path for SAS so builtin PROC names do not become nodes.

core-ingestion/src/index.ts
```diff
@@ -266,6 +283,14 @@ function normalizeCapturedImport(rawValue: string, language: SupportedLanguages)
     return unwrapped.replace(/^\.+/, '');
   }
 
+  if (language === SupportedLanguages.SAS && !unwrapped.includes('/') && !unwrapped.includes('\\')) {
```
josephismikhail: **Windows single-backslash paths leak into IMPORTS/module names.** Real SAS uses single backslashes, e.g. `libname OUTMSR "&PATHNAME.\SASData";`. The pre-normalization at the top of this function only converts *double* backslashes (`replace(/\\/g, '/')` matches `\\`), so a single backslash survives; this SAS branch's `!unwrapped.includes('\\')` guard is then false and the value falls through to the general handler, which splits only on `/` and returns the whole string. Result: an IMPORTS dstName (and a shadow `module` node) literally named `&PATHNAME.\SASData`. This is the source of the 7 path-fragment modules, and `&`-only macro-var targets fall through here unchanged too.

Fix: for SAS, convert single backslashes to `/` before the `includes`/`split` checks (or split on `/[\\/]/`), and drop import targets that still contain `&` (unexpanded macro var) so they do not create bogus nodes/edges.

core-ingestion/src/index.ts
```diff
@@ -32,7 +32,11 @@ function tryLoadGrammar(pkg: string): any {
   try { return _require(pkg); } catch { return null; }
 }
 const Kotlin = tryLoadGrammar('tree-sitter-kotlin');
-const Swift = tryLoadGrammar('tree-sitter-swift');
+const Swift  = tryLoadGrammar('tree-sitter-swift');
+// tree-sitter-sas uses ESM bindings with top-level await — incompatible with tryLoadGrammar (CJS require)
+let SAS: any = null;
+// @ts-ignore
+try { SAS = (await import('tree-sitter-sas')).default; } catch { }
```
josephismikhail: Two low-severity notes on this load:

1. The bare `catch {}` swallows every error, so a *broken native build* (ABI mismatch, missing prebuild) is indistinguishable from "grammar not installed" — the user gets zero SAS nodes and no diagnostic (the `.sas` files are then silently dropped by the grammar-supported gate). Consider surfacing non-`MODULE_NOT_FOUND` errors: `catch (e) { if (e?.code !== 'ERR_MODULE_NOT_FOUND' && e?.code !== 'MODULE_NOT_FOUND') process.stderr.write('[tree-sitter-sas load failed] ' + e + '\n'); }`.

2. This `await import` runs eagerly at module evaluation, so the SAS native addon is dlopen'd once per worker thread on every ingest, including pure non-SAS repos. It mirrors the existing Kotlin/Swift eager loads so it is acceptable; just noting it now sits on the per-worker hot path. Both points are optional/deferrable.

core-ingestion/src/index.ts
```diff
@@ -1921,6 +1946,24 @@ export function buildGlobalResolutionIndex(
       }
     }
 
+    // SAS: scan macro definitions (%macro name) for cross-file Tier-3 resolution.
+    const sasMacroRe = /^\s*%macro\s+([a-zA-Z_][a-zA-Z0-9_]*)\s*[(/;]/gim;
```
josephismikhail: Two notes on cross-file macro resolution:

1. This regex (not the parser) feeds the cross-batch index. It handles indentation (the Bug-2 fix), but misses `%macro /*comment*/ name` — the parser treats the comment as an `extra` and still captures the macro, so the regex diverges from what is parsed in-batch. Rare; low priority. Long-term, deriving the cross-batch index from the parser's `definition.macro` entities would remove the divergence entirely.

2. Bigger picture (corpus reality, not a code bug here): SAS macro libraries redefine the same macro name across many files. In AHRQ, `MDX` is defined in 7 files and `label_qis` in 20. The Tier-3 resolver drops multiply-defined names as ambiguous (skipped-ambiguous=59 on AHRQ), so `ix callees`/`ix callers` on those library macros return empty. The `%include`-then-call idiom still resolves via Tier-2, so it is bounded, but worth deciding whether to (a) emit low-confidence edges to all definers, or (b) prefer the `%include`-reachable definer plus same-directory proximity, so the macro call graph (the main point of SAS support) is not severed for the most-reused macros.

core-ingestion/src/index.ts
```diff
@@ -2112,7 +2155,7 @@ export function resolveEdges(
     if (stripped && stripped !== normalizedModName) candidates.push(stripped);
     // Strip file extensions so "explain.js" resolves to the "explain" stem
     // (TS/JS ESM imports use .js extensions that map to .ts source files)
-    const noExt = normalizedModName.replace(/\.(js|ts|mjs|cjs|jsx|tsx|py|scala|java|c|cc|cpp|cxx|h|hh|hpp|hxx)$/, '');
+    const noExt = normalizedModName.replace(/\.(js|ts|mjs|cjs|jsx|tsx|py|scala|java|c|cc|cpp|cxx|h|hh|hpp|hxx|sas)$/i, '');
```
josephismikhail: Adding `/i` here makes extension-stripping case-insensitive for **all** languages, not just SAS. An import target ending in uppercase `.TS`, `.C`, `.H`, `.JAVA` now gets its suffix stripped and the bare stem added as a resolution candidate (it was not before). Uppercase `.C`/`.H` are legitimate C++ source extensions, so this is a narrow cross-language behavior change. Low risk (the stem is only added as a candidate and only matches when a real same-stem file exists), but if SAS is the only reason for `/i`, consider gating it to the `sas` alternative and keeping the rest case-sensitive.

core-ingestion/src/queries.ts
BrandonGarate177: Okay so I fixed in `0da2a7b`. Now grabs the whole `(dataset_name) @name` node, so `work.foo` stays one entity instead of splitting into two, and the `.sas` filename double captures are gone.


`_null_/_data_/_last_` sentinels and `&`-names get filtered out in `index.ts`.

core-ingestion/src/queries.ts
BrandonGarate177: Fixed in `0da2a7b`. PROC steps are now keyed on their `out=` dataset instead of the procedure name. 

If there's no output ---`proc print`, or `proc means` without `out=` --- the step just gets dropped rather than spawning a generic `means`/`sort` node.

core-ingestion/src/index.ts
BrandonGarate177: Makes sense. SAS import normalization converts single backslashes to `/` before it looks at the path, and drops anything still holding a `&`

core-ingestion/src/index.ts
BrandonGarate177: Got it. Newest commit -> non-`MODULE_NOT_FOUND` load errors now go to stderr, so a broken native build doesn't look the same as "grammar not installed."

I'm going to skip 2 for now.

core-ingestion/src/index.ts
BrandonGarate177: the cross-batch index is built from the parser's `definition.macro` entities now instead of the regex. 


I honestly have no idea which direction to lean towards here.

core-ingestion/src/index.ts
BrandonGarate177: Fixed in `0da2a7b` scoped the case-insensitivity to just the SAS extension (`[Ss][Aa][Ss]`).

core-ingestion/src/index.ts
BrandonGarate177: Newest commit addresses this. we ended up leaning towards option B, was relatively easy.

 Include SAS alongside Java in the proximity-based disambiguation used by resolveEdges so multiply

BrandonGarate177:
All six inline findings are in (`0da2a7b`):
- `dataset_name` grabs the whole node + filters sentinel/`&macro` junk → no more `.sas`-filename / `_null_` / macro-var noise
- PROC steps key off `out=`, dropped if no output
- single backslashes normalized, `&`-targets dropped
- bare `catch` surfaces real load errors instead of swallowing them
- cross-batch macro index is parser-derived now (can't drift)
- `/i` strip scoped to just `.sas`

On the multiply-defined macro thing (`MDX` in 7 files → empty `ix callees`) — just did option (b) instead of punting it (`<new SHA>`):
- duplicated macro calls resolve to the closest definer by path proximity instead of dropping
- reuses the path-prefix tiebreaker already there for Java/C++, just on for SAS
- stays safe: genuinely-equidistant definers still drop rather than guess
- only kicks in for the no-`%include` case (with `%include`, Tier-2 already handles it)
- tests for both the "picks closer" and "stays ambiguous" paths

Left two of your "optional" notes alone on purpose:
- eager-load hot path — just mirrors what Kotlin/Swift already do
- the bigger fan-out-to-all-definers flavor of (a) — needs an edge-level confidence field we don't have yet (pulls in backend + CLI), so out of scope here

@josephismikhail, let me know how this looks

josephismikhail (approved):
Verified the fixes for all 6 review items in code, plus extensive empirical testing on the AHRQ corpus (84 files, node:22): zero parse failures, fully deterministic, module-node noise eliminated (all categories 0, down from ~30%), and macro-call resolution fixed (skipped-ambiguous 59 -> 0 via path proximity). Added a small follow-up so PROC out= datasets in body OUTPUT/TABLES statements are captured too. No regressions: the only failing unit tests are pre-existing on the merge-base. Merged current main and resolved the language-registry conflicts.

PR #209, opened 2026-05-16
feat(cli): add --detailed flag to ix subsystems --list

jcorn925:
## Summary
- Adds `--detailed` to `ix subsystems --list`, calling `GET /v1/subsystems?detailed=true`
- Returns full score payloads including `member_files` and enriched `calls_*` / `imports_*` edges
- Bumps CLI version to 0.6.1

## Type
- [x] Feature

## Changes
- `ix-cli/src/cli/commands/subsystems.ts` — new flag, validation, JSON output paths
- `ix-cli/src/client/api.ts` — `listSubsystems({ detailed })` query param

## Validation
- `curl "http://localhost:8090/v1/subsystems?detailed=true"` returns `member_files`, `calls_*`, `imports_*`
- `node ix-cli/dist/cli/main.js subsystems --list --detailed --format json` matches API shape
- `ix subsystems --detailed` without `--list` errors as expected

## Checklist
- [x] `ix-cli/package.json` version bumped to 0.6.1
- [ ] After merge: tag `v0.6.1` on upstream to trigger release workflow

Made with [Cursor](https://cursor.com)

### DIFF
--- a/ix-cli/package-lock.json
+++ b/ix-cli/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@ix/cli",
-  "version": "0.5.1",
+  "version": "0.6.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@ix/cli",
-      "version": "0.5.1",
+      "version": "0.6.1",
       "dependencies": {
         "chalk": "^5.3.0",
         "commander": "^12.0.0",
--- a/ix-cli/package.json
+++ b/ix-cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ix/cli",
-  "version": "0.6.0",
+  "version": "0.6.1",
   "type": "module",
   "description": "Ix Memory — Persistent Memory for LLM Systems",
   "main": "dist/index.js",
--- a/ix-cli/src/cli/commands/subsystems.ts
+++ b/ix-cli/src/cli/commands/subsystems.ts
@@ -39,6 +39,7 @@ export function registerSubsystemsCommand(program: Command): void {
     .description("Show the persisted architectural map saved by 'ix map'")
     .option("--format <fmt>", "Output format (text|json)", "text")
     .option("--list",         "List stored subsystem health scores instead of the persisted architecture map")
+    .option("--detailed",     "Include member files and enriched call/import edges (requires --list)")
     .option("--target <target>", "Scope subsystem output to a persisted architecture region")
     .option("--pick <n>", "Resolve an ambiguous region target by numbered candidate")
     .option("--level <n>",    "Filter to level (1=module, 2=subsystem, 3=system)")
@@ -57,6 +58,8 @@ ingestion or clustering.
 Examples:
   ix subsystems
   ix subsystems --list
+  ix subsystems --list --detailed
+  ix subsystems --list --detailed --format json
   ix subsystems api
   ix subsystems --target "Cli / Client"
   ix subsystems api --pick 2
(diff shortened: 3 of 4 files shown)


josephismikhail:
Re-reviewed against the updated server side (`reapply/subsystems-detailed-api-v2` at `5165246`). The memory-layer half is good. OOM fix shipped, response is bounded at ~6 MB worst case, schema includes new `*_total` fields for truncation indicators, and 400s are returned on invalid params.

  **However, the CLI side hasn't been updated to match the new server contract**, which causes a real UX regression:

  ### Repro

  Fresh ingest of guava (178 scored regions) against the updated memory-layer:

  ```
  $ ix subsystems --list           # baseline, no flag
  → 178 regions

  $ ix subsystems --list --detailed
  → 50 regions (silently truncated by server default limit)
  ```

  The CLI calls `GET /v1/subsystems?detailed=true` with no query params, so it gets the server's default page (50 regions, default caps). There is no indication to the user that 128 more regions exist, no way to access pages 2+, and no way to scope the request to specific regions.

  ### Cause

  `listSubsystems({ detailed })` in `ix-cli/src/client/api.ts` doesn't pass through any of the new server-side query params (`limit`, `offset`, `regions`, `edge_cap`, `member_file_cap`). The `subsystems --list --detailed` command in `ix-cli/src/cli/commands/subsystems.ts` has no
  flags for them either.

  ### Suggested fixes, pick whichever fits your taste

  1. **Auto-paginate.** CLI loops with `offset += returned` until `returned < limit`, stitches the result. User sees the full set with no new flags. Simplest UX, but issues N HTTP calls for large maps.
  2. **Pass-through flags.** Add `--limit`, `--offset`, `--regions`, `--edge-cap`, `--member-file-cap` to `subsystems --list`. Maximum control; matches REST contract verbatim. Requires user to know about pagination.
  3. **Default to max limit (200) and warn on truncation.** Send `?limit=200` by default; if response's `pagination.returned < total_regions` (or however the server signals truncation), print a one-line warning telling the user to use `--regions` or page manually. Cheapest middle
  ground.

  My vote is **(1) + (2)**. Auto-paginate by default for the common case, plus expose flags for power users / scripts that want to scope or stream.

  ### Other small notes from the re-review

  - The CLI's text-format `--detailed` path falls through to a raw JSON dump (no `printScores` formatter). Same as before — acceptable, but now that responses can have `pagination` metadata, it's worth either including that in the JSON dump or omitting it deliberately.
  - Version bump to 0.6.1 in `package.json` still applies cleanly; no rebase needed against current main.

  Memory-layer side LGTM, CLI side blocked on this. Once the CLI is wired up, both can ship together as 0.6.1.

  Adjust the "my vote" line if you have a strong preference different from mine.

josephismikhail (approved):
Tested against the paired memory-layer reapply branch with guava plus lintr ingested (179 scored regions). Every behavior from the prior review comment is addressed, and you added two unprompted UX touches that I think are genuinely valuable.

Verified end-to-end:
- Default ix subsystems --list --detailed auto-paginates and returns all 179 regions in one shot
- --limit 50 still returns all 179 because the CLI loops until returned < limit (correctly uses the pagination.returned + pagination.limit metadata to terminate)
- --offset disables auto-pagination and returns a single page, as expected
- --regions, --edge-cap, --member-file-cap all pass through correctly
- Page-full warning fires when explicit --offset is used and the page is full, AND it includes the exact next-offset value the user should try. That is a really nice touch.
- Per-region truncation warning fires when any region has *_total > array length, with a sensible suggestion to use --edge-cap, --member-file-cap, or --regions
- All error paths land cleanly: --limit without --detailed, invalid integer values, negative values, empty --regions, etc.

One tiny non-blocker for a future commit if you want: the auto-paginate while-loop has no max-iteration cap. If a server bug ever caused returned >= limit indefinitely, the CLI would hang. A safety guard of something like 200 iterations would be cheap insurance.

Approving. Thanks for both the original implementation and the quick turnaround on the pagination feedback. Really appreciate the open-source contribution. The two-tier warning system is a nicer UX than what I suggested.

PR #211, opened 2026-05-21
feat: new Makefile parser

### LINKED ISSUE #128: Add Makefile parser support (Makefile, *.mk)
## Problem

Ix does not currently parse `Makefile` or `*.mk` files. These files are treated as unrecognized and are excluded from analysis, meaning build logic and task definitions are invisible to the system.

## Proposed solution

Add a Makefile parser that can identify and extract:
- Target definitions (e.g. `build:`, `test:`, `clean:`)
- Variable declarations (e.g. `CC = gcc`, `FLAGS ?= -O2`)
- Include directives (e.g. `include config.mk`)
- Phony targets (`.PHONY`)

Support the following file patterns: `Makefile`, `GNUmakefile`, `makefile`, `*.mk`

## Why it matters

Makefiles are ubiquitous in C/C++, Go, and polyglot repos as the primary build orchestration layer. Without parsing them, Ix has a blind spot over how projects are built, what tasks exist, and how source files relate to build outputs. This is especially relevant for infrastructure and systems codebases where Make is the entry point for build, test, and deploy workflows.

## Merge requirements

Before this can be merged to main, the following must pass:

- [ ] Tested on 3 repos
- [ ] No regressions on existing parsers (`ix map` on a known TS/JS repo, verify counts unchanged)
- [ ] Unit + smoke tests pass (`npm test` in `ix-cli`)
- [ ] Target definitions ingested as entities with their prerequisites
- [ ] `.PHONY` declarations reflected on their target entities
- [ ] Variable declarations ingested (`=`, `:=`, `?=`, `+=`)
- [ ] `include` directives resolved and linked to target file where possible
- [ ] All filename patterns ingested: `Makefile`, `GNUmakefile`, `makefile`, `*.mk`
- [ ] CONTAINS chains correctly: file → target → recipe lines
- [ ] `ix text` returns results with `language: makefile`
- [ ] `ix contains` returns targets for a known Makefile

NKlein77:
Closes #128 

## Summary
What does this PR change?
    Adds a Makefile language parser that ingests Makefile, GNUmakefile, makefile, and *.mk files into the Ix graph. No previous makefiles or like files were recognized. 

## Type
- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
-Added makefile-parpser.ts with support for target definitions and prerequisites
-Added .PHONY declarations detections and reflection on target entities
-Added variable declarations ingestion for =, :=. ?=. and += operators
-Added include directive resolution and linking
-Added support for all filename patterns, (Makefile, GNUmakefile, makefile,*.mk)
-Added CONTAINS chains: file -> target -> recipe lines
-Added makefile-parser.tsst.ts with unit tests covering all entity types
-Added Makefile to SupportedLanguages in language.ts

## Validation
How did you test this?
   -All unit tests pass (npm test)
    -Tested against 3 real repos with Makefiles (can be found in comments sections)

## Checklist
- [x] Tests pass
- [ ] Smoke tests pass


NKlein77:
tested repos
    https://github.com/curl/curl
    https://github.com/redis/redis
    https://github.com/git/git

josephismikhail (changes requested):
Needs significant changes before merge. Most issues are about how language support integrates with the rest of the pipeline, not the parser logic itself.

## How language support works in core-ingestion

Every existing language plugs into `parseFile()` in `core-ingestion/src/index.ts` (around line 1230). Flow:

1. **`languages.ts`** defines the `SupportedLanguages` enum + extension-to-language map.
2. **`index.ts`** registers a tree-sitter grammar in `GRAMMAR_MAP` (see how `Kotlin`, `Swift`, `SAS` are wired — search for `tryLoadGrammar` or the SAS dynamic import).
3. **`queries.ts`** defines tree-sitter query patterns that extract entities and edges per language.
4. `parseFile()` reads the file, looks up the grammar from `GRAMMAR_MAP`, runs the queries, returns a `FileParseResult` (entities + intra-file edges) that the rest of the pipeline consumes.

This PR's `makefile-parser.ts` lives outside that dispatch entirely. Nothing in the codebase calls `parseMakefile()` or `emitEntities()`. After this PR merges, `ix map .` on a repo with Makefiles still skips them — same behavior as before the PR.

The right pattern: register a tree-sitter grammar for Make and write queries against it, same way SAS (#208) was added. References:
- [PR #208](https://github.com/ix-infrastructure/Ix/pull/208) — the most recent example, end-to-end.
- `core-ingestion/src/queries.ts` — search for `SAS_QUERIES` for the query format.
- `core-ingestion/src/index.ts` — see how `SAS` is loaded into `GRAMMAR_MAP` (around line 35-40).

`tree-sitter-make` is published on npm. Steps:

1. Add `tree-sitter-make` to `core-ingestion/package.json` under `optionalDependencies`
2. Add `Makefile = 'makefile'` to `SupportedLanguages` in `languages.ts`
3. Add file patterns (`Makefile`, `GNUmakefile`, `makefile`, `*.mk`) to language detection in `languages.ts`
4. Add a `MAKEFILE_QUERIES` constant in `queries.ts`
5. Load the grammar in `index.ts` and add to `GRAMMAR_MAP`
6. Delete `makefile-parser.ts`

## Unrelated change in `index.ts` that must be reverted

The diff removes the `packageScope` field from `ParsedEntity` and the `'_go_param_scope'` capture-name from the pass1 detection in `parseFile()`. This has nothing to do with Makefiles and would break Go parsing. The deleted docstring explains the consequences:

> Used by downstream layers to prefix qualifiedName so the symbol table indexes entities under their fully-qualified path. Without this, bare-name resolution at edge-resolver collides between packages (e.g. kubernetes has 191 distinct `addKnownTypes` functions, one per API group) and CALL targets are picked non-deterministically.

Likely caused by a stale branch — old content came back in during a rebase or merge. To find and revert:

```bash
git fetch origin main
git diff origin/main -- core-ingestion/src/index.ts
```

Revert the `packageScope` field (around line 51-60) and the `_go_param_scope` capture name (around line 1244). Leave the rest of the diff alone.

## PR description doesn't match the diff

- Body says "Added Makefile to SupportedLanguages in language.ts" — `languages.ts` isn't in the changed files. Update after the rewrite.
- Body claims include resolution and linking — `resolveIncludes()` exists but isn't called from anywhere. Decide whether include resolution is in v1 or punted.

## Smaller issues

- Typos in the description: "makefile-parpser.ts", "makefile-parser.tsst.ts".
- Logic bug in the current parser: `if (line.includes(':') && line.startsWith('\t'))` is contradictory — Makefile targets contain `:` and don't start with tab; recipes start with tab and don't contain `:`. That block never fires. (Goes away in the tree-sitter rewrite.)
- The single unit test only checks variable count. Post-rewrite, add tests asserting the entities and intra-file edges. See `core-ingestion/src/__tests__/queries.sas.test.ts` for the pattern.

## Suggested split

**Commit 1:** revert the unrelated Go changes. Restore `packageScope` and `_go_param_scope`.

**Commit 2:** replace the hand-rolled parser with tree-sitter-make integration per the 6 steps above.

If `tree-sitter-make` turns out to be missing features needed, that's a fair time to come back and discuss — at that point a hand-rolled fallback might be the right answer, but make that call deliberately rather than as the default.

Related: ix-infrastructure/ix-memory-layer#27 (parked) tracks changing the language-support contract so contributors can add languages without backend coordination. Not in scope here — follow the existing pattern.

NKlein77:
Ran into a tree-sitter version incompatibility: tree-sitter-make only has versions 1.1.0 and 1.1.1 on npm, both of which require tree-sitter@^0.22.1. This project is on tree-sitter@0.21. The grammar loads but crashes at parse time with a native ABI error. Upgrading tree-sitter would require bumping all ~15 grammar packages simultaneously.

All the wiring is in place (languages.ts, index.ts, queries.ts, ingest.ts). How do you want to proceed — hold for a tree-sitter upgrade PR, or discuss another way to approach this?

josephismikhail:
Following up on the tree-sitter question. The blocker isn't real at the runtime level, so we don't need a tree-sitter upgrade or the hand-rolled parser. I wired the tree-sitter route locally and verified it end to end (ingested a repo with Makefiles: entities land in the graph, 0 parse errors, no regressions in the existing test suite).

### tree-sitter-make works on tree-sitter@0.21

The `^0.22.1` is a peer-dependency *declaration* npm checks at install time, not a runtime requirement. The grammar is compiled at ABI 14 and the 0.21.x runtime accepts ABI 13 and 14, so it loads and parses. Repro:

```bash
mkdir ts-make && cd ts-make && npm init -y
npm install tree-sitter@0.21.1 tree-sitter-make@1.1.1 --legacy-peer-deps
```

`probe.js`:
```js
const Parser = require('tree-sitter');
const Make = require('tree-sitter-make');
const p = new Parser();
p.setLanguage(Make); // throws here if the ABI were incompatible; it doesn't
const tree = p.parse('CC := gcc\nall build test: $(OBJ)\n\t@echo "linking: $@"\ninclude config.mk\n');
console.log('hasError:', tree.rootNode.hasError);
const q = new Parser.Query(Make, '(rule (targets (word) @name)) @def (variable_assignment name: (word) @n) (include_directive (list (word) @i))');
console.log('matches:', q.matches(tree.rootNode).length);
```

`node probe.js` prints `hasError: false` and a non-zero match count, through the same `new Parser.Query(...).matches(...)` path the pipeline uses.

### The only blocker is npm peer resolution

What I checked so you don't lose time on dead ends:
- `--legacy-peer-deps` resolves and **keeps tree-sitter at 0.21.1**. Don't bump tree-sitter and don't use `--force`; bumping the runtime is what breaks the other ~14 grammars (they're all built for 0.21).
- `overrides` does **not** fix it (npm overrides don't satisfy a peer-dependency conflict).
- `optionalDependencies` alone does **not** fix it; npm still tries to pull `tree-sitter@0.22.4`.

Fix: add `core-ingestion/.npmrc` with `legacy-peer-deps=true`, and list `tree-sitter-make` under `optionalDependencies`. This is also required for the build: `ix-cli/scripts/build-core-ingestion.mjs` runs a bare `npm ci` in core-ingestion, and the `.npmrc` is what lets it resolve. (`npm ls` will then report an unmet peer for the make grammar; that's cosmetic, the runtime is fine.)

### There's a step the earlier list missed

`ix-cli` has its own file-discovery allowlist, separate from core-ingestion: `SUPPORTED_EXTENSIONS` and `isSupportedSourceFile()` in `ix-cli/src/cli/commands/ingest.ts`. Makefiles are extensionless, so without adding `.mk` and the bare names `makefile`/`gnumakefile` there, the CLI never discovers them and nothing gets ingested (no error, it just silently skips them). I only caught this by running an actual ingest.

### Branch state

`39bd9d1` as pushed still only contains `makefile-parser.ts` and its test. The `languages.ts` / `index.ts` / `queries.ts` wiring you mentioned isn't in the diff, so make sure it gets committed.

### Full plan

1. `core-ingestion/package.json`: add `tree-sitter-make` to `optionalDependencies`. Add `core-ingestion/.npmrc` with `legacy-peer-deps=true`.
2. `languages.ts`: add `Makefile = 'makefile'` to `SupportedLanguages`, `.mk` to the ext map, and bare-name detection (`makefile`/`gnumakefile`) in `languageFromPath`.
3. `index.ts`: load via `tryLoadGrammar('tree-sitter-make')` and add to `GRAMMAR_MAP` (same pattern as Kotlin/Swift).
4. `queries.ts`: add `MAKEFILE_QUERIES` and an entry in `LANGUAGE_QUERIES`. Captures that work with the existing pass-1 loop, no new entity code in `index.ts`:
   - `(rule (targets (word) @name)) @definition.function`
   - `(variable_assignment name: (word) @name) @definition.macro`
   - `(include_directive (list (word) @import.name))`
   Filter `.PHONY` out of the target capture so it doesn't get recorded as a target.
5. `ix-cli/src/cli/commands/ingest.ts`: add `.mk` to `SUPPORTED_EXTENSIONS` and the bare names to `isSupportedSourceFile()`.
6. Delete `makefile-parser.ts` and its test. Add tests asserting the entities and intra-file edges (file CONTAINS target/variable, file IMPORTS include), following `queries.sas.test.ts`.

No tree-sitter upgrade and no 15-grammar bump required.

josephismikhail:
Rebuilt this at the PR head and ran it. The feature itself is in good shape: dropping the hand-rolled parser and wiring the tree-sitter route is exactly right, and a probe on a realistic Makefile (simple/recursive/append vars, `.PHONY`, multi-prereq target, pattern rule, `include` + `-include`) produces clean output: targets become functions, variable assignments become macros, `include`/`-include` become IMPORTS, `.PHONY` is filtered, prereq-only names are not captured as targets, and there are no bogus edges. `queries.makefile.test.ts` passes 3/3. The dependency fix (`.npmrc` + `optionalDependencies`) is the correct approach.

Two things to fix before merge:

**1. CI is red on a stale assertion (the actual blocker).** `ix-cli/src/cli/__tests__/ingest-discovery.test.ts:30` still asserts:
```ts
expect(isSupportedSourceFile('Makefile')).toBe(false);
```
Your `ingest.ts` change (correctly) makes that `true` now, so the pre-existing assertion fails:
```
AssertionError: expected true to be false
```
Flip it to `.toBe(true)`, and worth adding positive cases for `GNUmakefile` and `*.mk`.

**2. Unrelated snapshot deletion.** The fix commit also dropped 1853 lines from `core-ingestion/src/__tests__/__snapshots__/parseFile.test.ts.snap` (the `Node.scala` and `Identifiers.scala` blocks). Those fixtures live in the separate ix-memory-layer repo and aren't in this monorepo, so this looks like a `vitest -u` run without the `memory-layer` sibling checked out. It has no CI impact (core-ingestion's vitest isn't run by CI), but it's unrelated to this PR. Revert it:
```
git checkout origin/main -- core-ingestion/src/__tests__/__snapshots__/parseFile.test.ts.snap
```

Also: the branch is ~4 weeks behind main (merge-base is May 6), so a rebase is worth doing.

I'll push a commit with these two so CI goes green.

josephismikhail (approved):
Changes from the earlier review are all addressed and verified: tree-sitter-make route wired (languages/index/queries + ix-cli discovery), hand-rolled parser removed, dependency fixed via .npmrc + optionalDependencies, stale ingest-discovery assertion fixed, and the special-target filter broadened. Verified against 135 real Makefiles (0 crashes) and end-to-end to the GraphPatch. CI green.

PR #220, opened 2026-05-26
Feature/elixir parser

### LINKED ISSUE #139: Add Elixir parser support (.ex, .exs)
## Problem

Ix does not currently parse `.ex` or `.exs` files. Elixir source and script files are treated as unrecognized and excluded from analysis, meaning application logic, process definitions, and configuration written in Elixir are invisible to the system.

## Proposed solution

Add an Elixir parser that can identify and extract:
- Module definitions (e.g. `defmodule MyApp.User do`)
- Function definitions (`def`, `defp`)
- Macros (`defmacro`, `defmacrop`)
- Use, import, alias, and require statements
- Struct definitions (e.g. `defstruct`)
- Pattern matching and guard clauses
- GenServer, Supervisor, and OTP behaviour callbacks
- Mix configuration and script patterns in `.exs` files

Support the following file patterns: `*.ex`, `*.exs`

## Why it matters

Elixir is built on the Erlang VM and is widely used for fault-tolerant, distributed, and real-time systems (e.g. Phoenix web framework, LiveView). Without parsing Elixir, Ix cannot reason about OTP process trees, module dependencies, or how application logic is structured across a concurrent, distributed codebase.

## Merge requirements

Before this can be merged to main, the following must pass:

- [ ] Tested on 3 repos
- [ ] No regressions on existing parsers (`ix map` on a known TS/JS repo, verify counts unchanged)
- [ ] Unit + smoke tests pass (`npm test` in `ix-cli`)
- [ ] Module definitions (`defmodule`) ingested as entities
- [ ] Public and private functions (`def`, `defp`) ingested as entities
- [ ] Macros (`defmacro`, `defmacrop`) ingested as entities
- [ ] `use`, `import`, `alias`, and `require` statements resolved
- [ ] `defstruct` ingested and fields visible
- [ ] OTP callbacks (GenServer, Supervisor) ingested and linked to their module
- [ ] CONTAINS chains correctly: module → function/macro/struct
- [ ] `.exs` files ingested (Mix config and scripts visible)
- [ ] `ix text` returns results with `language: elixir`
- [ ] `ix contains` returns members for a known Elixir module

NKlein77:
Closes #139 

## Summary
What does this PR change?
    Adds tree-sitter-elixir as a dependency in core-ingestion
    Registers .ex/.exs extentions in SupportedLanguages and EXT_MAP (languages.ts)
    Loads the grammar via tryLoadGrammar and adds it to GRAMMAR_MAP (index.ts)
    Adds ELIXIR_QUERIES to queries.ts capturing madoules (defmodule), public/private functions (def/defp), macros (defmacro/defmacrop), module directives (use/alias/import/require) as IMPORTS relationships, and qualified/unqualified calls
    Adds .ex/.exs to SUPPORTED_EXTENSIONS in ingest.ts so Elixir files are discovered during ix ingest
    Removed the Ex_Parser/ sandbox directory that I originally committed with the branch
    Replaces the stanalone regex-based elixir-parser.ts with the tree-sitter inregrations, added 6 integration tests in queries.elixir.test.ts covering modules, functions, imports, calls, .exs files, and macros
## Type
- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
-Replaced elixir-parser.ts with tree-sitter grammar integration following the same pattern as Kotlin/Swift
-Added ELIXIR_QUERIES covering defmodule, def, defp, defmacro, defmacrop, use, import, alias, require and qualified calls
-Added .ex and .exs to SupportedLanguages, EXT_MAP, and SUPPORTED_EXTENSITONS
-Added queries.elixir.test.ts with 8 integrations tests via parseFile()

## Validation
How did you test this?
    Tested against 3 real Elixir repos, repos can be found in the commented section
    8/8 integration tests pass (npm test -- queries.elixir in core-ingestion/)
    No regressions observed on existing TS/JS parsers

## Checklist
- [x] Tests pass
- [ ] Smoke tests pass


NKlein77:
Tested repos
    https://github.com/phoenixframework/phoenix
    https://github.com/elixir-ecto/ecto
    https://github.com/sorentwo/oban

josephismikhail (changes requested):
Needs significant changes. The core issue is the same one in #211: the parser doesn't integrate with `parseFile()`, so `ix map` never calls it. There are also extra problems specific to this PR.

## Empirical verification

Ran `ix map` against `sorentwo/oban` (one of the three repos listed as tested) with this branch's state:

| Signal | Result |
|---|---|
| `.ex` / `.exs` files in oban | 119 |
| File entities added to the graph after `ix map` | 24 — all from `.md` files (Markdown parser) |
| Elixir modules added to the graph | **0** |
| `ix search Oban` | "No results found" |
| `ix inventory --kind file --path _oban` | only `.md` files |

All 119 Elixir source files were silently skipped. The unit tests pass because they test `parseElixir()` directly — `parseFile()` (the real entry point) never calls it. Same root cause as #211: the parser lives outside the dispatch in `core-ingestion/src/index.ts`, isn't registered in `GRAMMAR_MAP`, and `languages.ts` doesn't know about Elixir.

See the review on #211 for the integration pattern (tree-sitter grammar registration, `languages.ts` enum, `queries.ts` queries, `GRAMMAR_MAP` entry). `tree-sitter-elixir` is published on npm and is the standard pattern here.

## Extra issues specific to this PR

### `Ex_Parser/` directory at repo root

The PR adds a complete separate Node project at the repo root: `Ex_Parser/package.json`, `Ex_Parser/tsconfig.json`, `Ex_Parser/jest.config.js`, `Ex_Parser/package-lock.json` (4,804 lines), plus a duplicate copy of `elixir-parser.ts` and its test. This looks like a personal local-development sandbox that got committed by accident.

Delete the entire `Ex_Parser/` directory. The actual parser belongs only at `core-ingestion/src/elixir-parser.ts` (and after the tree-sitter rewrite, even that goes away in favor of queries in `queries.ts`).

The 4,804-line lockfile alone is most of this PR's diff. Removing the sandbox cuts the PR from 5,542 lines to ~250.

### Duplicate parser file

`Ex_Parser/elixir-parser.ts` and `core-ingestion/src/elixir-parser.ts` are byte-identical copies. Drop the `Ex_Parser/` one along with the rest of that directory.

### PR description doesn't match the diff

Body says "Added Elixir to SupportedLanguages in language.ts" — `languages.ts` isn't in the changed files. After the rewrite using `tree-sitter-elixir`, it will be. Update the description to match what actually changes.

## Smaller issues

- The current parser declares `ElixerAST` (typo for `ElixirAST`). After the tree-sitter rewrite this goes away.
- `OTP_CALLBACKS` is a hardcoded `Set` — same approach as the SAS builtins list in `index.ts`, which is fine, but it's only useful if the parser is actually called.
- The 9 unit tests prove the standalone parser works in isolation. After integration, port the same intent into `core-ingestion/src/__tests__/queries.elixir.test.ts` against tree-sitter output (see `queries.sas.test.ts` for the format).

## Plan

1. Delete `Ex_Parser/` and the duplicate makefile changes that came along with it.
2. Replace `core-ingestion/src/elixir-parser.ts` with `tree-sitter-elixir` integration following the same 5 steps in the #211 review.
3. Confirm Elixir files actually land in the graph by re-running `ix map` against oban or phoenix — `ix inventory --kind file --path <repo>` should show `.ex` files; `ix search SomeModuleName` should find entities.

Related: ix-infrastructure/ix-memory-layer#27 (parked) tracks changing the language-support contract so contributors can add languages without backend coordination. Not in scope here — follow the existing pattern.

josephismikhail:
Integration is fixed (thanks for splitting out `Ex_Parser/` and wiring up `languages.ts` / `GRAMMAR_MAP` / `ingest.ts`). The 6 tests pass and `parseFile()` actually produces entities for `.ex` files now. But the query content has three bugs that corrupt CALLS edges on every Elixir file, and the tests don't catch them because they all use `arrayContaining` / `toContainEqual`.

I checked out the branch, ran the tests (pass), then ran a probe with no assertions against a realistic module:

```elixir
defmodule MyApp.User do
  use GenServer
  alias MyApp.Repo
  import Ecto.Query

  def start_link(opts) do
    GenServer.start_link(__MODULE__, opts)
  end

  defp validate(user) do
    Repo.get(User, user.id)
    user |> Map.put(:validated, true) |> Repo.update()
  end
end
```

Actual CALLS edges produced:

```
MyApp.User              CALLS  defmodule
MyApp.User              CALLS  use
MyApp.User              CALLS  alias
MyApp.User              CALLS  import
MyApp.User.start_link   CALLS  def
MyApp.User.start_link   CALLS  start_link    (self-call)
MyApp.User.validate     CALLS  defp
MyApp.User.validate     CALLS  validate      (self-call)
MyApp.User.validate     CALLS  get           (was Repo.get)
MyApp.User.validate     CALLS  put           (was Map.put)
MyApp.User.validate     CALLS  update        (was Repo.update)
```

### 1. Catch-all CALLS query matches every keyword

```
(call target: (identifier) @call.name) @call
```

In Elixir, **every keyword-like form is a `call` node with an `identifier` target**: `def`, `defp`, `defmacro`, `defmacrop`, `defmodule`, `defprotocol`, `defimpl`, `defstruct`, `use`, `alias`, `import`, `require`, `quote`, `if`, `case`, `with`, `for`, `try`, `receive`, etc. Every Elixir file in oban/phoenix/ecto will inject dozens of bogus `CALLS -> defmodule|def|use|...` edges.

Fix: add `(#not-match? @call.name \"^(def|defp|defmacro|defmacrop|defmodule|defprotocol|defimpl|defstruct|use|alias|import|require|quote|unquote|if|unless|case|cond|with|for|try|receive|raise|throw|fn|do)$\")` to the catch-all.

### 2. Every function definition self-calls itself

`def foo(args)` parses as `call{target=def, args=[call{target=foo}]}`. The inner `call` matches the catch-all, so every function gets a `foo CALLS foo` edge. This is graph-corrupting (every Elixir function ends up with a self-loop). Confirmed for every def I tried.

Fix: the inner call inside `(arguments)` of a `def*` form must be excluded from the catch-all. Either anchor the catch-all to skip calls whose parent is `(arguments)` of a `def*` call, or use a predicate that suppresses emission for the captured def name.

### 3. Qualified calls drop the module

```
(call target: (dot left: (alias) right: (identifier) @call.name)) @call
```

`GenServer.start_link`, `Repo.get`, `Map.put` all collapse to bare `start_link` / `get` / `put`. The `(alias)` on the left is thrown away, so cross-module CALLS resolution against the alias table is impossible. Capture it as `@_qualifier` (see how other languages do it in `index.ts:1511`) so the dispatch can resolve `Repo` to `MyApp.Repo`.

### 4. `defmacro` / `defmacrop` produce kind `function`, not `macro`

`DEFINITION_KIND_MAP` already has `definition.macro -> 'macro'`. Your query lumps macros into `@definition.function`. Split them out so kind is preserved.

### 5. `defprotocol` / `defimpl` not captured

A `defprotocol MyApp.Serializer do ... end` produces no class entity, and its `def` clauses become container-less orphans. Same for `defimpl`. Both are extremely common in real Elixir code. Add patterns mirroring the `defmodule` one but for these two forms.

### 6. Tests don't catch (1) or (2)

All current assertions use `arrayContaining` / `toContainEqual`, so noise is invisible. Add **negative** assertions:

```ts
const callTargets = result!.relationships
  .filter(r => r.predicate === 'CALLS')
  .map(r => r.dstName);
expect(callTargets).not.toContain('def');
expect(callTargets).not.toContain('defp');
expect(callTargets).not.toContain('defmodule');
expect(callTargets).not.toContain('use');
expect(callTargets).not.toContain('alias');
expect(callTargets).not.toContain('start_link'); // self-call from the def itself
```

These would have caught (1) and (2) on the first run.

### 7. Unrelated Makefile parser bundled in

This branch also contains commit `feat: new Makefile parser` adding `core-ingestion/src/makefile-parser.ts` (+209) and a test file (+19). Neither is mentioned in the PR description, neither relates to Elixir, and the makefile test is only a `typeof === 'function'` smoke check with no real coverage. Pull it out into its own PR (or disclose it here and add real tests).

### 8. PR description typos

Body says `language.ts` and `SUPPORTED_EXTENSITONS`. Actual: `languages.ts` and `SUPPORTED_EXTENSIONS`.

### Why this matters before merge

Once this ships, every Elixir-using repo we ingest gets a graph full of `CALLS -> defmodule`, `CALLS -> def`, self-call edges on every function, and qualified calls with the module stripped. That's hard to clean up retroactively. Fixing the queries before merge is much cheaper than re-ingesting later.

Plumbing is solid. It's just the query content that needs another pass.

core-ingestion/src/queries.ts
```diff
@@ -1164,6 +1164,66 @@ export const SCALA_QUERIES = `
...
+  target: (identifier) @_defimpl
+  (arguments (alias) @name)
+  (#eq? @_defimpl "defimpl")) @definition.class
+
+; ── Function definitions with parens: def foo(args) do … end ─────────────────
+(call
+  target: (identifier) @_def
+  (arguments
+    (call target: (identifier) @name))
+  (#match? @_def "^(def|defp)$")) @definition.function
```
josephismikhail, line 1419: **Critical: guarded functions are dropped entirely, and they leak a bogus CALLS edge.**

This def query (and the defmacro one below) requires the function head to be a direct `(call)` child of `(arguments)`. But a guard clause `def f(x) when is_integer(x) do` parses with the head wrapped in a `when` binary_operator:

```
(arguments (binary_operator left: (call target:(identifier) ...) right: (call target:(identifier) ...)))
```

so the head call is no longer a direct child of `arguments`. Neither this query nor the zero-arity query matches, so **no `@definition.function` entity is created for the clause**. Worse, the unqualified-call query still matches the inner `f(x)` call, and the self-edge skip in index.ts (L1535) only fires when the inner call's grandparent is `arguments` (here it is `binary_operator`), so a spurious `<enclosing module> CALLS f` (and `CALLS is_integer`) edge is emitted. Guards are pervasive in idiomatic Elixir/Ecto, so this silently drops a large class of functions while injecting noise. The ecto ingest counted 1350 functions precisely because those are the *unguarded* clauses.

The same `(arguments (binary_operator ...))` blind spot also drops operator definitions like `def a + b do ... end` (the head is a `binary_operator`, not a `(call)` or bare identifier).

Fix: add a guarded-def variant that descends the `when` operand, and the same for defmacro/defmacrop:
```
(call target: (identifier) @_defg
  (arguments (binary_operator (when) left: (call target: (identifier) @name)))
  (#match? @_defg "^(def|defp)$")) @definition.function
```
(note: `when` is a named node, match it as `(when)`, not the string `"when"`). Then extend the index.ts self-edge skip (L1535) to walk through an intervening `binary_operator` so the guarded head does not emit a bogus CALLS edge. Please add guard + operator-def test cases.

core-ingestion/src/queries.ts
```diff
@@ -1164,6 +1164,66 @@ export const SCALA_QUERIES = `
...
+; ── Qualified calls: Foo.bar() — captures qualifier so resolution is possible ─
+(call
+  target: (dot
+    left: (alias) @_qualifier
+    right: (identifier) @call.name)) @call
+
+; ── Unqualified calls — keywords excluded to avoid bogus edges ────────────────
+(call
+  target: (identifier) @call.name
+  (#not-match? @call.name "^(def|defp|defmacro|defmacrop|defmodule|defprotocol|defimpl|defstruct|use|alias|import|require|quote|unquote|if|unless|case|cond|with|for|try|receive|raise|throw|fn|do)$")) @call
```
josephismikhail: **High: module attributes `@behaviour` / `@impl` / `@derive` emit bogus CALLS to phantom nodes and discard the real target.**

`@behaviour GenServer`, `@impl true`, `@derive Jason.Encoder` parse as a unary `@` operator whose operand is `call{target: behaviour, args:[GenServer]}`. This unqualified-call query captures the inner identifier (`behaviour`/`impl`/`derive`), none of which are in the exclusion list, so each becomes a `Module CALLS behaviour` edge to a phantom node, while the real target module (`GenServer`, `Jason.Encoder`) is thrown away. These attributes appear on nearly every GenServer/Ecto/protocol module, so this is high-volume graph pollution that also loses genuine behaviour/protocol signal.

Empirically on a probe module: `@behaviour GenServer / @impl true / @derive Jason.Encoder` produced `Module CALLS behaviour`, `Module CALLS impl`, `Module CALLS derive`, with GenServer/Jason.Encoder absent.

Fix: detect that the call's ancestor is a unary `@` operator and skip it (cleanest), or add `behaviour|behavior|impl|derive|moduledoc|doc|spec|type|typep|opaque|callback|enforce_keys|...` to this `#not-match?` list. Better still, add a dedicated `@behaviour <Alias>` pattern emitting an EXTENDS/IMPORTS edge to the behaviour module to recover the signal.

core-ingestion/src/queries.ts
```diff
@@ -1164,6 +1164,66 @@ export const SCALA_QUERIES = `
...
+  target: (identifier) @_defmacro0
+  (arguments
+    (identifier) @name)
+  (#match? @_defmacro0 "^(defmacro|defmacrop)$")) @definition.macro
+
+; ── Module directives → IMPORTS ───────────────────────────────────────────────
+(call
+  target: (identifier) @_directive
+  (arguments (alias) @import.source)
+  (#match? @_directive "^(alias|import|use|require)$")) @import
```
josephismikhail, line 1446: **High: grouped multi-alias `alias MyApp.{User, Repo, Post}` captures zero IMPORTS.**

This directive query matches a single bare `(alias)` argument. The grouped form (the idiomatic way to alias siblings) parses as `(arguments (dot left: (alias) right: (tuple)))`, not a bare `(alias)`, so it matches nothing: no IMPORTS edges for `User`/`Repo`/`Post` and no module entities for them. Every subsequent qualified call through those aliases then also fails import-scoped resolution, so this systematically under-counts IMPORTS (the 326 figure is missing all grouped aliases).

Fix: add a grouped-alias pattern that captures the prefix `(alias)` plus each `(alias)` member inside the `(tuple)` and emits one IMPORTS per `prefix.member`. (The Scala query already has a brace-splitter for the analogous `import a.{b, c}` form via the `@import.stmt` capture, but it is gated to Scala and never sees the Elixir `@import.source` capture.)

core-ingestion/src/queries.ts
```diff
@@ -1164,6 +1164,66 @@ export const SCALA_QUERIES = `
...
+
+; ── Qualified calls: Foo.bar() — captures qualifier so resolution is possible ─
+(call
+  target: (dot
+    left: (alias) @_qualifier
```
josephismikhail, line 1460: **High: qualified calls (`Repo.insert`) never resolve cross-file and become phantom nodes.**

This query correctly captures the qualifier, but resolution can't use it. A call `Repo.insert(x)` yields dstName `Repo.insert`, qualifier `Repo`. The defining module, however, is stored under its FULL dotted name: `defmodule MyApp.Repo` is captured as `@name = "MyApp.Repo"`, so the defining file's qualified key is `MyApp.Repo.insert` and its symbol set holds `MyApp.Repo`, never the short alias `Repo`. So in resolveEdges Tier-1b, neither `fileDefinesQualifiedMember(fp, "Repo", "insert")` nor `fileHasSymbol.has("Repo")` ever matches, and the qualified branch `continue`s without falling through to the bare-name tiers. The edge is left unresolved and (on this pre-`external://` base) buildPatchWithResolution falls back to `nodeId(filePath, "Repo.insert")` -- a same-file **phantom** node. So qualified calls, the cleanest signal Elixir offers, currently resolve worse than the inline comment here claims: they neither link to the real function nor drop, they spawn phantoms.

Compounding it: Elixir `alias`/`alias ... as:` does not populate `importAliases` (this directive query captures only `@import.source`, no `@import.alias`), so there is no short-qualifier to full-module translation available.

Fix: for Elixir, translate the call qualifier through the alias map to the full module name, match it to a module entity whose name ends with `.Repo` (or equals it), then resolve `insert` among that module's CONTAINS children, emitting dstQualifiedKey `MyApp.Repo.insert`.

Related (low): Erlang-style atom-qualified calls (`:ets.new(...)`, `:erlang.foo()`) are dropped entirely here because the qualifier is an atom, not an `(alias)`.

core-ingestion/src/index.ts
```diff
@@ -1525,6 +1527,25 @@ export function parseFile(filePath: string, source: string): FileParseResult | n
           if (isDecoratorCall) continue;
         }
 
+        // Elixir: skip the function-name node inside def/defp/defmacro(p) argument lists.
+        // `def foo(args)` parses as call{target:def, args:[call{target:foo, args:[...]}]}.
+        // callName.node is the identifier `foo`; its parent is the inner call node,
+        // whose parent is `arguments`, whose parent is the outer def* call.
+        // Without this, the inner call{target:foo} would emit a `foo CALLS foo` self-edge.
+        if (language === SupportedLanguages.Elixir) {
```
josephismikhail, line 1615: This self-edge skip is correct for the plain `def foo(args)` shape (inner call -> arguments -> def call), but it only fires when `argsNode.type === 'arguments'`. For a guarded clause `def foo(x) when g(x)` the inner head call's parent is a `binary_operator` (the `when`), so the skip does not fire and a bogus `<enclosing> CALLS foo` / `CALLS g` edge leaks (see the queries.ts def-query comment for the full picture, since the guarded function is also not captured as an entity at all).

Fix: walk through an intervening `binary_operator` here, e.g. if `argsNode` is a `binary_operator` whose parent is `arguments`, use that grandparent's parent as the def call before the `def|defp|defmacro|defmacrop` check.

Separately, a nearby precision note (not this line): an Elixir function's qualifiedKey is `container.name` with no arity, so multiple clauses of the same name in one module collapse to a single node keyed by name, and buildPatch keeps the FIRST occurrence, so the surviving node's line span covers only the first clause (later clauses' spans are lost) and `changeset/2` vs `changeset/3` are conflated. Consider spanning first-clause-start to last-clause-end, or including arity in the key.

core-ingestion/package.json
```diff
@@ -21,6 +21,7 @@
     "tree-sitter-c": "^0.21.0",
     "tree-sitter-c-sharp": "^0.21.0",
     "tree-sitter-cpp": "^0.22.0",
+    "tree-sitter-elixir": "^0.3.5",
```
josephismikhail: **Medium: move `tree-sitter-elixir` to `optionalDependencies`.** It is the only grammar in `dependencies`; every other native grammar (kotlin, swift, and the in-flight sas) is in `optionalDependencies` (just below, L35). Native grammars can fail to build when no prebuild matches the platform/ABI. In `dependencies`, that failure aborts `npm ci` for the **entire** core-ingestion package, taking down every language, not just Elixir. The runtime loads it via `tryLoadGrammar` (try/catch), so a missing grammar already degrades gracefully; `optionalDependencies` is the correct home and matches the established pattern.

josephismikhail (commented):
Built this branch in Docker (a node:22 harness, a merge-base baseline at `ff7df41`, and a fresh **main** memory-layer image, all from source), ran the suite, and ingested `elixir-ecto/ecto` (148 files) into the main memory-layer.

**Solid:** 8/8 Elixir tests pass; the full core-ingestion suite has no new failures vs the merge-base baseline (the 6 failures are pre-existing Scala/Rust/TS snapshot drift); ecto ingests cleanly (5937 nodes, 13051 edges, exit 0). Modules/functions/macros get the right kinds, the Ecto.Query macro DSL is captured, and the `def foo()` self-edge fix works on the happy path.

**The concern is what the tests do not cover.** A grammar-level review (verified against tree-sitter-elixir 0.3.5) found that several pervasive idiomatic Elixir constructs are silently mishandled, and none of them surface as a crash or a failed test, so the ecto ingest "looked" healthy while actually missing data. The most important is that guarded functions (`def f(x) when ...`) are dropped entirely. Details inline, roughly by severity. None are regressions (this is all new code). Happy to send a fix PR.

Recommendation: the Elixir test suite needs cases for guards, qualified-call resolution, multi-alias, and module attributes before this is trustworthy on real repos.

Minor/FYI: `defstruct` and Erlang atom-qualified calls (`:ets.new`) are uncaptured (low priority), and the `ix-cli/package-lock.json` change (+5/-249) is benign (no dropped deps, no version drift) but is unrelated churn bundled into the PR.

josephismikhail:
Re-ran the parser at `48c179f` against probe modules to check the items from the last review. Where things stand:

### Confirmed fixed (verified by running the parser, not just reading)

- **Guarded defs** (`def f(x) when g`) are now captured as functions, and the bogus self/guard CALLS edges are gone. On a guarded `fetch/1` calling `validate/1`, the only CALLS emitted are the real ones (no `fetch -> fetch`, no `fetch -> is_integer`). 
- **`@behaviour` / `@impl` / `@derive`** no longer emit phantom `CALLS behaviour|impl|derive` edges.
- **Single-alias qualified calls resolve cross-file.** `alias MyApp.Repo` + `Repo.insert(x)` resolves to `MyApp.Repo.insert` (conf 0.9). This was the big one.
- **`tree-sitter-elixir` moved to `optionalDependencies`.**
- The negative-assertion tests you added are exactly what catch this class of regression. 

### Still open

**1. Grouped alias `alias MyApp.{User, Repo}` is only half-handled.**

The grouped-alias query captures the members as **bare** names (`User`, `Repo`), not the full `MyApp.User` / `MyApp.Repo` that the single-alias form produces. Two consequences I reproduced:

- The IMPORTS edge / module entity is created for bare `Repo`, which doesn't match the real `MyApp.Repo` node (the single-alias form correctly yields `MyApp.Repo`). So the two alias forms point at different targets.
- Grouped members never populate `importAliases`, so a qualified call through a grouped alias does **not** resolve:

```elixir
alias MyApp.Repo             # Repo.insert(x)    -> resolves to MyApp.Repo.insert  (good)
alias MyApp.{Account, Post}  # Account.lookup(x) -> phantom (unresolved)
```

Fix direction: in the grouped-alias query also capture the prefix `(alias)` and emit `prefix.member` (so `MyApp.Account`), and add the grouped members to `importAliases` the same way the single `alias` directive does in `index.ts`.

**2. Multi-clause / arity collapse** (the sub-note from the guarded-def thread).

`qualifiedKey` is `container.name` with no arity, so multiple clauses of one function collapse to a single node. Now that guards are captured, this hits more real code. Repro:

```elixir
def handle(x) when is_atom(x), do: x
def handle(x, y), do: {x, y}
def handle(x), do: x
```

produces 3 entities all keyed `MyApp.Multi.handle`, which fold into one node: only the first clause's line span survives, and `handle/2` vs `handle/1` conflate. Consider spanning first-clause-start to last-clause-end, or including arity in the key.

**Nit:** the Elixir `importAliases` population block in `index.ts` is duplicated verbatim (two identical `if (!importAlias && language === ... Elixir ...)` blocks). Harmless (idempotent) but worth collapsing to one.

The criticals are all resolved, so this is close. The grouped-alias capture is the main remaining correctness gap.

josephismikhail (approved):
Review items addressed and verified extensively (span-merge approach): grouped-alias full module names + cross-file resolution (incl. nested grouped aliases and bound short names), multi-clause functions merge into one node spanning all clauses (def vs defmacro kept distinct via kind), deduped importAliases block, guards/@behaviour/@impl handled. Verified: 13/13 Elixir unit tests; 421 real files (Oban+ecto+phoenix) parse with 0 crashes and 2855 cross-file calls resolved; fully deterministic; robust on adversarial inputs; no regressions. Per-arity function identity (changeset/2 vs /3) is tracked as a follow-up.

Issue #206, opened 2026-05-08
[feature] feat: add `--llm` output format for token-optimized LLM consumption


KageBinary:
## Problem

The two existing output formats do not serve LLM consumers well.

- `--format text`
  - Human-readable
  - Includes decorative headers, spacing, alignment padding, prose, and separators
  - Wastes tokens in agent workflows

- `--format json`
  - Machine-readable
  - Structurally verbose
  - Nested metadata, repeated keys, null fields, quotes/braces, and full object trees consume significantly more tokens than necessary

AI coding tools like Cursor, OpenCode, Claude Code, and Codex call `ix` dozens or hundreds of times per session. At ~400–1200 tokens per JSON response, this adds up extremely quickly.

Today the plugins themselves spend large amounts of code filtering, flattening, summarizing, and reshaping ix JSON responses before returning them to the LLM.

---

## Proposed Solution

Add a global:

```bash
--llm

josephismikhail:
## Concrete spec proposal (v2)

The original issue cut off mid-paragraph, so this is a concrete proposal to react to rather than a finished plan. Pushback welcome on any of it. (Replaces an earlier draft.)

### Measured today (against a ~3,300-file backend)

| Command | text | json | json/text |
|---|---:|---:|---:|
| `smells` | 98,178 | 296,938 | 3.0× |
| `subsystems --list` | 5,376 | 11,939 | 2.2× |
| `imports <file>` | 1,742 | 6,679 | 3.8× |
| `rank --by dependents --top 10` | 315 | 989 | 3.1× |
| `stats` | 349 | 1,417 | 4.1× |
| `search <term> --limit 10` | 18 | 245 | 13.6× |
| `inventory --kind class --limit 20` | 2,352 | 1,793 | 0.76× |

JSON is consistently 2-4× larger than text on tree-structured output and dominates the context cost on `smells`, `subsystems`, `impact`, `overview`, and `map`. These are also the commands AI agents call repeatedly per session.

### Design: flag surface vs renderer implementation

Splitting these two questions removes most of the per-command debate:

- **Flag surface.** The `--format llm` value is accepted on **every** command that already accepts `--format`. AI consumers (ix-claude-plugin, Cursor extension, future tools) pass it unconditionally without maintaining a lookup table of "which commands support it." Cost of declaring the option: ~3 lines per command.
[...]

PR #232, opened 2026-06-03
feat(cli): add `--format llm` token-optimized output (PR 1, #206)

josephismikhail:
PR 1 of the rollout in #206. Introduces the `--format llm` output convention and the first six renderers, capturing the bulk of the byte savings up front. Adopters (ix-claude-plugin etc.) opt in separately; the flag is additive and nothing else changes.

## What's in this PR

- **Shared renderer** (`ix-cli/src/cli/llm.ts`): newline-delimited `key=value` lines with a leading record-kind token. Drops null/empty fields, quotes+escapes values containing space/`=`/`"`/`\`/control chars (newlines encoded as `\n`, so a record never spans lines), and a uniform `error code=… message="…"` helper.
- **Renderers (6):** `stats`, `smells`, `subsystems`, `map`, `impact`, `overview` (Tier 1 + the `stats` headline example). Hierarchies are emitted flat with explicit `parent=<id>` so consumers re-tree from `id`/`parent=` alone.
- **Uniform flag surface:** `--format llm` is advertised on every `--format`-capable command (except deprecated `query`). Commands without a hand-written renderer fall through to `text` (the most compact existing format), so consumers can pass the flag unconditionally without a per-command lookup.
- **Docs:** `docs/llm-format.md` documents the wire format.
- **Tests:** unit coverage for the helper (quoting/escaping/omission edge cases) and the `map`/`subsystems` renderers.

## Example

```
$ ix stats --format llm
nodes total=98979 method=49180 module=38199 class=6833 file=3285
edges total=354283 CALLS=177418 CONTAINS=57163 IMPORTS=38199
```

## Verification

- `tsc --noEmit` clean.
- New llm + renderer tests pass (38/38 across llm/renderers/format suites). Full suite: 453 pass, 1 pre-existing failure (`resolver-type-preference.test.ts`) that also fails on the untouched `origin/main` base, unrelated to this change.
- Live e2e against a backend was not run: the local memory-layer endpoint is unreachable in this environment. Renderer logic is covered by unit tests with synthetic data.

## Follow-ups (per #206 rollout)

- PR 2: Tier 2 renderers (`inventory`, `rank`, `depends`, `trace`, `contains`, `callers`, `callees`, `imports`/`imported-by`).
- PR 3: Tier 3 (`search`, `text`, `history`, `patches` — `stats` shipped here).
- PR 4: Tier 4 (`entity`, `locate`, `diff`, `conflicts`).
- Adoption tracked in ix-infrastructure/ix-claude-plugin#21.


josephismikhail:
### Live e2e verification (local memory-layer)

Ran the renderers against a real local memory-layer (`localhost:8090` + ArangoDB), populating the graph via in-process `ix map`. All six renderers produce correct record streams on real data:

- `stats`, `map`, `subsystems --list`, `overview`, `impact`, `smells` — all emit valid `key=value` records; flat `parent=<id>` hierarchy confirmed on `map`/`subsystems`.
- Byte savings on a 103-file map: `map` llm≈2950 vs json 3461; `subsystems --list` llm 2097 vs json 2262; `stats` llm 28 vs json 110. (Ratios are modest here because the JSON paths are already compacted and 36-char region UUIDs dominate both formats — shortening region ids is a worthwhile future optimization.)

**Bug found and fixed during e2e** (commit b913f8e): on a workspace's first `ix map`, `bootstrap()` printed the `Ix / Registered workspace …` setup notice to **stdout**, corrupting any machine-readable output (`json` was affected too). Those side-effect notices now go to stderr, and `map` treats `llm` like `json` (quiet ingest, no progress bar/timing on stdout). Verified `ix map --format llm` stdout now starts cleanly with the `map files=…` record.

`tsc --noEmit` clean. Test suite: 453 pass, 1 pre-existing failure (`resolver-type-preference.test.ts`) that also fails on the untouched base.

josephismikhail:
### Re-ran e2e on an isolated backend (not the shared dev instance)

To avoid sharing state with concurrent work, stood up a dedicated memory-layer + ArangoDB as a separate compose project (`ix-ml-e2e`, ports 18090/8531, own volume + network, schema v3). Populated it via in-process `ix map` and re-ran the renderers there. Corrected, clean byte measurements on real data:

| command | llm | json | reduction |
|---|---:|---:|---:|
| `stats` | 191 | 1000 | 81% |
| `smells` | 822 | 2352 | 65% |
| `subsystems --list` | 2097 | 2262 | 7% |

`stats` matches the ~80% reduction predicted in #206. `subsystems` is modest because 36-char region UUIDs dominate both formats (shortening region ids is a worthwhile follow-up). All six renderers emit valid records; `map`/`subsystems` flat `parent=<id>` hierarchy confirmed; `ix map --format llm` stdout verified clean (first line is the `map files=…` record, banner now on stderr per b913f8e).

josephismikhail:
### Closed the four review gaps (commit 3c2bccb)

1. **Structured errors across all six commands.** Every llm command now emits a uniform `error code=<slug> message="..."` record on stdout, not just human prose on stderr. `impact`/`overview` emit `unresolved_target` on resolution failure; `map`/`stats`/`smells`/`subsystems` emit `backend_error` on request failure. Verified live: `ix impact ThisDoesNotExist999 --format llm` → `error code=unresolved_target message="No entity resolved for \"ThisDoesNotExist999\"."` on a clean stdout.
2. **All six renderers now have unit tests.** Extracted the inline rendering into exported pure functions returning `string[]` (`renderStatsLlm`, `renderSmellsRunLlm`/`renderSmellsListLlm`, `impactHeaderLlm`/`impactPropagationLlm`/`impactTailLlm`, `overviewContainerLlm`/`overviewLeafLlm`, `renderScopedSubsystemLlm`, `renderSubsystemErrorLlm`) and covered each.
3. **Regression test for the stdout fix.** Extracted `emitSetupNotice()` and added a test asserting it writes to stderr and never stdout.
4. **Zero-noise.** Drop `chunks_per_file` / `chunks` fields when 0 (uncomputed, no signal). `subsystems --list` no longer emits `chunks_per_file=0` on every row.

`tsc --noEmit` clean. Full suite: 466 pass, 1 pre-existing failure (`resolver-type-preference`) unrelated to this branch. All re-verified on the isolated `ix-ml-e2e` backend.

PR #326, opened 2026-08-05
chore: raise Node floor to 22 to match chalk 6

josephismikhail:
Follow-up to #318 (chalk 5.6.2 -> 6.0.0).

chalk 6's **only** breaking change is `engines.node` -> `>=22`. That bump landed while `ix-cli/package.json` still declared `>=20.0.0`, and six other places advertised Node 20 support — so the package required a runtime the project documented as unsupported.

## What changed

| File | Change |
|---|---|
| `ix-cli/package.json` | `engines.node` `>=20.0.0` -> `>=22.0.0` |
| `ix-cli/package-lock.json` | root `engines` kept in sync |
| `scripts/install/install.sh` | `NODE_MIN_MAJOR=20` -> `22` |
| `scripts/install/install.ps1` | `$NodeMinMajor = 20` -> `22` |
| `scripts/build-cli.sh` | version guard + both error messages |
| `scripts/dev/setup.sh` | install hint |
| `README.md` | stated prerequisite (2 places) |
| `CONTRIBUTING.md` | stated prerequisite |

## Why this is low risk

Nothing in the tested or packaged paths moves:

- CI already runs Node **22 and 24** only (`ci.yml` matrix).
- `Formula/ix.rb` already pins `node@22`.
- There is no `engine-strict` anywhere, so this was a warning rather than a hard failure.

The exposure it closes is the **curl-installer path**, which admitted Node 20 and builds from source. Those users would have hit `EBADENGINE` warnings on a configuration the README explicitly promised to support.

Worth noting: chalk 6 would in fact still *run* on Node 20 — its only imports are `node:process`, `node:os` and `node:tty`, with no Node 22-only APIs in its shipped source. The `>=22` bound is upstream EOL policy, not a technical requirement. Raising our floor is the honest resolution rather than pinning chalk back, since **Node 20 reached end-of-life in April 2026**.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,7 @@
 
 ### Prerequisites
 
-- Node.js 20+
+- Node.js 22+
 - Docker and Docker Compose
 
 ### Local Setup
--- a/README.md
+++ b/README.md
@@ -126,12 +126,12 @@ irm https://raw.githubusercontent.com/ix-infrastructure/ix-cursor-plugin/main/in
 
 The install script sets up everything for you on macOS and Linux. It checks for and installs anything that is missing:
 
-- Node.js 20 or newer
+- Node.js 22 or newer
 - Git
 - ripgrep (powers `ix text`)
 - Docker and Docker Compose (for the local backend)
 
-All you need beforehand is a terminal with `curl` (or `wget`). On Windows, install Node.js 20+ and Docker Desktop first, then run the installer.
+All you need beforehand is a terminal with `curl` (or `wget`). On Windows, install Node.js 22+ and Docker Desktop first, then run the installer.
 
 Works on macOS, Linux, and Windows, on both x86-64 and arm64.
 
--- a/ix-cli/package-lock.json
+++ b/ix-cli/package-lock.json
@@ -30,7 +30,7 @@
         "vitest": "^4.1.0"
       },
       "engines": {
-        "node": ">=20.0.0"
+        "node": ">=22.0.0"
       }
     },
     "node_modules/@babel/helper-string-parser": {
--- a/ix-cli/package.json
+++ b/ix-cli/package.json
@@ -13,7 +13,7 @@
     "ix": "dist/cli/main.js"
   },
   "engines": {
-    "node": ">=20.0.0"
+    "node": ">=22.0.0"
   },
   "scripts": {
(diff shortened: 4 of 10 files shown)


josephismikhail:
Reviewed by an independent agent, plus two follow-up passes. Completeness checks out: a repo-wide sweep for `>=20`, `node@20`, `node:20-*`, `.nvmrc`/`.node-version`/`volta`, workflow `node-version:` pins and `engines` blocks found **no remaining Node 20 floor**. CI already runs 22/24 only, `Formula/ix.rb` already pins `node@22`, and the lockfile's `packages[""].engines` matches `package.json`. Merges clean onto current `main`.

One real gap fixed here: **`$NodeMinMajor` was dead code**, so bumping it to 22 changed nothing. The Windows installer's Node check was existence-only — a user on Node 20 passed the installer cleanly, got "Ix is ready", and only met the floor later when the CLI's own runtime guard rejected their first `ix` command (the `ix --version` smoke test at the end is non-fatal, so it warned and carried on). `install.sh` has always enforced its floor, and its Windows branch also errors rather than auto-installing, so this brings the two in line rather than inventing new behavior. Nobody this newly rejects could run `ix` anyway.

The version parse went through two rounds. Using `-match` + `$Matches` was unsafe: with an array left-hand side `-match` becomes the *filter* operator and does not populate `$Matches`, so multi-line output either threw `Cannot index into a null array` or silently reused a stale `$Matches` from the caller's scope — reachable because the README ships this as `irm ... | iex`, which executes in the caller's session. Seeded `$Matches[1]=99` and confirmed the original form accepted a Node 20 machine. Now uses `[regex]::Match` on `@(& node -v)[0]`, which also preserves `$LASTEXITCODE` (a `| Select-Object -First 1` pipeline loses it on 7 and sets `-1` on 5.1). Full matrix — v22/v24/nightly/no-`v` prefix/exact floor accept; v20/v21/empty/no-output/garbage/multi-line reject — verified on both 5.1 and 7, no input throwing and none wrongly accepted.

Two follow-ups **not** taken here, worth separate issues:
- `ix-cli/package.json` still pins `@types/node` at `^20.19.43`, so the CLI requires Node 22 at runtime but typechecks against Node 20's API surface. `.github/dependabot.yml` ignores its major, so it will never auto-advance.
- The two macOS `mktemp` templates in `install.sh` (`.pkg`/`.dmg`) randomize nothing under BSD `mktemp` and yield predictable `/tmp` paths.

PR #329, opened 2026-08-08
fix(install): diagnose GHCR "denied" instead of blaming Docker Hub

### LINKED ISSUE #328: [bug] Windows installer fails because the backend image is not publicly pullable.
## Description

The Windows installer fails while starting the Ix backend because the `ix-memory-layer` Docker image cannot be pulled from GHCR.

The installer successfully detects Node.js and Docker, downloads the Compose configuration, and then fails when pulling:

`ghcr.io/ix-infrastructure/ix-memory-layer:latest`

A direct `docker pull` of the same image also returns `denied`.

## Steps to reproduce

1. Open PowerShell on Windows.
2. Run:

   ```powershell
   irm https://ix-infra.com/install.ps1 | iex
   ```
3. Wait for the installer to start the backend.

The issue can also be reproduced directly with:

```powershell
docker pull ghcr.io/ix-infrastructure/ix-memory-layer:latest
```

## Expected behavior

The `ix-memory-layer` image should be pulled successfully from GHCR and the Ix backend should start normally.

## Actual behavior

Docker cannot pull the `ix-memory-layer` image and returns:

```text
Error response from daemon: error from registry: denied
```

As a result, the installer fails while starting the backend.

## CLI output

```text
Ix Installer
Version: 0.8.1

-- Node.js --
[ok] Node v24.18.0

-- Docker --
[ok] Docker installed
[ok] Docker running

-- Backend --
Downloading compose...
[ok] Compose ready

Starting backend...
[+] up 2/2
✘ Image ghcr.io/ix-infrastructure/ix-memory-layer:latest Error error from registry: denied
! Image arangodb:3.12 Interrupted

Error response from daemon: error from registry: denied

[error] Docker compose failed
```

Direct pull:

```text
PS C:\Users\Win 10> docker pull ghcr.io/ix-infrastructure/ix-memory-layer:latest
Error response from daemon: error from registry: denied
```

## Environment

* OS: Windows
* Node: v24.18.0
* Ix version: Installer 0.8.1
* Docker: Docker Desktop, installed and running

KageBinary:
Closes #328 (installer side).

## What #328 actually is

The report says the backend image "is not publicly pullable." It is. I verified the whole anonymous pull path against GHCR with no credentials:

| step | result |
|---|---|
| anonymous token exchange for `ix-infrastructure/ix-memory-layer` | `200` |
| `manifests/latest` (OCI index) | `200` |
| `linux/amd64` image manifest | `200` |
| first layer blob | `200` |

The package is public, `latest` resolves to `1.0.13`, and both `linux/amd64` and `linux/arm64` are published.

The real cause is almost always **stale `ghcr.io` credentials**. If Docker has any saved `ghcr.io` login, it sends those credentials instead of pulling anonymously, and if the token is expired or revoked GHCR replies `denied` rather than falling back. An invalid credential against this same public image returns:

```
{"errors":[{"code":"DENIED","message":"denied"}]}
```

which is exactly the `Error response from daemon: error from registry: denied` in the report. The user-side fix is `docker logout ghcr.io`.

## The two bugs this fixes

**`install.sh` gave the opposite of the fix.** It matched `denied` but folded it into the Docker Hub rate-limit branch, so this user would have been told to run `docker login` — which cannot help, and if anything entrenches the bad credential. Now split into two branches keyed on the actual error.

**`install.ps1` had no diagnosis at all** — it printed a bare `[error] Docker compose failed`, which is why this took an issue to resolve. It now captures compose output with `Tee-Object` (still streamed live to the console) and applies the same diagnosis.

## Verification

Both branches were exercised against the reporter's verbatim log and a Docker Hub rate-limit log:

| input | routes to |
|---|---|
| reporter's log from #328 | GHCR branch → `docker logout ghcr.io` |
| `toomanyrequests ... pull rate limit` | Docker Hub branch → `docker login` |

The `sh` and PowerShell matchers were checked to agree on both. `sh -n scripts/install/install.sh` passes.

`$ErrorActionPreference` is pinned to `Continue` across the native docker call so that `2>&1` redirection cannot turn docker's stderr into a terminating error on PS 5.1/7, and restored in a `finally`. `$LASTEXITCODE` still reflects docker's exit code through the `Tee-Object` pipeline.

## Note for a separate PR

`scripts/install/install.sh:462` has an unrelated pre-existing bug that shellcheck flags as an error:

```sh
_fetch https://get.docker.com | sh < /dev/null   # SC2259
```

The `< /dev/null` overrides the pipe, so `sh` reads its script from `/dev/null` and the Docker auto-install silently does nothing while appearing to succeed. Left alone here to keep this PR scoped.


josephismikhail:
Reviewed by an independent agent, plus two follow-up passes on the fixes. Three defects found and fixed on the branch; everything below was verified by execution on Windows PowerShell 5.1 and 7, not just read.

**Blocking — the box characters made `install.ps1` unparseable on Windows PowerShell 5.1.** The file is UTF-8 without a BOM, so 5.1 decodes it with the system ANSI code page. Every box character contains byte `0x94`, which is U+201D in CP1252 — a smart closing quote, which PowerShell accepts as a string delimiter. Inside `Write-Host` literals that terminated the strings mid-line: **3 parse errors, against 0 on `main`**, where the box characters live only in comments. That bricks `install.cmd` (`WebClient.DownloadString` ignores the `charset=utf-8` header and decodes as CP1252 — confirmed against the live endpoint) and any run from disk. Nothing executed at all — no Node check, no Docker check, no CLI install.

Adding a BOM looked like the obvious fix and is wrong: it breaks the documented `irm ... | iex` path on both 5.1 and 7, because the leading U+FEFF binds to the first token (`The term '#' is not recognized`). Switched the box art to ASCII instead — verified clean under UTF-8, CP1252, CP437 and CP850, with every line the same width so the boxes stay aligned.

**Blocking — the `429` matcher reintroduced the bug this PR exists to fix.** A bare `429` also matches ordinary pull output: byte counts like `429.4MB/1.02GB` and hex layer IDs. Since the rate-limit branch is tested first, a GHCR denial during the ~1GB `arangodb` pull routed to the Docker Hub box and told the user to `docker login` — the exact wrong advice. Narrowing it then went too far and dropped a real Docker Hub message (`error parsing HTTP 429 response body: ... "Too Many Requests (HAP429)."`, the edge limiter answering 429 with a non-JSON body), which carries none of the other keywords. Final pattern adds `http 429`; checked against six real rate-limit spellings and five non-rate-limit lines, with `sh` and PowerShell agreeing on every one.

**`install.ps1` had no `else`**, which is what turned a missed pattern into total silence — `install.sh` has always printed the log tail. Added the equivalent.

Also: `2>&1` into `Tee-Object` made 5.1 render docker's progress as red `NativeCommandError` blocks on a **successful** install; stringifying before the tee fixes that with `$LASTEXITCODE` and the log contents intact. And `uninstall.ps1` had the identical decode hazard (two em dashes in `Write-Host` strings, 6 parse errors under CP1252) — two characters, fixed here since it is the same defect class.

Verified `$LASTEXITCODE` really does survive the `Tee-Object` pipeline, so the failure branch is live rather than dead code.

josephismikhail (approved):
Independent agent review plus two follow-up passes; two blocking defects found and fixed on the branch (install.ps1 unparseable on Windows PowerShell 5.1 under CP1252, and a 429 matcher that misrouted GHCR denials). CI green, verified on 5.1 and 7.

PR #331, opened 2026-08-08
fix(install): Docker auto-install silently did nothing on Linux

KageBinary:
Found while working #328. Unrelated to that bug, so it's split out here.

## The bug

`scripts/install/install.sh:462`

```sh
_fetch https://get.docker.com | sh < /dev/null
```

The `< /dev/null` redirect binds to `sh` and **overrides the pipe**. `sh` reads its script from `/dev/null`, does nothing, and exits `0`.

So on any Linux machine without Docker, the installer printed `Installing Docker Engine via get.docker.com...`, installed nothing, reported success, and then failed further down at the compose step — with nothing indicating Docker had never been installed.

Demonstrated:

```console
$ printf 'echo "DOCKER INSTALL RAN"\n' | sh < /dev/null
$ echo $?
0        # no output, and success
```

## Why the redirect is there

It's deliberate everywhere else in this script. The installer is normally run as `curl -fsSL https://ix-infra.com/install.sh | sh`, so the installer's own source is on stdin; every command gets `< /dev/null` to stop it consuming the rest of the script. That's correct — it's only wrong when the command being protected is *itself* on the receiving end of a pipe.

The Homebrew call at line 193 does the same job safely, because it uses command substitution rather than a pipe:

```sh
/bin/bash -c "$(_fetch https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)" < /dev/null
```

This line was the only place the two patterns collided.

## The fix

Download to a temp file, then run it with stdin still pointed at `/dev/null`. The pipe is no longer load-bearing, so the redirect and the script no longer contend for stdin.

A failed download or a failed install now aborts with an actionable message pointing at the manual install docs, rather than being swallowed. Previously **both** were silent. The temp file is removed on every path.

## Verification

| path | result |
|---|---|
| happy path | script executes, reaches post-install, no temp left |
| download fails | aborts with actionable error, no temp left |
| install script exits 7 | aborts with actionable error, no temp left |

Each failure case stops before the post-install step, confirming the abort works.

- `sh -n scripts/install/install.sh` passes
- shellcheck error-level findings: **1 → 0** (this was SC2259, the file's only one)

## Note

This changes behavior: a genuine Docker install failure now stops the installer instead of continuing silently. That's the intent — the current silence is what makes this bug hard to spot from the user side — but it is a behavior change worth a look during review.

### DIFF
--- a/scripts/install/install.sh
+++ b/scripts/install/install.sh
@@ -509,7 +509,11 @@ install_docker() {
           err "Could not download the Docker Compose plugin.
   Install it manually: https://docs.docker.com/compose/install/linux/"
         fi
-        sudo install -m 0755 "$compose_bin" /usr/local/lib/docker/cli-plugins/docker-compose
+        if ! sudo install -m 0755 "$compose_bin" /usr/local/lib/docker/cli-plugins/docker-compose; then
+          rm -f "$compose_bin"
+          err "Could not install the Docker Compose plugin.
+  Install it manually: https://docs.docker.com/compose/install/linux/"
+        fi
         rm -f "$compose_bin"
       fi
       ;;
@@ -739,7 +743,7 @@ else
 
     # Pull images first so we catch rate limits / auth errors clearly
     echo "  Pulling Docker images (this may take a few minutes on first run)..."
-    PULL_LOG=$(mktemp /tmp/ix-pull-XXXXXX.log)
+    PULL_LOG=$(mktemp /tmp/ix-pull-XXXXXX) || err "Could not create a temporary file for the image pull log."
     dc -f "$COMPOSE_DIR/docker-compose.yml" pull < /dev/null >"$PULL_LOG" 2>&1 &
     PULL_PID=$!
     while kill -0 "$PULL_PID" 2>/dev/null; do
@@ -774,7 +778,7 @@ else
 
     # Start services
     printf "  Starting backend services..."
-    START_LOG=$(mktemp /tmp/ix-start-XXXXXX.log)
+    START_LOG=$(mktemp /tmp/ix-start-XXXXXX) || err "Could not create a temporary file for the startup log."
     if ! dc -f "$COMPOSE_DIR/docker-compose.yml" up -d < /dev/null >"$START_LOG" 2>&1; then
       echo ""
       echo "  Failed to start backend. Error output:"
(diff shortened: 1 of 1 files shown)


josephismikhail:
Reviewed by an independent agent, plus two follow-up passes on the fixes. The root-cause analysis in the description holds up — `cmd | sh < /dev/null` really does make `sh` read from `/dev/null` and exit 0, confirmed empirically and by shellcheck (SC2259, the file's only error-level finding, now gone). Two defects found and fixed on the branch.

**`mktemp /tmp/get-docker-XXXXXX.sh` fails on Alpine.** BusyBox passes the template straight to `mkstemp()`, which requires it to end in `XXXXXX`; the `.sh` suffix gives `EINVAL`. The failure was also unhandled, so under `set -eu` the installer died with a bare `mktemp: Invalid argument` and no indication of which step failed. Dropped the suffix (`sh FILE` ignores the extension) and handled the failure. GNU accepts both forms, so there is no regression on Debian/Ubuntu/Fedora/RHEL — that half I ran; the BusyBox half is from source (`coreutils/mktemp.c` + musl's `mkstemp`), since no BusyBox was available here.

That fix also didn't reach far enough on its own: `PULL_LOG` and `START_LOG` carry the identical template and sit on the **universal** Step 3 path, so an Alpine user just died one step later with the same message. Both fixed too. Left the macOS `.pkg`/`.dmg` templates alone deliberately — BSD `mktemp` doesn't reject them, and `installer -pkg`/`hdiutil attach` may care about the extension, so they need a different fix (`mktemp -d` plus a named file inside). Worth a follow-up: on macOS those templates randomize *nothing* and produce a literal, predictable `/tmp` path.

**`sudo _download ...` could never work.** `_download` is a shell function, and `sudo` execs a binary, so functions do not survive the exec — this line could only ever produce `sudo: _download: command not found`. shellcheck flagged it as SC2032/SC2033 on `main`. It is reached whenever the compose plugin is absent after the engine install; before this PR it was hit on *every* fresh Linux install, so that path always died there. Now downloads as the current user and places the binary with `sudo install -m 0755`, with the surrounding `sudo mkdir` and `sudo install` both guarded so a sudo-less user gets the manual-install URL instead of a bare sudo error.

shellcheck error-level findings across the stack: 1 on `main` → 0, with no new findings introduced.

josephismikhail (approved):
Independent agent review plus two follow-up passes; two blocking defects found and fixed on the branch (non-portable mktemp template with unhandled failure, and sudo invoking a shell function). CI green, shellcheck error-level 1 -> 0.

PR #333, opened 2026-08-08
fix(cli): render backend-down as guidance instead of an undici stack trace

KageBinary:
## The bug

`main.ts` ended in a bare `program.parse()` inside an async IIFE — no try/catch, no process-level handlers. Commander's `parse()` does not await async action handlers, so a rejection escaping a command floated to Node's default handler.

What that looks like on the single most common first-run failure — install the CLI, forget to start the backend:

```console
$ ix inventory --kind function
node:internal/deps/undici/undici:14976
      Error.captureStackTrace(err);
            ^
TypeError: fetch failed
    at node:internal/deps/undici/undici:14976:13
    at async IxClient.post (/home/<user>/.../ix-cli/src/client/api.ts:520:18)
    at async Command.<anonymous> (/home/<user>/.../ix-cli/src/cli/commands/inventory.ts:67:19) {
  [cause]: Error: connect ECONNREFUSED 127.0.0.1:8090
```

Node internals, the user's absolute install path, and no indication of what to do about it.

After:

```console
$ ix inventory --kind function

  Ix backend not reachable at http://localhost:8090.

  Next
  Start it with `ix docker start`, then check `ix status`.
```

## This wires up code that was already here

`errors.ts` has contained a complete, well-written error renderer since March with **zero call sites**. `renderCliError`, `parseBackendError`, `renderStructuredError`, `CliUsageError` and `CliResolutionError` were all dead. This PR mostly connects them rather than writing anything new.

- **`parseAsync` + try/catch**, plus `unhandledRejection` / `uncaughtException` handlers as a backstop for anything that escapes a command's own catch.
- **`isBackendUnreachable`** classifies transport failures. It checks both the error and the `cause` that Node's fetch nests the real error under, so "cannot reach the backend" stays distinct from "the backend returned an error" — the latter still renders the backend's own message and `next` field.
- The generic path now routes through **`formatFetchError`**, so a bare `fetch failed` surfaces its transport cause instead of saying nothing.
- **`IX_DEBUG=1`** appends stack traces for anyone who wants the old detail.

Endpoint resolution reads config off disk, so it's wrapped in a `try` — an error renderer must not be able to throw.

## One non-obvious change

`checkForUpdate()`'s floating promise (`upgrade.ts:134`) never had a `.catch()`. That was survivable before; it isn't now. It is deliberately not awaited, so without the catch a failed background version check would reach the new `unhandledRejection` handler and abort an otherwise successful command. Added.

## Verification

Reproduced against a dead endpoint (`IX_ENDPOINT=http://127.0.0.1:59999`) before and after.

| check | result |
|---|---|
| `ix inventory` / `patches` / `conflicts` / `show` emit `node:internal` | before: yes → **after: 0 occurrences** |
| `--version` | `0.6.1`, unchanged |
| `--help` | unchanged |
| unknown command | `error: unknown command 'boguscmd'`, exit 1, unchanged |
| missing required option | `error: required option '--kind <kind>' not specified`, unchanged |
| test suite | 536 → **550 passing** (14 new, 41 files) |
| `tsc --noEmit` | clean |
| `eslint src` | 0 errors; 39 pre-existing warnings, none in the touched files |

The new `errors.test.ts` pins the behaviour that matters — that the backend-down path contains `ix docker start` and does **not** contain `node:internal` or `at async` — so this cannot silently regress.

## Deliberately left for a follow-up

Seven commands (`reset`, `map`, `watch`, `stats`, `init`, `subsystems`, `smells`) catch backend errors themselves and print `Error: fetch failed` via their own local handler. That's unhelpful, but it is not a crash and not a leak, and routing them through the shared renderer touches seven more files. Worth its own PR now that `isBackendUnreachable` and `renderCliError` exist to route them to.

### DIFF
--- a/ix-cli/src/cli/__tests__/errors.test.ts
+++ b/ix-cli/src/cli/__tests__/errors.test.ts
@@ -19,6 +19,20 @@ function fetchFailure(code: string): Error {
 }
 
 /**
+ * The same shape for a socket that was established and then died — Node records
+ * the failing phase in `syscall`, which is what separates "never got there"
+ * from "was working, then dropped".
+ */
+function midFlightFailure(code: string): Error {
+  const err = new TypeError("fetch failed");
+  (err as { cause?: unknown }).cause = Object.assign(
+    new Error(`read ${code} 127.0.0.1:8090`),
+    { code, errno: -104, syscall: "read" },
+  );
+  return err;
+}
+
+/**
  * renderCliError writes to stderr and then exits. It reaches stderr two ways —
  * process.stderr.write directly, and console.error via renderStructuredError —
  * so both are captured here. Exit is turned into a throw so asserting on the
@@ -84,11 +98,20 @@ describe("isBackendUnreachable", () => {
   });
 
   it("does not claim unreachability for a connection dropped mid-flight", () => {
-    // A healthy backend behind the k8s ingress resets idle connections — see the
-    // 5-minute signal in client/api.ts. Calling that "backend not started" sends
-    // the user to `ix docker start` for a backend that is up and serving.
-    expect(isBackendUnreachable(fetchFailure("ECONNRESET"))).toBe(false);
-    expect(isBackendUnreachable(fetchFailure("ETIMEDOUT"))).toBe(false);
+    // A backend behind an ingress that drops idle connections is up and serving.
+    // Calling that "not started" sends the user to `ix docker start` for nothing.
+    expect(isBackendUnreachable(midFlightFailure("ECONNRESET"))).toBe(false);
+    expect(isBackendUnreachable(midFlightFailure("ETIMEDOUT"))).toBe(false);
+    // undici's "other side closed" is only ever an already-established socket.
+    expect(isBackendUnreachable(fetchFailure("UND_ERR_SOCKET"))).toBe(false);
+  });
+
+  it("still claims unreachability when the connect attempt itself failed", () => {
+    // `connect ETIMEDOUT` is a SYN nobody answered — firewall, VPN off, wrong
+    // host. That user needs the guidance, so the phase is what decides, not the
+    // code alone.
+    expect(isBackendUnreachable(fetchFailure("ETIMEDOUT"))).toBe(true);
+    expect(isBackendUnreachable(fetchFailure("ECONNRESET"))).toBe(true);
   });
 });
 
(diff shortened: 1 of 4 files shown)


josephismikhail (approved):
Independent agent review of the PR, plus three further independent reviews across the correction rounds.

The core of the PR is right and was verified rather than taken on trust: `parseAsync` genuinely fixes the leak (commander's `parse()` discards the promise `parseAsync` awaits), ~14 commands with no local try/catch previously leaked stack traces and now render cleanly, commander's own exit codes for `--version`/`--help`/unknown-command/missing-option are untouched (no `exitOverride` anywhere, so those never reach the new catch), there is no double-rendering, and no path exits 0 on failure or hangs.

Four defects found and fixed:

1. **The PR's own `unhandledRejection` handler turned a latent bug fatal, and the `.catch()` it added did not cover it.** That catch guards the inner fetch chain; `checkForUpdate()`'s own promise covers the synchronous cached-read path. Reproduced on the PR head: `ix config get endpoint` prints the right answer, then exits 1 with `latest.split is not a function`. Fixed at the call site, and the root cause fixed in `readCache` — which needed a second pass, since validating only `latest` left `compassLatest`/`backendLatest` reaching `isNewer()` and throwing the identical error.

2. **`ix docker start` was advised for any unreachable endpoint.** Pro syncs `config.endpoint` to a cloud instance, where that advice cannot help and starts a backend the user is not pointed at. Now conditional on a loopback endpoint; verified live on both branches.

3. **`IX_DEBUG=1` printed `err.stack`, which is empty for a fetch `TypeError`** — the frames are lost across the async boundary and the whole diagnostic lives in `err.cause`. Debug mode emitted one useless line for the exact failure this PR exists to handle. Now walks the cause chain, emitting only stack, message and code — deliberately not `util.inspect`, which dumps every own property of whatever was thrown, and this is the process-wide boundary.

4. **Transport classification.** Took three rounds and ended simpler than it started. `ECONNRESET`/`ETIMEDOUT` are out as genuinely mid-flight. `UND_ERR_SOCKET` is in: it reads like an established socket, but measured against a real socket, accept-then-FIN is what a booting or dead container behind a published port produces — the fresh-install case. An intermediate `syscall === "connect"` split was removed as inert: the default endpoint is a hostname, and Happy Eyeballs collapses per-address failures into an AggregateError carrying `code` but not `syscall`, and undici's 10s connect timeout preempts the OS one anyway with a code already in the set.

554 passing (from 536 on main), tsc clean, eslint 0 errors, no new warnings. Each added test was mutation-checked — reverting its production change fails it.

Known follow-up, out of scope and flagged by the author: seven commands (`stats`, `reset`, `init`, `subsystems`, `smells`, `watch`) still print bare `Error: fetch failed` from their own handlers, so identical failures render two different ways depending on the command. Worth routing through `renderCliError` now that it exists. `readCache` also has no test coverage — there is no upgrade test file at all.

PR #334, opened 2026-08-08
fix: remove raw NUL bytes that made the two largest sources unsearchable

KageBinary:
## The bug

Two files carried literal NUL bytes, used as separators in composite `Set` keys inside template literals:

```ts
const k = `${pkg}<NUL>${consumerNodeId}`;     // ingest.ts:544
```

| file | raw NULs | lines | `file` reports |
|---|---|---|---|
| `core-ingestion/src/index.ts` | 3 | 3,665 | `data` |
| `ix-cli/src/cli/commands/ingest.ts` | 4 | 1,558 | `data` |

A single NUL makes a file binary as far as grep, `git grep` and ripgrep are concerned — and they then skip it **silently**. No match, no warning, exit 0.

These are the two largest files in the repo. A contributor grepping for a symbol defined in either one gets nothing back and reasonably concludes it doesn't exist. It also degrades the product itself: `ix text` shells out to ripgrep (`text.ts:37`), so **ix could not search its own source**.

This is not theoretical — it repeatedly produced empty results during a codebase audit before the cause was found.

## The fix

Replaced each raw byte with the escape **already used elsewhere in the same file**, so each file stays internally consistent:

- `ingest.ts` → the unicode escape, matching its own line 1470 (`const sep = ""`)
- `index.ts` → the hex escape, matching its own lines 812 and 820

## This is a no-op at runtime

Verified rather than assumed — the escaped forms and a raw NUL produce the same string:

```
uni escape === raw NUL : true
hex escape === raw NUL : true
separator charCode     : 0 0
identical lengths      : true
```

Only the on-disk encoding of the literal changes.

## Verification

| check | before | after |
|---|---|---|
| `file` on both | `data` | `JavaScript source, Unicode text, UTF-8` |
| `grep -c stitchConsumeSeen ingest.ts` | 0 (skipped silently) | **3** |
| `grep -c seenEnv index.ts` | 0 (skipped silently) | **3** |
| raw NULs in any tracked `.ts` | 7 | **0** |
| `tsc --noEmit` (both packages) | clean | clean |
| ix-cli suite | 536 passing | 536 passing |
| core-ingestion suite | 6 failed / 271 passed | 6 failed / 271 passed |

The core-ingestion failures are **pre-existing on `main`** and unrelated to this change — I ran the suite on `main` and on this branch and diffed the failing test list, which is byte-identical. (They're the `memory-layer/` fixture tests and two query tests; separate issue, and that suite isn't run by CI at all today.)

## Regression guard

Added a check to the fast `static` gate. It needs no toolchain, so it runs before install and fails in seconds with a per-file GitHub annotation:

```
::error file=<path>::N raw NUL byte(s) — use the \x00 escape instead; a raw NUL
makes this file binary to grep/ripgrep, which silently skips it
```

Confirmed both directions: passes on the fixed tree, and fails when a NUL is deliberately reintroduced.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,21 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false
+      # A raw NUL byte makes a file "binary" to grep/ripgrep, which then skips it
+      # silently — including `ix text`, which shells out to rg. Two of the
+      # largest sources in this repo carried them and were invisible to search.
+      # Needs no toolchain, so it runs first and fails in seconds.
+      - name: No raw NUL bytes in tracked sources
+        run: |
+          status=0
+          for f in $(git ls-files '*.ts' '*.tsx' '*.js' '*.mjs' '*.cjs'); do
+            n=$(tr -dc '\000' < "$f" | wc -c)
+            if [ "$n" -gt 0 ]; then
+              echo "::error file=$f::$n raw NUL byte(s) — use the \\x00 escape instead; a raw NUL makes this file binary to grep/ripgrep, which silently skips it"
+              status=1
+            fi
+          done
+          exit $status
       - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
           node-version: "22"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,15 +30,28 @@ jobs:
       # largest sources in this repo carried them and were invisible to search.
       # Needs no toolchain, so it runs first and fails in seconds.
       - name: No raw NUL bytes in tracked sources
+        # `shell: bash` for -o pipefail: without it the `tr | wc` pipeline below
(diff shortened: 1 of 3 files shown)


josephismikhail (approved):
Independent agent review of the PR, plus a second independent review of the fix commit.

The NUL-byte diagnosis and the escape swap are correct and provably a runtime no-op — the escapes were verified byte-identical at each of the six sites, all plain template literals with no `String.raw`/regex/JSON context, and the mixed `\x00`/`\u0000` style matches each file's own existing convention. The motivation reproduces: ripgrep silently skipped the whole file, exit 0, nothing on stderr.

One defect found and fixed in 17d172b: the CI guard shared the failure mode it exists to prevent. `for f in $(git ls-files ...)` word-splits on IFS, and with no `pipefail` the failed redirect still let `wc -c` print 0 — scoring the file clean. Demonstrated with a tracked `nul component.ts` containing a real NUL: old guard exit 0, new guard exit 1. Also added a `checked == 0` assertion so the guard can't pass vacuously if the pathspecs ever stop matching, and covered .mts/.cts/.jsx.

Verified: guard clean on this tree (162 files), exit 1 on a reintroduced NUL including one in a path with a space; binary assets correctly excluded; `static` still gates the required `CI Passed` context.

PR #335, opened 2026-08-08
fix(compass): stop shipping an empty compass/ and unblock the repair path

KageBinary:
`ix view` has been broken for **every fresh OSS install** since at least v0.7.0. Three independent failures stacked — any one alone was fatal.

## 1. The published tarball ships an empty directory

Checked against the real published asset, not the source:

```console
$ tar tzf ix-0.8.1-linux-amd64.tar.gz | grep -i compass
ix-0.8.1-linux-amd64/compass/          # the only entry
$ tar tzf ix-0.8.1-linux-amd64.tar.gz | grep -c 'compass/.'
0                                       # zero files
```

## 2. `COMPASS_TOKEN` is set but invalid — and three `continue-on-error` layers hid it

The secret has not been rotated since 2026-03-28. Every recent release log (v0.7.0, v0.8.0, v0.8.1) shows:

```
remote: Invalid username or token. Password authentication is not supported...
fatal: Authentication failed for '.../system-compass.git/'
##[error]Process completed with exit code 128
##[warning]No files were found with the provided path: compass-dist.
```

`continue-on-error` swallowed the exit 128 **and** the workflow's own deliberate `::error::Compass build did not produce dist/index.html` guard. Result: `conclusion: success`, every job green, compass-less tarballs published on all four platforms, zero signal to anyone. The compass job completed in **9 seconds** — impossible for a clone plus `npm ci` plus a build.

## 3. The installer poisoned the one working repair path

`install.sh` wrote `$IX_HOME/cli/compass/.version` **without downloading a single compass byte** (there is no compass download anywhere in the installer). `ix upgrade` has a correct, working downloader — but it is gated on `isNewer(latest, stamp)`, saw `0.2.0 → 0.2.0`, and never fired.

The comment above it said the quiet part out loud:

```sh
# -- Stamp installed versions so upgrade checker doesn't nag --
```

So a fresh install produced a `compass/` directory containing exactly one file — `.version` — and `view.ts` looks for `index.html`.

There is a perverse detail: if the version API call failed (offline, rate-limited), no stamp was written and `ix upgrade` then worked correctly. **Failure produced the working outcome.**

> This also explains why it works on maintainer machines: `ix upgrade` does `rmSync(~/.ix/cli)` *before* the compass block, which wipes the stamp, so `compassCurrent` becomes `0.0.0` and the download fires. You get compass by upgrading through a CLI version bump. A fresh installer run can never reach that state.

## What this PR does

Rotating `COMPASS_TOKEN` is the one step that cannot be done in code — a fine-grained PAT with `Contents: read` on `ix-infrastructure/system-compass`. **This is the token-independent half**: it makes the failure loud instead of silent, and recoverable instead of permanent.

**`release.yml`**
- Drop `continue-on-error` from the compass build. A failed compass now fails the release — which the existing `needs.compass.result != 'failure'` gate already knows how to handle.
- A missing or rejected token is a hard error naming the exact fix, instead of `exit 0`. Clone output is piped through `sed` to redact the token.
- Assert `compass-dist/index.html` exists after the copy, before upload.
- Packaging refuses to publish when the bundle is absent (`REQUIRE_COMPASS`, false only on PR dry-runs, which legitimately skip the compass job).

**`install.sh` / `install.ps1`** — only stamp `.version` when `compass/index.html` is actually present, and warn when a build shipped without compass instead of silently pretending it is installed.

**`upgrade.ts`** — `getInstalledCompassVersion()` treats a stamp with no `index.html` beside it as `0.0.0`, at all three comparison sites. This is the belt-and-braces part: even users already carrying a poisoned stamp self-heal on their next `ix upgrade`, without reinstalling.

**`view.ts`** — point users at `ix upgrade` rather than "reinstall Ix", which would not have helped: a reinstall reproduces the same empty directory.

## Verification

Simulated the exact broken install — a `.version` stamp and nothing else:

| | `ix upgrade --check` says |
|---|---|
| before | `[ok] Compass already on the latest version (0.2.0)` |
| after | `Compass update available: none → 0.2.0` |

And end to end with the bundle in place:

```console
$ ix view --no-open --port 8791
  [ok] Visualizer started
  http://localhost:8791
  scope: workspace "ix-work"

GET /  -> HTTP 200
title  -> <title>System Compass - Systems | Ix Compass</title>
```

`release.yml` is valid YAML, `install.sh` passes `sh -n`, `tsc --noEmit` clean, `eslint` 0 errors, 536 tests passing.

## Note for the maintainer

Until `COMPASS_TOKEN` is rotated, **this PR will make the next release fail rather than silently ship broken**. That is the intent — but it means the token needs rotating before cutting v0.9.0.

Worth deciding separately: `ix-compass-dist` is a **public** repo and its built bundle is anonymously downloadable (verified: HTTP 200, 584 KB, 22 files, the full UI). The source in `system-compass` is private, but the compiled UI is not. If the built UI is also meant to be private, that repo is the gap — though note `ix upgrade`'s repair path depends on it being reachable.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -78,9 +78,14 @@ jobs:
             echo "::error::COMPASS_TOKEN is not set. It is required to build System Compass from the private ix-infrastructure/system-compass repo. Add a fine-grained PAT with Contents:read on that repo as the COMPASS_TOKEN secret."
             exit 1
           fi
-          # Never interpolate the token into a traceable command.
+          # Do not pipe this clone through a redaction filter. This job runs on
+          # the runner's default `bash -e {0}`, which has no `pipefail`, so the
+          # pipeline would exit with the filter's status (0) and the guard below
+          # could never fire — the failure would surface as a confusing `cd:
+          # system-compass: No such file or directory` two lines down instead of
+          # the message telling you to rotate the token. The runner already masks
+          # `secrets.*` in log output, so the token in this URL is redacted anyway.
           git clone "https://x-access-token:${COMPASS_TOKEN}@github.com/ix-infrastructure/system-compass.git" 2>&1 \
-            | sed "s/${COMPASS_TOKEN}/***/g" \
             || {
               echo "::error::Could not clone ix-infrastructure/system-compass. COMPASS_TOKEN is set but was rejected — it is most likely expired or lacks Contents:read on that repo. Rotate it and re-run."
               exit 1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -65,17 +65,26 @@ jobs:
         with:
           node-version: "22.14"
 
+      # No continue-on-error. It previously swallowed the clone's exit 128 *and*
+      # the explicit ::error:: guard below, so an expired COMPASS_TOKEN produced
+      # a fully green release that shipped an empty compass/ directory and a
+      # broken `ix view` for every user. A compass failure must fail the release.
       - name: Build System Compass
-        continue-on-error: true
         env:
           COMPASS_TOKEN: ${{ secrets.COMPASS_TOKEN }}
         run: |
           mkdir -p compass-dist
           if [ -z "$COMPASS_TOKEN" ]; then
-            echo "COMPASS_TOKEN not set — skipping compass build (tarballs ship without compass)"
-            exit 0
+            echo "::error::COMPASS_TOKEN is not set. It is required to build System Compass from the private ix-infrastructure/system-compass repo. Add a fine-grained PAT with Contents:read on that repo as the COMPASS_TOKEN secret."
+            exit 1
           fi
-          git clone https://x-access-token:${COMPASS_TOKEN}@github.com/ix-infrastructure/system-compass.git
+          # Never interpolate the token into a traceable command.
+          git clone "https://x-access-token:${COMPASS_TOKEN}@github.com/ix-infrastructure/system-compass.git" 2>&1 \
+            | sed "s/${COMPASS_TOKEN}/***/g" \
+            || {
+              echo "::error::Could not clone ix-infrastructure/system-compass. COMPASS_TOKEN is set but was rejected — it is most likely expired or lacks Contents:read on that repo. Rotate it and re-run."
+              exit 1
+            }
           cd system-compass
           npm ci --legacy-peer-deps
           npm run build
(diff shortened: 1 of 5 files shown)


josephismikhail (approved):
Independent agent review of the PR, plus a second independent review of the fix commits.

The three-part diagnosis holds up and the fail-closed design is right. Verified independently: the job graph does what the description claims — a failed compass skips `build`, which skips `release`, so there is no half-published release and no unsatisfiable `needs`; `REQUIRE_COMPASS` reaches all four matrix legs and evaluates `true` for every real release and `false` only on `pull_request`; the `index.html` assertion is unreachable past a failed copy. The "three comparison sites" claim for `getInstalledCompassVersion()` was checked by enumerating call sites — accurate, none missed. I also confirmed the `ix-compass-dist` tarball puts `index.html` at the root after `--strip-components=1`, so the 0.0.0 fallback cannot cause a re-download loop.

Two defects found and fixed in ec07d5d:

1. The `::error::` telling you to rotate COMPASS_TOKEN — the point of the change — could never fire. The `compass` job has no `shell: bash`, so it runs under the default `bash -e {0}` with no `pipefail`; piping the clone through `sed` gave the pipeline sed's exit status of 0, leaving `|| { ... }` unreachable and `set -e` untripped. An expired token would have surfaced two lines later as `cd: system-compass: No such file or directory`. Dropping the pipe was verified safe: git withholds the credential from its own error output on every failure mode this step hits, and neither bash nor the PAT format introduces an encoding step that could defeat the runner's automatic `secrets.*` mask. f642e43 adds an explicit `::add-mask::` so that guarantee no longer depends on where the value is sourced from.

2. The new warning claimed "System Compass was not included in this build", which is false on every Windows install — `Expand-Archive` and `unzip` do not strip the top-level directory, so a good bundle lands in `cli\ix-<version>-windows-amd64\compass\`. Corrected the message rather than repointing the check: repointing would stamp a version for a bundle `ix view` can never read, re-creating the poisoned stamp this PR removes.

Two follow-ups, neither blocking:
- **COMPASS_TOKEN must be rotated before the next release.** This PR makes the failure loud, and the tag is already pushed when it fires (recovery: rotate, delete the tag, re-push).
- **The Windows bundle is still stranded.** Every Windows user pays a redundant `ix upgrade` download to get a working `ix view`. The clean fix is moving the extracted `compass\` into `cli\compass\` after extraction — worth its own PR.

PR #337, opened 2026-08-08
fix(upgrade): stop `ix upgrade` destroying the CLI on Windows

KageBinary:
**Release blocker.** This is dormant today only because there is nothing newer than v0.8.1 to upgrade to. Tagging v0.9.0 turns `checkForUpdate`'s *"Update available — Run: `ix upgrade`"* notice into a brick for every Windows user, so it wants fixing before the tag rather than after.

## Two faults, either one fatal

**1. Delete-then-extract, with a tool Windows doesn't have.**

```ts
rmSync(installDir, { recursive: true, force: true });   // deletes ~/.ix/cli
mkdirSync(installDir, { recursive: true });
execFileSync("unzip", [...]);                           // no fallback
```

`unzip` is not on stock Windows — it exists only with Git Bash/MSYS. The delete lands, the extract throws, and the catch prints `Failed to extract CLI update` and exits **with `~/.ix/cli` already gone**. There's a `tar` branch right below it, but that's the POSIX path.

**2. The launcher is never repointed.**

`install.ps1:276` writes `%IX_HOME%\bin\ix.cmd` with a **version-encoded** path:

```
"%~dp0..\cli\ix-$Version-windows-amd64\ix.cmd" %*
```

Upgrade *did* refresh a shim — but at `~/.local/bin/ix`, the bash shim `install.sh` writes under Git Bash. `install.ps1` never creates that file (zero references to `.local/bin` in the entire script). So a PowerShell-installed user ends up with a launcher pointing into the version directory the upgrade just deleted.

## The fix

- **Stage, verify, then swap.** Extract into a staging dir beside the install, confirm the CLI entry point is actually there, and only then replace. Staging lives under `IX_HOME` so the swap is a same-filesystem rename.
- **The swap is recoverable.** The old install is moved aside rather than deleted, and restored if the rename fails part-way — there is no window where the user has nothing.
- **`extractZipOnWindows` tries every extractor a Windows box might have:** bsdtar (ships with Windows 10 1803+ and reads zip), then PowerShell `Expand-Archive` (always present), then `unzip` (MSYS shells where it may be the only one). It reports what it tried when all fail.
- **Both shims are rewritten** — `%IX_HOME%\bin\ix.cmd` and the bash shim — so either install shape is repointed. The nested directory name is read back from disk rather than assumed, so a naming change in the archive can't silently break the launcher.

## Verification

Simulated both sequences against a real filesystem:

| scenario | old | new |
|---|---|---|
| extraction fails (no `unzip`) | **BROKEN — CLI destroyed** | install preserved |
| archive missing entry point | — | install preserved |
| swap fails mid-way | — | previous install restored |
| success | — | swapped, backup removed |

8 new tests cover archive-shape resolution and all four swap outcomes. **564 tests passing**, `tsc --noEmit` clean, `eslint` 0 errors.

## Note

This does not change the on-disk layout, so it stays compatible with what `install.ps1` produces today. Flattening the Windows extraction to match POSIX (`--strip-components`) would let the shim stop encoding a version at all — worth doing, but it changes the installer contract, so it's deliberately not in this PR.

### DIFF
--- a/ix-cli/src/cli/__tests__/upgrade-archive-shape.test.ts
+++ b/ix-cli/src/cli/__tests__/upgrade-archive-shape.test.ts
@@ -1,0 +1,117 @@
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { mkdtempSync, mkdirSync, writeFileSync, rmSync, existsSync, renameSync } from "node:fs";
+import { join } from "node:path";
+import { tmpdir } from "node:os";
+
+import { soleChildDir } from "../commands/upgrade.js";
+
+let root: string;
+
+beforeEach(() => {
+  root = mkdtempSync(join(tmpdir(), "ix-upgrade-test-"));
+});
+
+afterEach(() => {
+  rmSync(root, { recursive: true, force: true });
+});
+
+describe("soleChildDir", () => {
+  it("finds the single nested directory a Windows release zip extracts to", () => {
+    // Windows zips wrap everything in ix-<version>-<platform>/; the launcher
+    // shim has to point inside it, so the name is read back rather than assumed.
+    const nested = join(root, "ix-0.9.0-windows-amd64");
+    mkdirSync(nested, { recursive: true });
+    expect(soleChildDir(root)).toBe(nested);
+  });
+
+  it("ignores loose files beside the directory", () => {
+    const nested = join(root, "ix-0.9.0-windows-amd64");
+    mkdirSync(nested, { recursive: true });
+    writeFileSync(join(root, "README.txt"), "x");
+    expect(soleChildDir(root)).toBe(nested);
+  });
+
+  it("returns null when the layout is already flat (POSIX tarball)", () => {
+    // --strip-components=1 leaves several top-level dirs, so there is no single
+    // child to descend into and the caller must use the directory as-is.
+    mkdirSync(join(root, "cli"), { recursive: true });
+    mkdirSync(join(root, "core-ingestion"), { recursive: true });
+    mkdirSync(join(root, "compass"), { recursive: true });
+    expect(soleChildDir(root)).toBeNull();
+  });
+
+  it("returns null for an empty or missing directory", () => {
+    expect(soleChildDir(root)).toBeNull();
+    expect(soleChildDir(join(root, "does-not-exist"))).toBeNull();
+  });
+});
+
+describe("upgrade install swap", () => {
+  const entry = (base: string, ver: string) =>
+    join(base, "cli", `ix-${ver}-windows-amd64`, "cli", "dist", "cli", "main.js");
+
+  function seedInstall(home: string, ver: string) {
+    mkdirSync(join(home, "cli", `ix-${ver}-windows-amd64`, "cli", "dist", "cli"), { recursive: true });
+    writeFileSync(entry(home, ver), `// ${ver}\n`);
+  }
+
+  it("leaves the existing install intact when extraction fails", () => {
+    // The regression this guards: the old sequence deleted the install dir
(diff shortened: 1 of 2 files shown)


josephismikhail:
Reviewed this ahead of the version push and pushed a follow-up commit (eab6037) rather than blocking it — the staging/swap approach is right, but a few paths could still leave the CLI unusable, three of them on Windows specifically.

**The one that matters most.** The cleanup `rmSync`s were inside the same `try` as the two renames:

```ts
renameSync(stagingDir, installDir);                     // swap is DONE here
rmSync(backupDir, { recursive: true, force: true });    // still inside try
```

`rmSync` defaults to `maxRetries: 0`, and `backupDir` is the tree this process was running from moments earlier — an AV or indexer handle is enough to throw. The upgrade has already succeeded at that point, but control goes to the catch, the rollback is skipped (its guard is `!existsSync(installDir)`, false by then), and it exits **before the shim refresh**. `ix.cmd` is left pointing at the version directory the swap just deleted. Now extracted as `swapInStagedTree`, with cleanup past the failure boundary.

**Two more Windows brick paths:**

- The new `ix.cmd` embedded an absolute path written with `"ascii"`. Node's ascii encoding masks non-ASCII code units instead of failing, so `C:\Users\José\...` silently becomes a path that doesn't exist — and the write succeeds, so it reports `[ok]`. This was a regression: install.ps1's shim is relative (`%~dp0..\cli\...`) and never embedded the profile path. Reverted to that form.
- `install.sh`'s `pick_bin_dir()` prefers `/usr/local/bin` when writable and *deletes* the `~/.local/bin` copy (line 853). Refreshing only `~/.local/bin` recreated the one it removed and left the live shim stale. Now refreshes whichever exists — and only on Windows, since the POSIX shim is `exec "$INSTALL_DIR/ix"` and carries no version, so it never goes stale.

**Also:** `Expand-Archive`'s per-file errors are non-terminating and powershell.exe exits 0 for them, so a partial extraction read as success and the `unzip` fallback never ran — now runs under `$ErrorActionPreference='Stop'`, each attempt starts from an empty destination and must produce something. The staged CLI is now actually executed before the old one is deleted (`existsSync` on main.js can't see a truncated tree). Extractor stderr is captured instead of discarded. A dev shim from `scripts/dev/setup.sh` is left alone, matching what the @ix/pro refresh already does. `sweepUpgradeOrphans` reclaims `.cli-staging-*`/`.cli-backup-*` and restores the install if a run died between the renames — that window leaves no CLI at all, so nothing in-process can recover it.

**On the tests:** the swap cases hand-rolled their own `mkdirSync`/`renameSync`/`rmSync` sequence, so they passed regardless of what the shipped code did. I pointed them at `swapInStagedTree`/`sweepUpgradeOrphans` and mutation-checked them — reintroducing the cleanup-inside-try bug, dropping the rollback, or disabling the sweep each fail the suite now; none of them failed the originals.

**Left alone:** two concurrent `ix upgrade` runs can still interleave between the renames and leave the loser's rollback disabled. Closing that needs a lock file, which felt like more than this fix should carry — the orphan sweep at least reclaims the abandoned backup now. Happy to split it out if you want it before the release.

Full suite green locally (566 passed) plus lint and typecheck.

josephismikhail (approved):
Separate-agent review run (xhigh, 51 agents) plus fixes applied in eab6037 / 97a3527. Nine confirmed defects closed, including three fresh ways to brick the Windows CLI: cleanup inside the swap's failure boundary, the lossy `"ascii"` shim path, and refreshing the wrong bash shim. Tests now drive the real `swapInStagedTree`/`sweepUpgradeOrphans` and were mutation-checked against three reintroduced regressions. Concurrent-upgrade race knowingly left open and flagged on the PR. 18/18 checks green.

PR #338, opened 2026-08-08
perf(release): stop shipping the build toolchain in the release tarball

KageBinary:
The packaging step copies `node_modules` verbatim after a plain `npm ci`, so every release tarball carries the full dev toolchain:

```sh
cp -rL ix-cli/dist ix-cli/node_modules ix-cli/package.json "$STAGING/cli/"
```

Measured on this repo:

| | size | note |
|---|---|---|
| `ix-cli/node_modules` | 88 MB | |
| └ runtime deps | **4 MB** | exactly 3 packages: `chalk`, `commander`, `yaml` |
| └ direct devDependencies | 48 MB | typescript alone is 23 MB |
| `core-ingestion` direct devDependencies | 34 MB | |

None of it is executed by the CLI. It's downloaded, extracted, and left on every user's disk. The published v0.8.1 assets are 72–91 MiB; the dry-run artifact I pulled today was 114 MB.

## The change

One prune step between build and packaging. `npm run build` has already consumed the toolchain by then, and everything downstream only copies or scans files.

**`--omit=dev` keeps `dependencies` and `optionalDependencies`**, so all 27 tree-sitter grammars survive — 13 required plus 14 optional, ~417 MB of genuine runtime deps the parser cannot work without. Only devDependencies go. I verified the grammar declarations before writing this; none of them are devDependencies.

## Two things fall out beyond the size

**The SBOM becomes accurate.** The SBOM step scans `path: ix-cli`, so until now it described the build toolchain as shipped content. Pruning first means it describes what actually ships.

**The dismissed advisories become true.** Advisories against `vitest`, `esbuild`, `postcss` and friends were auto-dismissed as *development-scope* — accurate for the repo, but not for the tarball, where those packages were landing on user disks. Now they genuinely don't.

## Safety

The step prints before/after sizes, so the real saving shows up in the release log rather than being asserted here. It then runs:

```sh
node ix-cli/dist/cli/main.js --version
```

If pruning ever removes something load-bearing, the release fails at that line instead of shipping a broken tarball. Verified locally that this command works against a built tree (`0.6.1`).

`release.yml` is valid YAML. The full effect is visible on the next dry run.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -193,6 +193,29 @@ jobs:
           npm run build
           npm version "$VERSION" --no-git-tag-version --allow-same-version
 
+      # Drop the build toolchain before anything reads node_modules. `npm run
+      # build` above has already consumed it, and every later step only copies
+      # or scans files, so nothing downstream needs it.
+      #
+      # --omit=dev keeps `dependencies` AND `optionalDependencies`, so all 27
+      # tree-sitter grammars (13 required + 14 optional) survive — they are
+      # genuine runtime deps and the parser is useless without them. Only
+      # devDependencies go.
+      #
+      # This also fixes the SBOM below, which scans `ix-cli` and until now
+      # described the build toolchain as shipped content.
+      - name: Prune dev dependencies before packaging
+        run: |
+          size() { du -sm "$1" 2>/dev/null | cut -f1 | head -1; }
+          before_cli=$(size ix-cli/node_modules)
+          before_ing=$(size core-ingestion/node_modules)
+          (cd ix-cli && npm prune --omit=dev)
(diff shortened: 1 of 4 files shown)


josephismikhail (approved):
Separate-agent review run (xhigh, 45 agents, combined with 339/340/341) plus fixes applied in 0a57b24. Verified by building that dist shipped 41 test modules / 35 vitest imports into the tarball this step prunes vitest from — now zero, via a build-only tsconfig so typecheck still covers tests. Smoke test extended to actually parse with core-ingestion (confirmed it exits 1 on a removed grammar). `du` pipeline no longer able to abort the release under -eo pipefail. 23/23 green.

PR #339, opened 2026-08-08
fix(install): tell Intel Mac users the truth, and stop citing a script that does not exist

KageBinary:
Intel Macs resolve to `darwin-amd64`, which `release.yml` **deliberately never builds** — the matrix comment explains why (macos-13 runners queue indefinitely and would stall every release). The installer doesn't know that, so it attempts the download, gets a 404, and reports:

```
This likely means the release asset has not been uploaded yet.
Check available releases at: ...
```

That's wrong, and it implies waiting will fix it. It won't. Meanwhile `README.md:160` promised:

> Works on macOS, Linux, and Windows, on both x86-64 and arm64.

## Two bugs in that one error path

**1. Intel Macs get a misleading diagnosis.** Now detected before the download, with a route that actually works. `Formula/ix.rb` builds from source (`npm install` + `npm run build`, `depends_on "node@22"`) and carries no architecture restriction — so Homebrew is a genuine answer for Intel, not a consolation prize:

```
  [!!] Ix does not publish a pre-built CLI for Intel Macs (darwin-amd64).

  Install with Homebrew instead — it builds from source and works on Intel:
    brew tap ix-infrastructure/ix https://github.com/ix-infrastructure/Ix
    brew install ix

  Apple Silicon Macs are supported by this installer directly.
```

**2. The generic failure path cited a script that doesn't exist.** It told users:

```sh
git clone https://github.com/ix-infrastructure/Ix.git
cd Ix && ./setup.sh          # <- not in this repo, and never has been
```

Replaced with the Homebrew instructions. The tap command is taken from `homebrew/README.md` and checked against `Formula/ix.rb` rather than invented — there is no `homebrew-tap` repo in the org, so the `brew install <org>/tap/ix` shorthand would not have worked.

## README

Corrected to state what is actually published, and to point Intel users at Homebrew.

## Verification

`sh -n` passes; shellcheck reports **0 findings at both error and warning level**. Simulated the `darwin-amd64` branch to confirm the message renders and the script aborts before attempting a download that cannot succeed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,10 @@ The install script sets up everything for you on macOS and Linux. It checks for 
 
 All you need beforehand is a terminal with `curl` (or `wget`). On Windows, install Node.js 22+ and Docker Desktop first, then run the installer.
 
-Works on macOS, Linux, and Windows, on both x86-64 and arm64.
+Works on macOS, Linux, and Windows. Pre-built CLI packages are published for
+Apple Silicon macOS, Linux (x86-64 and arm64) and Windows x86-64. **Intel Macs**
+have no pre-built package — install with Homebrew, which builds from
+source: `brew tap ix-infrastructure/ix https://github.com/ix-infrastructure/Ix && brew install ix`.
 
 For the full list, including the endpoints the installer reaches and the directories it creates, see [docs/prerequisites.md](docs/prerequisites.md).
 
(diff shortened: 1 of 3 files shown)


josephismikhail (approved):
Separate-agent review run (xhigh, combined with 338/340/341) plus fixes applied in 5b873bc. The review caught a destructive ordering bug I had missed on my own pass: the guard sat after `rm -rf "$INSTALL_DIR"` (line 879), so an Intel Mac user re-running the installer had their working CLI deleted and was then told to use Homebrew. Guard moved above that branch and softened to preserve an existing install and continue to the backend setup; no longer fires when IX_VERSION is set, so pinned rollbacks to pre-#259 releases still work. Download-failure advice is platform-aware again (citing scripts/dev/setup.sh, since root ./setup.sh genuinely never existed). docs/prerequisites.md no longer contradicts the README that links to it. 22/22 green.

PR #342, opened 2026-08-08
docs(claude): mark Pro-only commands and steer agents to --format llm

KageBinary:
`CLAUDE.md` isn't documentation — it's the instruction payload injected into user repos and **executed by an agent**. Three things in it were wrong in ways that cost every OSS user tokens on every session.

## 1. Pro-only commands were unmarked — including in MANDATORY RULES

**Rule 2** (`ix decide`, after *every* design decision) and **Rule 6** (`ix truth add`, whenever the user states a goal) are both Pro-only. So an OSS user's assistant burns a tool call and gets:

```
The 'decide' command requires Ix Pro.
```

after every architecture decision, **forever**, with nothing telling it to stop trying.

It wasn't just the rules:

| section | Pro rows | marked? |
|---|---|---|
| High-Level Workflow Commands **(Preferred)** | 5 of 9 | ❌ none |
| History & Decisions | 9 of 12 | ❌ none |
| Semantic Boundaries | all 3 | ❌ none |
| Planning (Pro) / Workflows (Pro) | all | ✅ at heading level |

Every command in the file is now checked against the CLI's **actual registration** — 37 OSS + 13 Pro stubs, enumerated by loading `register/oss.ts` and reading `program.commands`, not by trusting any doc. Pro rows carry a `[Pro]` marker, and one instruction up top tells the agent what the failure looks like and to stop retrying for the rest of the session.

Verified mechanically after the edits:

```
unmarked Pro: 0   nonexistent commands: 0   OSS wrongly marked: 0   malformed table rows: 0
```

## 2. "All commands support `--format json`" was false

`config`, `ingest`, `init`, `reset`, `upgrade`, `view` and `watch` declare no `--format` option at all. An agent taking that sentence literally passes a flag those commands reject. Replaced with the real split between query and action commands.

## 3. `--format llm` appeared **zero** times

This is the token-minimal format built specifically for agents. Per `docs/llm-format.md`:

> typically cutting response bytes 2-4x versus `json` on tree- and table-shaped output
>
> It is accepted on every command that accepts `--format` … so consumers can pass the flag unconditionally without a per-command lookup.

The file agents actually read never mentioned it, while telling them to use `json` throughout — so every `ix` call an agent makes has been 2–4× more expensive than it needed to be. This is the change with the largest practical effect: query examples now use `llm`, and `json` is scoped to what it's actually for (chaining between commands, extracting a specific field).

For reference, `ix-claude-plugin` already had this right — 66 `llm` uses in its skills. The repo's own CLAUDE.md was the one lagging.

## Scope

Single file, docs-only, no code paths touched. `CLAUDE.md` is the only copy of this block in the repo (the `IX-MEMORY` markers are read by `scripts/disconnect.sh` to remove it), so there is no template to keep in sync.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -5,17 +5,32 @@ This project uses Ix Memory — persistent, time-aware context for LLM assistant
 
 ## Interface
 
-Use the `ix` CLI exclusively. All commands support `--format json` for machine-readable output.
+Use the `ix` CLI exclusively.
 
-Use JSON when chaining command results.
+**Output format.** Query commands accept `--format text|json|llm`:
+
+- **`--format llm` — prefer this when you are reading the result yourself.** It is
+  token-minimal and newline-delimited, typically 2-4x fewer bytes than `json` on
+  tree- and table-shaped output. It is accepted on every command that accepts
+  `--format`, so you can pass it unconditionally without checking per command.
+- `--format json` — use when chaining results between commands, or when you need
+  to pull a specific field out of the response.
+
+Action commands take no `--format` at all: `config`, `ingest`, `init`, `reset`,
+`upgrade`, `view`, `watch`.
+
+**Pro features.** Some commands below are marked **[Pro]**. If any of them prints
+`The '<name>' command requires Ix Pro.`, this install does not have them — skip
+that step, do not retry it, and do not mention it again for the rest of the
+session. Everything not marked [Pro] always works.
 
 ## MANDATORY RULES
 1. BEFORE answering codebase questions → use targeted `ix` CLI commands (see routing below). Do NOT answer from training data alone.
-2. AFTER every design or architecture decision → run `ix decide <title> --rationale <text>`.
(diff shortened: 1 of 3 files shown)


josephismikhail (approved):
Separate-agent review run (xhigh, 22 agents) plus fixes applied in 8713a9e.

The review found the [Pro] detection contract this PR documents did not actually work: registerProStubs set only `allowUnknownOption`, which covers flags but not operands, and commander 15 rejects excess operands before the action runs. `ix bug` printed the sentinel; `ix bug create "title" --affects Entity` died with `too many arguments for 'bug'`. Verified against the pinned commander 15.0.0. Fixed the stubs rather than documenting around it, with tests over 14 documented Pro invocations.

Two further corrections, both cases where my own first check read a label instead of the behaviour behind it: `explain`, `read` and `status` declare `(text|json|llm)` but branch only on json, so my "the one exception is ix query" note was wrong; and `ix ingest` does accept `--format` (single-quoted in the source, which my grep missed). Also reconciled the Pro marking — Planning and Workflows mark Pro at the header, leaving 14 rows without the token that "everything not marked [Pro] always works" turns into a guarantee. Cross-checked all 13 PRO_COMMANDS are covered.

18/18 green.

PR #343, opened 2026-08-08
fix(release): stop make_latest re-pointing latest on a dispatch re-run

josephismikhail:
Fix-forward on #341, which is already merged. Neither issue affects a fresh RC or GA tag push, so this does not block the version push — but both should land before anyone re-runs the workflow on an old tag.

## `make_latest` can move `/releases/latest` backwards

#341 added `make_latest` alongside `prerelease`. The action documents it as *"Uses GitHub api default if not provided"* — so before that change, the update path left an existing release's flag alone.

This job also runs on `workflow_dispatch` against an arbitrary tag (the input is `required`, "e.g. v0.4.9"), and there `softprops/action-gh-release` **updates** the existing release rather than creating one. So re-running `v0.8.1` to repair a bad asset now PATCHes it with `make_latest: true`, and `/releases/latest` starts serving 0.8.1 to every new install via `resolve_version()` and every existing user via `checkForUpdate()` — the exact pointer #341 exists to protect.

Dropped it. `prerelease` alone is sufficient: GitHub refuses to mark a prerelease as latest, which is the whole mechanism #341 relies on.

## The release body was never prerelease-gated

Every other part of the step keys off `prerelease`; the `body:` did not. An RC's own release page therefore handed testers:

```bash
curl -fsSL .../install.sh | bash
```

which resolves through `/releases/latest` — deliberately the *previous stable* build on an RC. A tester following the RC page's own instructions installs the wrong version and reports results against it. Prereleases now lead with the `IX_VERSION=` form.

Verified the workflow still parses and that `make_latest` is absent while `prerelease` is retained.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -326,15 +326,24 @@ jobs:
         with:
           tag_name: v${{ steps.version.outputs.version }}
           generate_release_notes: true
-          # Without these, a v0.9.0-rc.1 tag publishes as a normal release and
+          # Without this, a v0.9.0-rc.1 tag publishes as a normal release and
           # becomes "latest" — and both install.sh's resolve_version() and the
           # CLI's checkForUpdate() read /releases/latest, so an RC would install
           # itself onto every new user and prompt every existing one to upgrade
           # to it. GitHub excludes prereleases from /releases/latest, so this is
           # what makes a release-candidate tag safe to push at all.
+          #
+          # `make_latest` is deliberately NOT set. The action documents it as
+          # "uses GitHub api default if not provided", and this job also runs on
+          # workflow_dispatch against an arbitrary tag — where it *updates* an
+          # existing release. Passing make_latest:true there would re-point
+          # /releases/latest at whatever old tag was re-run, which is the exact
+          # pointer this block exists to protect. Marking the tag prerelease is
+          # sufficient on its own: GitHub refuses to make a prerelease latest.
           prerelease: ${{ steps.version.outputs.prerelease }}
-          make_latest: ${{ steps.version.outputs.prerelease == 'false' }}
           body: |
+            ${{ steps.version.outputs.prerelease == 'true' && format('> **Release candidate — the install commands below will NOT give you this build.** `/releases/latest` still points at the last stable release, by design. Install this one explicitly with `IX_VERSION={0}` in front of the install command, e.g. `IX_VERSION={0} curl -fsSL https://raw.githubusercontent.com/{1}/main/scripts/install/install.sh | bash`', steps.version.outputs.version, github.repository) || '' }}
+
             ## Install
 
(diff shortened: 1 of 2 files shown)


KageBinary:
Checked both halves against the action source and the API contract; the reasoning holds.

`make_latest: legacy` is right, and the comment explaining why omission isn't equivalent is the part I'd have gotten wrong. Confirmed against `softprops/action-gh-release` at the pinned SHA `3d0d988` — `action.yml` documents the input as `true` / `false` / `legacy`, so `legacy` passes through rather than being dropped as unrecognised. The `prerelease == 'true' && 'false' || 'legacy'` idiom is safe here because `'false'` is a non-empty string and stays truthy in the `&&`/`||` fallthrough.

**One gap, fixed in da7b9a9.** The banner ended:

> The PowerShell and Homebrew paths below cannot install a release candidate at all.

Accurate, and that's the problem. `install.ps1` calls `Get-LatestVersion` unconditionally against `/releases/latest`, which excludes prereleases by design — there was no override, so Windows had no path to an RC at all.

That's the platform the RC most needs. #337 — the reason v0.9.0 is worth tagging — is a Windows `ix upgrade` bug: `~/.ix/cli` deleted before extracting with `unzip`, which stock Windows doesn't have, and the `.cmd` shim never repointed. If Windows testers can't install the candidate, that fix reaches GA untested on the only platform it affects.

Added `$env:IX_VERSION` to `install.ps1` mirroring `resolve_version()` in `install.sh`, and documented it next to the `IX_HOME` the script already read but never wrote down.

**Your banner had a latent bug too**, which is why it moved. Actions expression strings have no escape sequences, so the `\n` I first tried writing arrives in the release body as a literal backslash-n — fenced blocks are unreachable from inside `format()`. The banner is now built in the version step and passed through as `steps.version.outputs.banner`, which also gets it off a 700-character single line.

Ran that step's script directly under both tags:

- `v0.9.0-rc.1` → `prerelease=true`, banner with both commands, version and repo interpolated, fences intact
- `v0.9.0` → `prerelease=false`, empty banner, body still opens on `## Install`

Homebrew stays excluded; the formula step was already gated on `prerelease`.

**One caveat on my own commit:** nothing parses `install.ps1`. No local PowerShell here, and CI's `windows-2022` leg packages the CLI without touching the installer, so the 6 lines I added are reviewed but not executed. It's the smallest shape the change can take — `if ($env:IX_VERSION) { return $env:IX_VERSION }` — but the first real exercise will be a Windows tester running the RC command off the release page. Worth being the first thing tried, not the last.

**Separate nit, not fixed:** `Get-LatestVersion`'s catch returns a hardcoded `"0.6.0"`, so a rate-limited or offline install silently downloads a release that's four versions stale instead of failing. `install.sh` has the same shape with `0.1.0`. Both look like they want to be errors.

PR #344, opened 2026-08-09
fix(upgrade): compare prerelease versions so RC users see the GA

josephismikhail:
Found while reviewing the release-candidate flow. This one directly breaks the RC → GA handoff, so it matters before **v0.9.0-rc.1** goes out.

`isNewer` did `split(".").map(Number)`:

```js
"0.9.0-rc.1".split(".").map(Number)   // [0, 9, NaN, 1]
```

Every NaN comparison is false, and `(l[i] || 0)` then coerces it to 0 — so the release triple compares equal and the function returns false.

| call | before | correct |
|---|---|---|
| `isNewer("0.9.0", "0.9.0-rc.1")` | **false** | true |
| `isNewer("0.9.0-rc.2", "0.9.0-rc.1")` | **false** | true |
| `isNewer("0.9.0-rc.1", "0.9.0")` | false | false |

So a tester who installs the RC is never shown the update notice when the real 0.9.0 ships, and `ix upgrade` tells them `[ok] CLI already on the latest version (0.9.0-rc.1)`. They are stranded on the candidate with no error and no way to find the GA build from the CLI. Candidates can't be moved forward to each other either.

Note the release workflow does `npm version "$VERSION"` on the packaged tarball, so an RC install really does carry `0.9.0-rc.1` in its package.json — this is the live path, not a hypothetical.

Replaced with semver precedence: numeric triple, then a release outranks any prerelease of itself, then prerelease identifiers left to right with numeric ones compared numerically and ranked below alphanumeric. Build metadata stripped.

Tests cover RC → GA, RC → RC including `rc.10` vs `rc.9` (which a lexical compare gets wrong), the ordering example from the semver spec, build metadata, and malformed input. Reverting to the old implementation fails four of them. Full suite green (573).

### DIFF
--- a/ix-cli/src/cli/__tests__/upgrade-version-compare.test.ts
+++ b/ix-cli/src/cli/__tests__/upgrade-version-compare.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from "vitest";
 
-import { isNewer } from "../commands/upgrade.js";
+import { isNewer, VERSION_RE } from "../commands/upgrade.js";
 
 describe("isNewer", () => {
   it("compares plain releases", () => {
@@ -81,7 +81,8 @@ describe("isNewer", () => {
   it("accepts a tag carrying both a pre-release and build metadata", () => {
     // VERSION_RE gates every version that reaches isNewer. It used to reject
     // this shape, which surfaced as "Could not reach GitHub" and exit 1.
-    const VERSION_RE = /^\d+\.\d+\.\d+(?:-[0-9A-Za-z.-]+)?(?:\+[0-9A-Za-z.-]+)?$/;
+    // Imported, not re-declared: a local copy of the literal passes no matter
+    // what upgrade.ts actually does, so it would guard nothing.
     expect(VERSION_RE.test("0.9.0-rc.1+abc1234")).toBe(true);
     expect(VERSION_RE.test("0.9.0-rc.1")).toBe(true);
     expect(VERSION_RE.test("0.9.0+abc1234")).toBe(true);
--- a/ix-cli/src/cli/__tests__/upgrade-version-compare.test.ts
+++ b/ix-cli/src/cli/__tests__/upgrade-version-compare.test.ts
@@ -52,9 +52,42 @@ describe("isNewer", () => {
     }
   });
 
+  it("handles pre-release identifiers that themselves contain hyphens", () => {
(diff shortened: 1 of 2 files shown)


KageBinary:
Reviewed the logic rather than the diff, since the failure mode here is silent.

**Verified correct.** I ran a differential fuzz of `isNewer` against node-`semver`'s `gt()` over every version shape `VERSION_RE` admits — the numeric triple crossed with 18 pre-release forms (`rc.1`, `rc.10`, `alpha.beta`, `rc-1`, bare `-0`, `x.7.z.92`) and three build-metadata forms, sampled to ~50k ordered pairs. Zero disagreements in either direction. The semver-spec ordering example, `rc.10 > rc.9`, and the hyphen-in-identifier cases all hold.

Two judgement calls worth naming because they're both places a reasonable implementation goes wrong:

- `indexOf("-")` + `slice` over `split("-", 2)`. The limit argument truncates instead of capturing the remainder, so `0.9.0-rc-1` and `0.9.0-rc-2` would compare equal — the same stranding bug, just narrower.
- `continue` rather than `return` when two numeric identifiers are textually different but numerically equal (`01` vs `1`). Returning there ends the whole comparison as "not newer" on an undecided field.

The fuzz can't reach that second one — node-`semver` rejects `1.0.0-rc.01` as invalid, so it's excluded from the corpus. Your own unit test covers it, which is the right place for it.

**One issue, fixed in fa4325a.** The last case built its own copy of the regex literal instead of importing `VERSION_RE`:

```js
const VERSION_RE = /^\d+\.\d+\.\d+(?:-[0-9A-Za-z.-]+)?(?:\+[0-9A-Za-z.-]+)?$/;
expect(VERSION_RE.test("0.9.0-rc.1+abc1234")).toBe(true);
```

That asserts a regex literal against itself and passes for whatever `upgrade.ts` actually holds. I reverted the source to the old single-group form and the suite stayed green — the case guarding the `-rc.1+sha` regression didn't notice the regression coming back. Exported the constant and imported it; the same revert now fails that case, which I checked before and after.

No behaviour change from my commit. 578 tests pass, `tsc --noEmit` clean.

**Nit, not fixed:** `getCurrentVersion()` falls back to `"0.0.0"`, so a CLI that can't read its own package.json is offered every release as an upgrade. Pre-existing and arguably the right default — just noting it's now reachable through a comparison that actually works.

PR #346, opened 2026-08-09
fix(windows): install Compass where the CLI looks, and stop hiding why it failed

KageBinary:
Found by installing `v0.9.0-rc.1` on Windows. `ix view` is unavailable on every Windows install, and the repair path meant to rescue it fails silently. This is the regression #335 closed everywhere except Windows.

```
[!!] System Compass is not installed at C:\Users\...\.ix\cli\compass — 'ix view' is unavailable
     until you run 'ix upgrade', which will fetch it.
...
Compass update available: none → 0.2.0
[!!] Could not download compass update. ix view may use the bundled version.
```

## The zip is fine

It carries all 22 Compass files including `index.html`. They are nested under `ix-<version>-windows-amd64/`, and `Expand-Archive` has no `--strip-components`, so `install.ps1` left them at `cli\ix-<version>-windows-amd64\compass\` while `COMPASS_DIR` (upgrade.ts) and `findCompassDist` (view.ts) read `cli\compass` and nothing else. `findCompassDist` returns `null`, `ix view` refuses to start. `install.sh` has always stripped that directory via tar, which is why only Windows is affected.

`install.ps1` now extracts to staging, collapses the top-level directory, and swaps the result in only after confirming `ix.cmd` is inside it — moving the old install aside instead of deleting it first, since deleting first is what left Windows users with no CLI at all in #337. The shim loses its version-qualified path.

## `ix upgrade` would have undone it

This is the part worth reviewing closely. `upgrade.ts` resolved `stagedRoot` to locate the CLI entry point, then called `swapInStagedTree(installDir, stagingDir, ...)` — the **outer** directory. So it reproduced inside `cli\` exactly the nesting `stagedRoot` had just seen through. A flat install would have been re-nested by the first upgrade and Compass would have broken again, which would have made this look fixed in testing and regress later.

It now swaps in `stagedRoot` and clears the husk the outer directory leaves behind. POSIX is untouched: `--strip-components` already flattens, `soleChildDir` returns `null`, and `stagedRoot` is the staging directory itself.

The shim agrees by construction — `refreshLaunchers` already emits `%~dp0..\cli\ix.cmd` when there is no sole child directory, which is now the case, and that is the form `install.ps1` writes.

## Why nobody could tell what failed

One `catch` covered both the download and the extract and blamed the download either way. `tar` ran with `stdio: "ignore"`, so the step that actually failed was also the only one that printed nothing — and `curl` runs with `-fsSL`, whose `-S` would have printed had curl been the problem. Failures now name the stage and carry the real error: `ENOENT` is spelled out as a PATH problem, and tar's stderr is piped and read back.

I verified the download itself is not at fault — the URL the CLI builds returns HTTP 200 anonymously, `compass-0.2.0.tar.gz` exists at 583,970 bytes, and the tarball nests under `compass-0.2.0/` so `--strip-components=1` is correct.

## Tests

The staged-swap **composition** is now covered — the two helpers were each correct alone, which is why the bug survived. Restoring `swapInStagedTree(installDir, stagingDir, ...)` fails the new case; I checked before and after. `describeExecFailure` is tested against real spawn failures rather than hand-built error objects.

601 tests pass, `tsc --noEmit` clean.

## Not verified

**No PowerShell is available to me and CI never parses `install.ps1`** — the `windows-2022` leg packages the CLI without running the installer. The extraction rewrite is reviewed but unexecuted. I ran the flatten algorithm against the real `v0.9.0-rc.1` zip and it produces `cli\compass\index.html`, `cli\ix.cmd` and `cli\cli\dist\cli\main.js`, matching the Linux layout — but that validates the algorithm, not the PowerShell.

This wants a Windows run before v0.9.0 GA: install, `ix view`, then `ix upgrade` and `ix view` again to confirm the upgrade no longer re-nests.

### DIFF
--- a/ix-cli/src/cli/__tests__/upgrade-archive-shape.test.ts
+++ b/ix-cli/src/cli/__tests__/upgrade-archive-shape.test.ts
@@ -3,7 +3,12 @@ import { mkdtempSync, mkdirSync, writeFileSync, rmSync, existsSync, readFileSync
 import { join } from "node:path";
 import { tmpdir } from "node:os";
 
-import { soleChildDir, swapInStagedTree, sweepUpgradeOrphans } from "../commands/upgrade.js";
+import {
+  installStagedTree,
+  soleChildDir,
+  swapInStagedTree,
+  sweepUpgradeOrphans,
+} from "../commands/upgrade.js";
 
 let root: string;
 
@@ -15,14 +20,20 @@ afterEach(() => {
   rmSync(root, { recursive: true, force: true });
 });
 
-// The two helpers above are each correct in isolation; the bug was in how the
-// upgrade path combined them. It resolved stagedRoot to read the CLI entry
-// point, then swapped in the *outer* staging directory, so a Windows install
-// ended up at cli\ix-<version>-windows-amd64\ — reproducing inside cli\ exactly
-// the nesting stagedRoot had just seen through. COMPASS_DIR and
+// soleChildDir and swapInStagedTree are each correct in isolation; the bug was
+// in how the upgrade path combined them. It resolved the staged root to read
+// the CLI entry point, then swapped in the *outer* staging directory, so a
+// Windows install ended up at cli\ix-<version>-windows-amd64\ — reproducing
+// inside cli\ exactly the nesting it had just seen through. COMPASS_DIR and
 // findCompassDist read cli\compass and nothing else, so `ix view` was broken on
 // every Windows install and broke again on the first upgrade after install.ps1
 // started laying the tree down flat.
+//
+// So these go through installStagedTree, which is where the upgrade path makes
+// that choice. Resolving the staged root here and calling swapInStagedTree
+// directly would pass against the bug too — the defect was never in either
+// helper, only in the line that joined them, and a test that re-derives the
+// composition cannot see it.
 describe("staged swap, as the upgrade path composes it", () => {
   function seedWindowsZipStaging(home: string, ver: string) {
     const staging = join(home, `.cli-staging-${ver}`);
@@ -38,8 +49,7 @@ describe("staged swap, as the upgrade path composes it", () => {
     const installDir = join(root, "cli");
     const staging = seedWindowsZipStaging(root, "0.9.0");
 
-    const stagedRoot = soleChildDir(staging) ?? staging;
-    swapInStagedTree(installDir, stagedRoot, join(root, ".cli-backup-1"));
+    installStagedTree(installDir, staging, join(root, ".cli-backup-1"), true);
 
     // COMPASS_DIR / findCompassDist read exactly this path.
     expect(existsSync(join(installDir, "compass", "index.html"))).toBe(true);
@@ -48,10 +58,23 @@ describe("staged swap, as the upgrade path composes it", () => {
     expect(existsSync(join(installDir, "ix-0.9.0-windows-amd64"))).toBe(false);
   });
 
+  it("clears the staging husk the inner move leaves behind", () => {
+    const installDir = join(root, "cli");
+    const staging = seedWindowsZipStaging(root, "0.9.0");
+
+    installStagedTree(installDir, staging, join(root, ".cli-backup-husk"), true);
+
+    // Moving the inner directory out leaves the outer one standing. Left in
+    // place it is a `.cli-staging-<pid>` that only the next run's sweep
+    // reclaims, and on a reused pid that sweep is what the *next* upgrade
+    // trips over.
+    expect(existsSync(staging)).toBe(false);
+  });
+
   it("keeps the launcher pointing at the install root once the tree is flat", () => {
     const installDir = join(root, "cli");
     const staging = seedWindowsZipStaging(root, "0.9.0");
-    swapInStagedTree(installDir, soleChildDir(staging) ?? staging, join(root, ".cli-backup-2"));
+    installStagedTree(installDir, staging, join(root, ".cli-backup-2"), true);
 
     // refreshLaunchers derives the shim target this way: a flat install has
     // several directories, so there is no sole child and the shim resolves to
@@ -67,12 +90,29 @@ describe("staged swap, as the upgrade path composes it", () => {
     mkdirSync(join(staging, "compass"), { recursive: true });
     writeFileSync(join(staging, "compass", "index.html"), "<!doctype html>\n");
 
-    // soleChildDir returns null here, so stagedRoot falls back to staging.
-    const stagedRoot = soleChildDir(staging) ?? staging;
-    expect(stagedRoot).toBe(staging);
-    swapInStagedTree(installDir, stagedRoot, join(root, ".cli-backup-3"));
+    // isWindows false: the staged root is the staging directory itself, so the
+    // tree is installed whole and there is no husk to clear.
+    installStagedTree(installDir, staging, join(root, ".cli-backup-3"), false);
 
     expect(existsSync(join(installDir, "compass", "index.html"))).toBe(true);
+    expect(existsSync(join(installDir, "cli", "dist", "cli", "main.js"))).toBe(true);
+  });
+
+  it("installs a Windows tree that is already flat without re-nesting it", () => {
+    // install.ps1 now lays the tree down flat, so the *next* upgrade sees a
+    // Windows staging directory with no sole child. soleChildDir returns null
+    // and the fallback has to install the staging directory itself.
+    const installDir = join(root, "cli");
+    const staging = join(root, ".cli-staging-flat");
+    mkdirSync(join(staging, "cli", "dist", "cli"), { recursive: true });
+    writeFileSync(join(staging, "cli", "dist", "cli", "main.js"), "// flat\n");
+    mkdirSync(join(staging, "compass"), { recursive: true });
+    writeFileSync(join(staging, "compass", "index.html"), "<!doctype html>\n");
+
+    installStagedTree(installDir, staging, join(root, ".cli-backup-4"), true);
+
+    expect(existsSync(join(installDir, "compass", "index.html"))).toBe(true);
+    expect(existsSync(join(installDir, "cli", "dist", "cli", "main.js"))).toBe(true);
   });
 });
 
--- a/ix-cli/src/cli/__tests__/upgrade-archive-shape.test.ts
+++ b/ix-cli/src/cli/__tests__/upgrade-archive-shape.test.ts
@@ -15,6 +15,67 @@ afterEach(() => {
   rmSync(root, { recursive: true, force: true });
 });
 
+// The two helpers above are each correct in isolation; the bug was in how the
+// upgrade path combined them. It resolved stagedRoot to read the CLI entry
+// point, then swapped in the *outer* staging directory, so a Windows install
+// ended up at cli\ix-<version>-windows-amd64\ — reproducing inside cli\ exactly
+// the nesting stagedRoot had just seen through. COMPASS_DIR and
+// findCompassDist read cli\compass and nothing else, so `ix view` was broken on
+// every Windows install and broke again on the first upgrade after install.ps1
+// started laying the tree down flat.
+describe("staged swap, as the upgrade path composes it", () => {
+  function seedWindowsZipStaging(home: string, ver: string) {
+    const staging = join(home, `.cli-staging-${ver}`);
+    const nested = join(staging, `ix-${ver}-windows-amd64`);
+    mkdirSync(join(nested, "cli", "dist", "cli"), { recursive: true });
+    writeFileSync(join(nested, "cli", "dist", "cli", "main.js"), `// ${ver}\n`);
+    mkdirSync(join(nested, "compass"), { recursive: true });
+    writeFileSync(join(nested, "compass", "index.html"), "<!doctype html>\n");
+    return staging;
+  }
+
+  it("leaves compass where the CLI looks for it, given a nested Windows zip", () => {
+    const installDir = join(root, "cli");
+    const staging = seedWindowsZipStaging(root, "0.9.0");
+
+    const stagedRoot = soleChildDir(staging) ?? staging;
+    swapInStagedTree(installDir, stagedRoot, join(root, ".cli-backup-1"));
+
+    // COMPASS_DIR / findCompassDist read exactly this path.
+    expect(existsSync(join(installDir, "compass", "index.html"))).toBe(true);
+    expect(existsSync(join(installDir, "cli", "dist", "cli", "main.js"))).toBe(true);
+    // The nesting must not survive into the install.
+    expect(existsSync(join(installDir, "ix-0.9.0-windows-amd64"))).toBe(false);
+  });
+
+  it("keeps the launcher pointing at the install root once the tree is flat", () => {
+    const installDir = join(root, "cli");
+    const staging = seedWindowsZipStaging(root, "0.9.0");
+    swapInStagedTree(installDir, soleChildDir(staging) ?? staging, join(root, ".cli-backup-2"));
+
+    // refreshLaunchers derives the shim target this way: a flat install has
+    // several directories, so there is no sole child and the shim resolves to
+    // %~dp0..\cli\ix.cmd — the form install.ps1 now writes.
+    expect(soleChildDir(installDir)).toBeNull();
+  });
+
+  it("is unchanged for a POSIX tarball, already flattened by --strip-components", () => {
+    const installDir = join(root, "cli");
+    const staging = join(root, ".cli-staging-posix");
+    mkdirSync(join(staging, "cli", "dist", "cli"), { recursive: true });
+    writeFileSync(join(staging, "cli", "dist", "cli", "main.js"), "// posix\n");
+    mkdirSync(join(staging, "compass"), { recursive: true });
+    writeFileSync(join(staging, "compass", "index.html"), "<!doctype html>\n");
+
+    // soleChildDir returns null here, so stagedRoot falls back to staging.
+    const stagedRoot = soleChildDir(staging) ?? staging;
+    expect(stagedRoot).toBe(staging);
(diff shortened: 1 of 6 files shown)


KageBinary:
Verified on Windows against `v0.9.0-rc.1`, installing with this branch's `install.ps1`:

```
-- CLI --
Extracting CLI...
  [ok] Extraction complete
  [ok] CLI installed
Testing CLI...
  [ok] CLI working: 0.9.0-rc.1

PS C:\Users\ianja> ix view
[ok] Visualizer started (PID 53852)
  http://localhost:8080
```

The `System Compass is not installed at ...\cli\compass` warning is gone, and `ix view` starts. That is conclusive rather than suggestive: `view.ts` calls `process.exit(1)` with `[error] Compass UI not found` when `findCompassDist()` returns null, so a started visualizer means Compass was resolved at `$IX_HOME\cli\compass` — the path that did not exist on Windows before this branch.

## A third bug the same run surfaced (e0dc938)

With a backend already running, the install stops dead on an interactive prompt:

```
Security Warning: Script Execution Risk
Invoke-WebRequest parses the content of the web page. Script code in the web page might be run when the page is parsed.
[Y] Yes  [A] Yes to All  [N] No  [L] No to All  [S] Suspend  [?] Help (default is "N"):
```

`Test-Healthy` called `Invoke-WebRequest` without `-UseBasicParsing`, so Windows PowerShell handed the response to the IE engine and asked permission before parsing it.

The timing explains why it survived this long. It needs a backend that is up **and returning a body** — on a first install nothing is listening, the request fails outright, and there is no parse to warn about. It fires only on re-runs and upgrades, which is every machine except the one you first test on. That is also why it did not appear in the earlier RC install on this same box and did appear on the second.

Piped through `irm ... | iex` the default answer is No, and `Test-Healthy` then reports a healthy backend as down, so the installer restarts it for nothing.

Neither response body is read — only the status matters — so basic parsing costs nothing, and the switch is inert on PowerShell 6+. No other `Invoke-WebRequest` under `scripts/` is missing it.

## Still unverified

`ix upgrade` no longer re-nesting the install. That fix is in the CLI bundle, not the installer, so it needs a build carrying it — `v0.9.0-rc.2` after this merges, then install → `ix view` → `ix upgrade` → `ix view` on Windows. The second `ix view` is the one that matters: it proves the upgrade did not put the nesting back and re-break Compass.

KageBinary:
Self-review of this branch, since green CI here means very little: nothing in the pipeline parses `install.ps1`, so the checks that passed never executed the code most likely to be wrong.

Two findings, both in my own swap logic, both fixed in 766de1d.

**1. The recovery path could destroy the install it exists to protect.** I ported `swapInStagedTree` to PowerShell but dropped its first line, `rmQuiet(backupDir)`.

`Move-Item` onto a directory that already exists moves the source *inside* it rather than renaming over it. So a stale `.cli-backup-<pid>` — left by an earlier run that died mid-swap, on a pid Windows has since reused — turns the old install into `.cli-backup-<pid>\cli`. The restore in the `catch` then reinstates that whole directory as `cli\`, yielding `cli\cli\` and a CLI that cannot start.

Narrow trigger, but the blast radius is the failure path itself: the backup is only ever read when something has already gone wrong, so this made the safety net most dangerous exactly when it was needed. Now cleared before use, matching the TS.

**2. The archive guard was lenient where `soleChildDir` is strict.** `Select-Object -First 1` took one of several top-level directories arbitrarily and could have collapsed the wrong tree into `cli\` — the `ix.cmd` probe would catch most such cases, but not one where the wrong directory happens to contain an `ix.cmd`. It now requires exactly one, the same rule `soleChildDir` applies, and names the count in the error when it does not hold.

Re-validated the flatten against the real `v0.9.0-rc.1` zip: exactly one top-level directory, `ix.cmd` inside it, `cli\compass\index.html` present after the swap, and the stale-backup scenario now resolving correctly.

**Unchanged caveat:** the PowerShell is still reviewed rather than executed here. What was verified on real Windows is the *previous* commit — install, no missing-Compass warning, `ix view` starting. Neither finding above affects that path, since both concern a stale backup and a malformed archive, but a re-run on Windows before GA would exercise the rewritten guard.

josephismikhail:
Reviewed and pushed four commits. The diagnosis is right and the core fix is right — I verified the release zip really does carry exactly one top-level directory with `ix.cmd` inside it (`release.yml:266,286`, `7z a -tzip "$STAGING.zip" "$STAGING"`), so the flatten's precondition is guaranteed by the packaging step rather than merely observed once. `refreshLaunchers` also resolves correctly against a flat tree: three top-level dirs means `soleChildDir` returns null and the shim lands as `%~dp0..\cli\ix.cmd`, which is what `install.ps1` writes and what `cli\ix.cmd` expects.

Four things needed fixing before this could go in. Two of them lose user data.

## The swap was not atomic

`Move-Item` on a directory is not a rename. When it cannot rename it falls back to a recursive copy-then-delete, and it throws part way through. I reproduced it on PS 5.1 and 7 by holding a read handle on `cli\compass\index.html`:

```
move FAILED: The process cannot access the file because it is being used by another process.
  InstallDir exists: True
  Backup exists    : True
  restore guard fires: False
   \.cli-backup-1\ix.cmd
   \.cli-backup-1\cli\dist\cli\main.js
   \cli\compass\index.html
```

The install ends up split across both paths, and the restore never fires — it guards on `-not (Test-Path $InstallDir)` and `cli\` still exists. So the branch written to protect the user's CLI is skipped in exactly the case that destroyed it, and the installer exits printing only the lock message. An open handle under `cli\` is routine on Windows: `ix view` serving compass out of it, `ix watch` holding tree-sitter's `.node` addons mapped for the life of the process, Defender scanning the native modules the extract just wrote.

`[System.IO.Directory]::Move` is a real rename — it either happens or the source is untouched, and it refuses to move into an existing destination rather than nesting inside it. That is the atomicity `swapInStagedTree` gets from `fs.renameSync`, and which this block only claimed to match. It also makes the `#337` guarantee true rather than aspirational.

Verified on **both PS 5.1 and 7** against a release-shaped zip, four cases: fresh install; upgrade over an existing install; upgrade with a locked file (old install now **fully intact**, exit 1, real IO message); and a stale `.cli-backup-<pid>` on a reused pid (clean install, no `cli\cli\`).

Also in that block: the backup pre-clear now fails loudly instead of hiding a *failed* removal behind `-ErrorAction SilentlyContinue`; the restore reports where the surviving copy is, the way `upgrade.ts:729` does; `$_` is captured before the nested catch rebinds it; and `-LiteralPath` throughout.

## `install.sh` was a missed site

The PR body says install.sh "has always stripped that directory via tar". That is true of the POSIX branch, but there is a `windows-amd64` branch, and it ran `unzip -q "$TMP_FILE" -d "$INSTALL_DIR"` with no strip. So `ix view` stayed broken for anyone installing through Git Bash — the bug this PR is named for, on one of the two Windows paths.

Worse, the two installers became mutually destructive. `check_installed_version()` probed `[ -d "$INSTALL_DIR/ix-${VERSION}-windows-amd64" ]`, so a flat install laid down by the new `install.ps1` reports `unknown` → `rm -rf "$INSTALL_DIR"` → re-laid nested, leaving the `bin\ix.cmd` shim `install.ps1` had just written pointing at nothing.

It now stages, collapses the sole top-level directory behind the same one-directory-containing-`ix.cmd` guard, writes a version-free shim, and reads the version out of `cli/package.json`. Verified in Git Bash: flat tree, compass at `cli/compass`, shim resolves, rerun is idempotent, malformed archive rejected. The stale comment telling the next reader not to point the compass stamp at `cli/compass` has gone with the reason for it.

## The regression tests did not test the fix

This is the one worth taking seriously, because the PR body says "Restoring `swapInStagedTree(installDir, stagingDir, ...)` fails the new case; I checked before and after."

It doesn't. The tests resolve the staged root themselves and call `swapInStagedTree` directly, so they never touch the line the bug was on. I put the bug back and ran them:

```
Test Files  1 passed (1)
     Tests  15 passed (15)
```

That is the trap the PR body correctly identifies and then falls into — both helpers were always right alone, the defect was in the line that joined them, so a test that re-derives the composition passes against the bug and against the fix equally. `expect(soleChildDir(installDir)).toBeNull()` has the same problem in miniature: it holds because the fixture seeds two directories.

`resolveStagedRoot` and `installStagedTree` now own that choice and the tests go through it. Restoring the bug fails 2 of them. I also added the flat-Windows-staging case: once `install.ps1` lays the tree down flat, the *next* upgrade stages a Windows tree with **no** sole child, and nothing covered that null fallback.

`describeExecFailure`'s test had it too — `toContain("definitely-missing.tar.gz")` passes off the `Command failed: tar -xzf <path>` prefix alone, so it held with `.stderr` stripped off the error entirely.

## `describeExecFailure` printed stderr twice, and misread ENOENT

`execFileSync` already folds piped stderr into `.message`, so `${e.message}: ${stderr}` repeated tar's entire complaint verbatim:

```
Command failed: tar -xzf /nonexistent/definitely-missing.tar.gz
tar (child): ...Cannot open: No such file or directory
...
: tar (child): ...Cannot open: No such file or directory      <-- again
```

And the PATH hint keyed on `e.code === "ENOENT"`, which every filesystem miss carries too — `writeFileSync` of `.version` failing rendered as `ENOENT: no such file or directory, open '...\.version' — is it installed and on PATH?`. The spawn syscall is what separates them (`spawnSync tar` vs `open`). Both fixed, both now covered by tests that fail against the old function.

## And one the fix newly exposed

The compass repair emptied `COMPASS_DIR` and extracted into the hole, so a failed tar left no compass at all. That cost nothing while the Windows bundle sat unreachable at `cli\ix-<version>-<platform>\compass` — there was never anything at `COMPASS_DIR` to lose. `install.ps1` now puts it exactly there, which turns the repair path into a way to destroy the copy the install just got right, on the one platform this PR exists to fix, via the exact tar failure it now bothers to report.

`installCompassBundle` unpacks beside the live copy and swaps, reusing `swapInStagedTree`, and rejects an archive with no `index.html` rather than swapping in a directory `findCompassDist` refuses and then stamping `.version` over it. The failure message now distinguishes a degraded upgrade from a broken `ix view`.

Building the tarball fixture surfaced the same class of problem the `cygpath` guard exists for: Git for Windows puts GNU tar on PATH ahead of System32's bsdtar, and GNU tar reads `C:\...` in `-f` as a `host:path` remote — `tar (child): Cannot connect to C: resolve failed`. Worth knowing it is reachable from a plain `tar` call on a developer box.

---

610 tests pass, `tsc --noEmit` clean, eslint 0 errors, knip clean, both scripts syntax-checked, CodeQL green.

## Not fixed — worth their own changes

**The compass `.version` stamp records the wrong version, and this PR activates it on Windows.** `release.yml:103-108` copies `system-compass`'s `dist/` into the bundle with no `.version` in it, so the bundle carries no version identity. Both installers then stamp it with the latest release of `ix-compass-dist` — a different repo. If a compass ships between a release being cut and a user installing it, the bundled compass N is stamped N+1 and `ix upgrade` never fetches N+1: `ix view` works, permanently one version stale. Dead on Windows until now because the `Test-Path $CompassIndex` guard never passed; already live on POSIX (`install.sh:1037-1039`). The fix belongs in the release: record the bundled compass version in the bundle and stamp *that*.

**`install.ps1` cannot be parsed as a file by Windows PowerShell 5.1.** No BOM, ~1200 non-ASCII bytes, so 5.1 decodes it as the ANSI codepage and the em dashes break the parse. Pre-existing — `origin/main` already fails with 2 errors, this branch has 16 purely from added comment text — and PS 7 parses both clean. The documented `irm | iex` path is unaffected because the content arrives already decoded, which is why nobody has hit it. I left it alone deliberately: adding a BOM to fix a path nobody is told to use could break the one everybody uses, and that deserves testing on its own.

**`install.ps1` still swaps on `ix.cmd` existing, not on the CLI running.** `upgrade.ts:693-705` runs `node <stagedEntry> --version` before the swap precisely so a truncated extraction cannot replace a working install; the installer checks one file exists, swaps, then tests afterwards and only warns.

**`sweepUpgradeOrphans` picks the recovery backup with a lexicographic sort** (`.cli-backup-9972` sorts after `.cli-backup-22736`), and `install.ps1` now writes into that same namespace.

The Windows run you asked for is still worth doing before GA — install, `ix view`, then `ix upgrade` and `ix view` again — but the upgrade half is now covered by tests that fail without the fix, rather than by tests that agree with it.

josephismikhail:
Two more review rounds after the comment above, both of which found bugs in **my** fixes rather than in the original change. Recording them because the pattern is the interesting part: each round's fix opened the next round's bug, always in the recovery path, and always in the direction of destroying the thing the recovery existed to protect.

**Round 3 — the compass cleanup deleted the copy it was protecting.** I added `rmQuiet(compassBackup)` unconditionally after the catch. But `swapInStagedTree` moves the live compass to the backup, then renames staging into place; if that second rename throws it attempts a restore, and if the restore throws too it swallows that and rethrows. In that state the user's compass exists *only* at `.compass-backup-<pid>` — and my next line erased it. The CLI swap two hundred lines up already handled this correctly; I copied the shape and dropped the guard. Also: nothing swept `.compass-staging-*` / `.compass-backup-*`, so an interrupted run leaked a full bundle copy into `~/.ix` and a kill between the two renames left the compass unrecoverable.

And `install.sh`'s `rm -rf` + `mv` had no rollback at all, on an assumption that `$INSTALL_DIR` is the empty directory `mkdir -p` just made. It isn't, whenever the two installers are mixed — `install.ps1` writes `%IX_HOME%\bin\ix.cmd`, not `$IX_BIN/ix`, so the `rm -rf` guard above is skipped and a populated install reaches it.

**Round 4 — two ways to brick an install, both in the code I had just added.**

A bare `rm -rf` that fails under `set -eu` aborts the script where it stands. The cleanup after a *successful* swap sits between the new tree and the shim write, so a locked leftover installed the CLI and then skipped the launcher, `ensure_path` and both version stamps — a working tree on disk with a stale or absent `ix` on PATH.

I initially mis-tested this and concluded it was safe. My check wrapped the case in `|| true`, which suppresses `set -e` and made a failing `rm` look survivable:

```
A: bare failing rm under set -eu:
  ...continued (NOT aborted)
```

Run as a real script file, with no `||` wrapper, it aborts:

```
== t2: rm -rf that genuinely fails (stubbed) ==
   exit=1        # "REACHED: shim write would happen here" never printed
```

The second: `mv "$ZIP_BACKUP" "$INSTALL_DIR" && warn "Restored..."` short-circuits when the restore fails, and the `else` fired only when the restore was never *attempted*. So the single path where the user's CLI survives **only** in the backup printed no mention of it. That was exactly the gap I had closed for `install.ps1` one round earlier and not for `install.sh`.

Plus: `cleanupCompassSwap` guarded on the compass *directory* while the message branch guarded on `index.html` inside it — and a `compass/` holding only a `.version` is a real state, since `install.ps1` used to create precisely that and v0.7.0–v0.8.1 shipped an empty `compass/`. In that state the code told the user their compass was at the backup and then deleted it. One predicate away from the round-3 bug.

And `Out-File`'s positional `-FilePath` still glob-expanded, so a home directory containing `[` or `]` killed the installer *after* the swap, at the launcher write, rather than before it — strictly worse than the failure it replaced.

---

**Corrections to my earlier comment.** I wrote that an interrupted compass swap leaked a bundle copy "permanently". It was really "until the next CLI upgrade" — the sweep only ran behind *the CLI itself needs updating*. It now also runs in the compass block, which is the case where the user cannot afford to wait, because the backup may hold their only compass.

**How this was verified.** Not by reading. `install.sh`'s block was lifted verbatim and run as a real script under live `set -eu` across seven states — fresh, populated install, restore-`mv`-fails, fresh-plus-failed-swap, cleanup-`rm`-fails, stale backup, un-removable stale backup. The two that matter:

```
===== F1: rm of the backup fails AFTER a successful swap =====
    REACHED THE SHIM WRITE
    exit=0
    installed main.js: // NEW main

===== F2: restore mv fails - must still name the surviving backup =====
    [warn] Your previous CLI is at .../.cli-backup-1291722 — move it to .../cli to restore it.
    [err] Could not install to .../cli.
    exit=1
```

`install.ps1`'s swap was run across fresh / upgrade / locked-file / stale-backup on **both Windows PowerShell 5.1 and PowerShell 7**. 614 tests, `tsc` clean, eslint 0 errors, knip clean, `sh -n` clean, CodeQL green.

Every new test was checked by restoring the bug and confirming it fails — including the ones from the earlier rounds.

## Left open, deliberately

Two symmetry nits that are not defects and did not seem worth another round: `install.sh`'s backup pre-clear checks existence only when `rm` returns non-zero, where `install.ps1` checks it unconditionally (the divergence needs a lying `rm` *and* a second independent failure before it has any consequence, and the lying `rm` could not be produced on MSYS); and one half of one new test is belt-and-braces rather than load-bearing, because `renameSync` onto an existing directory throws anyway.

The three items from my earlier comment still stand as their own changes — most importantly the compass `.version` stamp recording the latest `ix-compass-dist` release over a bundle that carries no version of its own.

**The Windows run before GA is still worth doing.** Four rounds of review each found something real in code that CI never executes, which is the argument for running it rather than against it.

josephismikhail (approved):
Approving. Reviewed across four rounds by independent agents; every finding is either fixed on the branch or recorded above as a follow-up with its reasoning.

The original diagnosis and the core flatten are correct, and I verified the release packaging guarantees the shape the fix depends on rather than it being observed once. What needed fixing was the recovery paths: a non-atomic Move-Item that split the install across two directories with the restore guard unable to fire, install.sh's Windows branch still nesting (and re-nesting a flat install), and regression tests that passed against the bug they described. Rounds 3 and 4 found bugs in my own fixes, which is why there were four.

All CI green on 3afd406. install.ps1's swap verified on both PowerShell 5.1 and 7 across fresh/upgrade/locked-file/stale-backup; install.sh's block run verbatim under set -eu across seven states.

The Windows run before GA is still worth doing — install, ix view, then ix upgrade and ix view again.

PR #347, opened 2026-08-09
fix(map): fail loudly when patches do not commit instead of reporting success

KageBinary:
`ix map` reported success when every write to the graph failed.

## The path

Commit failures in the per-file fallback (`ingest.ts:878`) were counted as `parseErrors` and printed only under `--debug`. `ingestFiles` never threw and never set an exit code, and `map.ts:253` passes `suppressOutput: true` unconditionally — so the function returned at the early exit before printing anything, and `map.ts` went on to render regions built from a graph that had not moved.

`map.ts` already had the right handling: a `try/catch` around `ingestFiles` calling `emitError` and setting exit 1. It simply never fired, because nothing ever threw.

## Why it is worth fixing on its own

CLAUDE.md RULE 5 tells every agent to run `ix map --silent` after each edit. A backend that accepted nothing left the agent believing the graph was current while every subsequent `ix search` and `ix impact` answered from a stale one. The failure surfaces as bad answers from the query commands, which is the hardest possible place to trace it back from — and on the `--silent` path there was no signal at all.

## What changed

Commit failures get their own counter, reported **before** the `suppressOutput` return.

- **Committed nothing** → throw. `map.ts` turns it into `emitError` + exit 1.
- **Partial** → warn on stderr. Some patches landed, the graph moved forward, and the mtime cache is not written on a run with commit errors, so the next run retries the rest. Failing the whole command there would be an overreaction.

Splitting the counter required naming `commitErrors` in all three guards that previously relied on commit failures inflating `parseErrors`. **The mtime-cache guard is the one that matters** — without it a failed commit would cache the file's mtime, the next run would skip it as unchanged, and it would stay missing from the graph until a `--force`. That would have been a worse bug than the one being fixed, introduced by the fix.

`skipReasons.parseError` now counts only parse errors. Commit failures appear as a sibling `commitErrors` field in the JSON and as their own line in the text summary.

The audit's fourth item — gating `All files unchanged since last ingest` on the commit count — is subsumed rather than skipped: that branch requires `patchesApplied === 0`, which now throws before reaching it.

## Testing

The decision is a pure exported function, `describeCommitOutcome`, covered by 6 cases including the single/plural boundary and a defensive negative count. A failure nobody can see deserved better than a rule nobody can test, and `ingestFiles` itself cannot be unit-tested without a backend.

600 tests pass, `tsc --noEmit` clean, and `ingest.ts` carries 19 lint warnings before and after — I checked against `main` rather than assuming.

## Not covered

The end-to-end path — a backend that accepts a patch request and then fails the commit — is not reproducible without a fault-injecting backend, so what is verified here is the rule and the wiring, not a live failed commit.

### DIFF
--- a/ix-cli/src/cli/__tests__/ingest-commit-outcome.test.ts
+++ b/ix-cli/src/cli/__tests__/ingest-commit-outcome.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from "vitest";
 
-import { describeCommitOutcome } from "../commands/ingest.js";
+import { describeCommitOutcome, ingestCompletedCleanly } from "../commands/ingest.js";
 
 describe("describeCommitOutcome", () => {
   it("says nothing when every patch committed", () => {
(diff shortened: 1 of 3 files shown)


josephismikhail:
Reviewed and pushed a fix commit. The diagnosis is right, and the part the description flags as load-bearing genuinely is — I confirmed `origin/main` had exactly three `parseErrors === 0` guards, that all three carry `commitErrors` here, that no fourth reader was missed, and that a file whose commit failed cannot get its mtime cached (the guard is all-or-nothing for the run, and `knownHashes` is loaded from the backend *before* any commit, so the file is re-attempted). "Re-run `ix map` to retry" is accurate.

Four things needed fixing.

## The threshold measured "did anything land", not "did the run succeed"

`describeCommitOutcome(999, 1)` returned `warn`, so `ix map` exited 0 and printed a map built from a graph missing 999 of 1000 files.

That is reachable, not theoretical: commits go out in chunks (`COMMIT_HTTP_MAX_FILES` = 1000, one call per 500-file parse batch), so a 2000-file run has four independent chunks and one landing is enough. One unhealthy replica, or a body-size limit that only bites large files, gets you there.

The sharper problem is that it made the severity of an *identical* fault depend on an unrelated variable. Under CLAUDE.md RULE 5 the typical run changes one file — so one transient commit failure is **fatal, exit 1** when one file changed, and **silent, exit 0** when two changed and one landed. That is the same failure mode this PR exists to fix, moved from 100% to 99.9%.

A partial failure still means the graph does not match the working tree, and the caller — usually a hook running `ix map --silent` — reads the exit code, not stderr. Partial now sets `process.exitCode = 1` **without throwing**: the graph did move forward, so the summary and JSON still print, but the run stops claiming success. That is less than failing the command outright, which the description rightly calls an overreaction.

## The remediation it offered does not exist

Both messages said "re-run with `--debug`". `ix map` has no `--debug`, registers none, and never passed one to `ingestFiles` — so on the one path where this failure is otherwise invisible, the only instruction given was inert, and the per-file `[commit error] <uri>` detail was unreachable from the command that produced the failure.

`map.ts` now passes its own `--verbose` through as `debug`, and the flag name is a parameter so each command names the one it actually has.

## A timed-out run blamed the backend

`ix map` installs a 15-minute deadline shared by every request. When it fires, the bulk commit throws, the per-file fallback runs, and every remaining patch aborts instantly straight into `commitErrors` — producing *"all N patches failed to commit … the backend rejected them"*. Wrong system: the cause is our own wall clock and the fix is `IX_MAP_DEADLINE_MS`. Now reported distinctly, which is the same information loss this PR fixes for parse-vs-commit.

## `--format json` printed nothing on the run you'd most want to parse

The throw sat before the JSON emit, so a total failure produced empty stdout and exit 1, where `main` emitted a valid document carrying `patchesApplied: 0`. Moved to the end of the function; the `suppressOutput` path still throws early because it has no report to emit.

## Tests

The load-bearing guard had none. Deleting `&& commitErrors === 0` from the mtime-cache condition left all 598 tests green — the six new cases covered only the 6-line pure function, which is the least likely thing here to regress. The three guards now route through an exported `ingestCompletedCleanly()` that a test pins; dropping `commitErrors` from it fails.

The "defensive negative count" case asserted `describeCommitOutcome(-1, 10)`. With `patchesApplied` positive the fatal branch is unreachable under *any* implementation of the first guard, so it asserted nothing about the invariant it names — `(-1, 0)` is the case that does.

605 tests, `tsc` clean, eslint 0 errors, knip clean.

## Not addressed — flagged

When `isCloudReady()`, `ix map` routes ingestion through `getRemoteRunner().runIngestion` and never reaches `ingestFiles` (`map.ts:231-243`). If that runner has the same shape, `ix map` against a cloud instance can still report success when every write failed. The implementation is in the Pro repo, so it needs its own change there — worth checking, since it is the same bug in the same command.

The end-to-end path you called out as unreproducible — a backend that accepts a patch request and then fails the commit — is still unreproducible here, so what is verified is the rule and the wiring, not a live failed commit.

josephismikhail (approved):
Approving. Reviewed by independent agents across two rounds; every finding is fixed on the branch or recorded in-code and in the thread as follow-up.

The diagnosis was right and the mtime-cache guard the description singles out is genuinely the trap — verified that no fourth reader of parseErrors was missed and that a failed commit cannot cache a file's mtime. What needed fixing was the severity rule (a 999-of-1000 failure exited 0, and severity flipped on how many files happened to change), a remediation message naming a flag `ix map` does not have, a timed-out run blaming the backend, and `--format json` emitting nothing on the run a consumer most needs to parse. Plus a defect I introduced myself: moving the throw made "All files unchanged since last ingest." reachable under "commit errors: N".

The moved throw was proved reachable on every non-suppressOutput path by AST-enumerating every exit in ingestFiles, so there is no exit-0-on-total-failure regression.

605 tests, tsc clean, eslint 0 errors, knip clean, all CI green on 25fc9ca.

Follow-ups in the thread: the cloud path (getRemoteRunner().runIngestion) bypasses ingestFiles entirely and can still report success when every write failed; and ingestFiles has no execution coverage, so the exit-code and debug wiring are verified by reading.

PR #348, opened 2026-08-09
fix(ingest): stop DEP0151 warnings from flooding the ix map progress bar

KageBinary:
Reported from `ix map` on Windows, Node 23.9.0: a wall of DEP0151 warnings interleaved with the progress bar, which shares stderr with them. The bar was unreadable and the run looked hung at `0% 0 / 125`.

```
(node:61632) [DEP0151] DeprecationWarning: Package ...\tree-sitter-c-sharp\ has a "main"
field set to "bindings/node", excluding the full filename and extension to the resolved
file at "bindings\node\index.js", imported from ...\core-ingestion\dist\index.js.
```

## Cause

All twelve required grammars publish `"main": "bindings/node"` — no filename, no extension, no `exports` field. Resolving the bare specifier from ESM therefore falls back to directory-index lookup, which is what DEP0151 objects to. The parse pool loads this module once per worker, so the warnings arrive twelve at a time, per worker.

**The fix already exists in this file.** `tree-sitter-css` is imported by explicit path with a comment describing this exact failure — "once per parse worker, flooding `ix map`". It was only ever applied to that one grammar. This extends it to the other twelve.

The optional grammars were never affected: `tryLoadGrammar` goes through `createRequire`, and CJS resolution does not warn. `tree-sitter` itself needs nothing — its `main` is a real filename.

## Verification

I checked the specifiers are interchangeable rather than assuming it. For all twelve, bare and explicit resolve to **identical export keys**, including the multi-grammar packages:

```
OK  tree-sitter-typescript   bare=[tsx,typescript]  explicit=[tsx,typescript]
OK  tree-sitter-php          bare=[php,php_only]    explicit=[php,php_only]
...                                                          ALL 12 IDENTICAL
```

Then built `dist` and parsed real snippets through it, since the failure is a runtime resolution one and the tests import from `src`:

```
OK  A.cs      entities=[A.cs,N,Greeter,Hi]
OK  a.ts      entities=[a.ts,Foo,bar]
OK  a.php     entities=[a.php,P,q]
OK  a.scala   entities=[a.scala,S,m]
OK  a.rs      entities=[a.rs,R,R,go]
OK  a.go      entities=[a.go,Go]
```

`tsc --noEmit` clean. The suite is unchanged at **6 failed / 271 passed** — byte-identical to `main`'s pre-existing failures, none of them touched by this.

## What I could not verify

**The warning itself does not reproduce here.** Neither Node 22.22 nor 24.19 emits DEP0151 for these packages, with or without `--pending-deprecation`; only the reporter's Node 23.9 did. Node 23 is odd-numbered and already out of support, so this may not fire on any currently supported runtime.

I still think the change is right: the deprecation is real, the packages genuinely have the shape it objects to, a future Node can reinstate it, and the repo had already decided this was worth fixing for css. But it is a fix I cannot demonstrate working — confirmation needs a `ix map` run on Node 23.

Worth noting separately that `package.json` sets `"engines": { "node": ">=22" }`, which admits 23. If Node 23 is not a runtime we intend to support, narrowing that is a different and possibly better conversation than this patch.

### DIFF
--- a/core-ingestion/src/index.ts
+++ b/core-ingestion/src/index.ts
@@ -16,7 +16,12 @@ import Parser from 'tree-sitter';
 //   @tree-sitter-grammars/tree-sitter-lua   optional  type=module    -> 1
 //   tree-sitter-css                         optional  type=module    -> 1
 //   the other 11 required grammars          required  type=commonjs  -> 0
-//   every other optional grammar            optional  type=commonjs  -> 0
+//   the other 11 optional grammars          optional  type=commonjs  -> 0
+//   tree-sitter-sas                         optional  type=module    -> 0
+//                                             (full-filename main + "exports")
+//
+// The two 11s are a coincidence, not a copy-paste: 12 required grammars minus
+// c-sharp, and 14 optional minus lua, css and sas.
 //
 // So the rule is `"type": "module"` + a directory `"main"` + no `"exports"`,
 // which today means exactly c-sharp (static, below), plus css and lua (dynamic,
--- a/core-ingestion/src/index.ts
+++ b/core-ingestion/src/index.ts
@@ -3,49 +3,62 @@ import * as crypto from 'node:crypto';
 import { createRequire } from 'node:module';
 // @ts-ignore — tree-sitter has no bundled types
 import Parser from 'tree-sitter';
-// Every grammar below is imported by the explicit path to its binding rather
-// than by the bare package name.
+// A grammar is imported by the explicit path to its binding ONLY when Node
+// would otherwise emit DEP0151 for it. That is a narrower set than it looks.
 //
-// All of these publish `"main": "bindings/node"` with no filename, extension or
-// `exports` field, so resolving the bare specifier from ESM makes Node fall
-// back to directory-index lookup and emit a DEP0151 deprecation warning. The
-// parse pool loads this module once per worker, so on a Node release that emits
-// DEP0151 the warnings arrive twelve at a time per worker and shred the `ix map`
-// progress bar, which shares stderr with them.
+// DEP0151 fires when an **ESM** resolution lands on a `"main"` with no filename
+// or extension. It is gated on the resolved format: a CommonJS package with
+// `"main": "bindings/node"` never warns, however it is imported. Almost every
+// grammar here is CJS, so almost none of them warn — measured per package in a
+// fresh process on Node 26:
 //
-// This is the same fix already applied to tree-sitter-css further down, and for
-// the same reason; it was only ever applied to that one grammar. The optional
-// grammars are unaffected because `tryLoadGrammar` goes through createRequire,
-// and CJS resolution does not warn. `tree-sitter` itself needs nothing — its
-// main is a real filename.
+//   tree-sitter-c-sharp                     type=module    -> 1 warning
+//   @tree-sitter-grammars/tree-sitter-lua   type=module    -> 1 warning
+//   tree-sitter-css                         type=module    -> 1 warning
+//   the other ten required grammars         type=commonjs  -> 0
 //
-// Verified equivalent: for all twelve, the bare and explicit specifiers resolve
-// to the same export keys, including the multi-grammar packages
-// (typescript -> {typescript, tsx}, php -> {php, php_only}).
+// So the rule is `"type": "module"` + a directory `"main"` + no `"exports"`,
+// which today means exactly c-sharp (static, below), plus css and lua (dynamic,
+// through tryImportGrammar further down). The parse pool loads this module once
+// per worker and every worker shares the parent's stderr, so each warning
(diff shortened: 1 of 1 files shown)


josephismikhail:
Reviewed and pushed a fix commit. The flood is real and the approach is right, but the audit behind the change had not been done — so it misses the warning that is actually firing and rewrites eleven imports that never fired one.

## DEP0151 is gated on the resolved format, not on `main`

A CommonJS package with `"main": "bindings/node"` never emits DEP0151, however it is imported. Measured per package in a fresh process, Node 26:

```
tree-sitter-c-sharp                     type=module    -> 1
@tree-sitter-grammars/tree-sitter-lua   type=module    -> 1
tree-sitter-css                         type=module    -> 1
tree-sitter-python                      type=commonjs  -> 0
tree-sitter-typescript                  type=commonjs  -> 0
tree-sitter-sas                         type=module    -> 0   (main is a real filename)
```

Of the twelve imports this PR rewrote, exactly one — `c-sharp` — does anything. The other eleven are CJS.

## The one still flooding is lua, and the comment explains why it was missed

`@tree-sitter-grammars/tree-sitter-lua` is `"type": "module"` with a directory `main` and no `exports`. The PR leaves it bare, on the reasoning that *"the optional grammars are unaffected because `tryLoadGrammar` goes through createRequire, and CJS resolution does not warn."*

Lua does not go through `tryLoadGrammar`. It goes through `tryImportGrammar`, which is a real ESM `import()` — the same helper `tree-sitter-css` already uses **with an explicit path, for exactly this reason, three lines below it**.

Built `dist` and loaded it, which is where this actually manifests:

```
PR as submitted : 1 DEP0151   (@tree-sitter-grammars/tree-sitter-lua)
with this commit: 0
```

One per parse worker, and the pool spawns `cpus-1` of them onto the same stderr the progress bar draws on. The reporter would have seen the bar still breaking and reopened this.

## The eleven no-ops go back to bare, deliberately

A subpath import is a dependency on a package's internal layout, and as a *static* import it fails hard — one that cannot resolve takes down this whole module and every language with it, not just the grammar concerned.

`tree-sitter-sas` is already the counterexample sitting in this same dependency set:

```
bare  'tree-sitter-sas'                        -> OK
path  'tree-sitter-sas/bindings/node/index.js' -> ERR_PACKAGE_PATH_NOT_EXPORTED
```

It ships `"exports": { ".": "./bindings/node/index.js" }`. That is the direction these packages are migrating, so eleven such dependencies bought nothing and bet against it. The three that remain (`c-sharp`, `lua`, `css`) each earn it by actually warning.

## The comment

Rewritten to state the real rule — `"type": "module"` **plus** a directory `main` **plus** no `exports` — with the measurements and the `sas` counterexample, so the next person extending this does not repeat the miss. The previous text asserted an audit that had not happened, which is what produced it.

---

Also correcting the PR description's diagnosis: this reproduces on **Node 26**, not only the reporter's 23.9. `c-sharp` warns on any Node implementing DEP0151 — the reporter's version was never the variable, the package's `type` field was. Which also means the "is Node 23 a runtime we support?" question at the end of the description, while worth asking on its own, is not what this hinges on.

Verified: `tsc` clean; all 15 language cases parse through the **built** module (js, ts, tsx, py, java, c, cpp, cs, go, rs, rb, php, scala, lua, css) — the compiled output matters here because a wrong specifier compiles cleanly and only fails at runtime; suite 6 failed / 271 passed, byte-identical to `main`, which I ran on `main` to confirm rather than assuming.

josephismikhail (approved):
Approving. Reviewed by independent agents across three rounds; every finding is fixed on the branch or recorded in the thread.

The approach was right, but the audit behind it had not been done: DEP0151 is gated on the resolved format, so of the twelve imports this rewrote only tree-sitter-c-sharp was ever emitting one — and the grammar still flooding, tree-sitter-lua, was left bare on reasoning that named the wrong loader. Measured on the built module: main = 2 warnings, PR as submitted = 1, now 0; and 16 -> 0 across an 8-worker pool, which is what actually shreds the progress bar.

The eleven no-op rewrites are back to bare specifiers. A static subpath import is a dependency on a package's internal layout that fails hard, and tree-sitter-sas in this same dependency set already ships an exports map where the subpath form throws ERR_PACKAGE_PATH_NOT_EXPORTED.

Verified: all 27 language snippets byte-identical across base/pre/post through the BUILT module (a wrong specifier compiles clean and only fails at runtime); all eleven reverts resolve to the same object by identity; the two remaining optional subpaths degrade gracefully when the platform has no prebuild; and tree-sitter-css/bindings/node/index.js has already shipped in v0.8.1 and v0.9.0-rc.1, so this packaging shape is proven through the prune and copy.

All CI green on 593ea6d.

Follow-up, not a blocker: the '(e.g. tree-sitter-sas has no win32 prebuild)' note further down is stale — sas loads and parses on win32 now.

PR #352, opened 2026-08-09
fix(install): stop the Windows installer dying on an 8.3 short TEMP path

### LINKED ISSUE #349: [bug] Windows installer — path with spaces
## Description

The Windows installer fails during CLI installation when the Windows user profile/TEMP path contains a space.

In my case, the user profile is `C:\Users\Win 10`. The CLI archive downloads and extracts successfully, but the installer then resolves part of the path as `C:\Users\WIN10\~1` and fails because that path does not exist.

Using a TEMP directory without spaces allows the installation to complete successfully.

## Steps to reproduce

1. Use a Windows account with a profile path containing a space, such as `C:\Users\Win 10`.
2. Run:

   ```powershell
   irm https://ix-infra.com/install.ps1 | iex
   ```
3. Wait for the CLI installation stage after the archive has been downloaded and extracted.

## Expected behavior

The installer should correctly handle Windows user profile and temporary paths containing spaces and complete the CLI installation.

## Actual behavior

The archive downloads and extracts successfully, but the installer fails while resolving the temporary path.

Changing `TEMP` and `TMP` to a path without spaces, such as `C:\Temp`, allows the installation to succeed.

## CLI output

```text
[ok] Downloaded to C:\Users\WIN10\~1\AppData\Local\Temp\ix-0.8.1-windows-amd64.zip
Extracting CLI...
[ok] Extraction complete

Ix installer failed.
Error:
An object at the specified path C:\Users\WIN10\~1 does not exist.
```

## Environment

* OS: Windows
* Node: v24.18.0
* Ix version: 0.8.1

KageBinary:
Refs #349. *(Was `Fixes #349` - downgraded so merging does not auto-close it, matching the "still needs a real Windows run to confirm" note below. See Review follow-up at the end.)*

## What the user saw

An install on a profile at `C:\Users\Win 10` downloads and extracts fine, then dies naming a path it was never asked to write:

```text
[ok] Downloaded to C:\Users\WIN10~1\AppData\Local\Temp\ix-0.8.1-windows-amd64.zip
Extracting CLI...
[ok] Extraction complete

Ix installer failed.
Error:
An object at the specified path C:\Users\WIN10~1 does not exist.
```

## Why

Not the space itself — the 8.3 short path it produces. Windows sets `TEMP` to the short form of the profile path when that path contains a space, so `C:\Users\Win 10` becomes `C:\Users\WIN10~1`. PowerShell's FileSystem provider cannot resolve a short-name segment, and when it fails it reports the first segment it could not resolve rather than the file it was handed — which is why the error names the profile root and looks unrelated to the install.

`curl.exe` and `Expand-Archive` go straight to Win32 and succeed on that same path. That is the whole reason the failure lands *after* `[ok] Extraction complete`: everything that talks to Win32 directly worked, and the first provider cmdlet to touch the temp path did not.

The live script at `https://ix-infra.com/install.ps1` is byte-identical to `main`, so this is current code, not something already fixed since 0.8.1.

## The fix

Route the two scratch files — the CLI zip and the compose pull log — through `IX_HOME` rather than `TEMP`. `IX_HOME` comes from `USERPROFILE` in its long form, so this sidesteps the class instead of expanding short names at each call site.

Both take the `.cli-staging-` prefix, because leaving `TEMP` also leaves the OS's own cleanup: `sweepUpgradeOrphans` already reclaims `.cli-staging-*` out of `IX_HOME`, so an installer killed mid-run still leaves nothing permanent behind. Deliberately **not** `.cli-backup-` — that prefix is a *recovery* candidate in the sweep, and a leftover zip would be renamed over `~/.ix/cli` on the next upgrade.

## Testing

`install.ps1` cannot be parsed or run here (no PowerShell on Linux, and CI's `windows-2022` leg packages the CLI without running the installer), so the two properties the script now depends on are pinned in `upgrade-archive-shape.test.ts` instead — the sweep had only ever been handed directories, never files:

- leftover `.cli-staging-<pid>.zip` and `.cli-staging-pull-<pid>.log` are reclaimed
- a stray `.cli-staging-*` file is never mistaken for an install worth recovering

19/19 pass in that file.

Still needs a real Windows run to confirm end to end. Worth folding into the existing rc.2 Windows check rather than a separate pass.

## Note on release timing

This does not need to block the v0.9.0 GA tag. `install.ps1` is fetched from `main`, not from the release zip, so it reaches users as soon as it merges regardless of what is tagged.

---

## Review follow-up (independent pass + fixes pushed)

Two commits added on top. The code change and both of Ian's tests stand — the tests were mutation-checked and each kills exactly the one test it claims to protect.

**1. Scratch leaked on every failure path.** Leaving `TEMP` also left the OS's cleanup, and the only delete of the zip was on the success path — five exits left it behind. `sweepUpgradeOrphans` is a weaker backstop than the original comment implied: it runs only from `ix upgrade`, and only when an update is available, so a *first* install that dies at the download strands a partial multi-MB zip in `~/.ix` with no `ix` on the machine to ever reclaim it. Cleanup now hangs off `Write-Err`, the single exit for every failure below it. Verified on Windows PowerShell 5.1 and pwsh 7 against the real function extracted from the script: both files reclaimed, install directory untouched, no throw when the paths are still unassigned (an early `Write-Err`) or when the file is already gone.

**2. The mechanism in the checked-in comment did not survive checking**, so it has been rewritten to separate what the report establishes from what it does not. Specifically:

- `Expand-Archive` resolves through the FileSystem provider (`Resolve-Path` in `Microsoft.PowerShell.Archive`), so "curl.exe and Expand-Archive go straight to Win32" is not the split.
- The provider resolves 8.3 segments fine on both 5.1 and 7 — `Test-Path C:\PROGRA~1` returns `True`.
- The reporter's own transcript shows `Test-Path -LiteralPath $tmp` and `Get-Item -LiteralPath $tmp` **succeeding** on the short TEMP path: both run before `Expand-Archive`, and extraction completes before the error appears. So "the first provider cmdlet to touch the temp path failed" cannot be what happened.
- The error string `An object at the specified path ... does not exist.` could not be reproduced from ~30 cmdlets, including every one this script calls on those paths — they report `Cannot find path` instead.
- Issue #349 quotes the path as `C:\Users\WIN10\~1`, not `C:\Users\WIN10~1`. Probably a copy artifact, but it was silently normalised in the PR body, and the 8.3 reading depends on that character.

**The fix is still right, which is why this is merging.** The reporter confirmed that pointing `TEMP`/`TMP` at a path without a space makes the same install succeed, so TEMP is implicated regardless of mechanism. The likelier reading is that 8.3 alias *creation* is disabled on that volume while `TEMP` still carries a stale short path — in which case nothing resolves it, and moving off TEMP fixes that too. Both readings are repaired by this change; only the explanation differs.

**Left open deliberately:**

- `ix upgrade` still stages through `os.tmpdir()` (`upgrade.ts:771`, `:1048`), which on Windows is `%TEMP%` verbatim. If this class is real, it is unfixed there.
- #349 stays open until the reporter confirms on the affected machine. That is why this is `Refs` rather than `Fixes` — it matches the "still needs a real Windows run to confirm end to end" note above.

*Pre-existing, unchanged, flagged for the record:* `install.ps1` has no BOM and 1203 non-ASCII bytes, so `powershell.exe -File` reports 16 parse errors on Windows PowerShell 5.1. Identical on base and head — neither Ian's commit nor mine moved it, and the documented `irm | iex` path is unaffected because the server sends `charset=utf-8`.

### DIFF
--- a/ix-cli/src/cli/__tests__/upgrade-archive-shape.test.ts
+++ b/ix-cli/src/cli/__tests__/upgrade-archive-shape.test.ts
@@ -253,6 +253,41 @@ describe("sweepUpgradeOrphans", () => {
     expect(readFileSync(join(installDir, "cli", "dist", "cli", "main.js"), "utf-8")).toBe("// live\n");
   });
 
+  // install.ps1 downloads to `.cli-staging-<pid>.zip` and tees the compose pull
+  // output to `.cli-staging-pull-<pid>.log`, both inside IX_HOME, because
+  // Windows hands it a TEMP path in 8.3 short form whenever the profile
+  // contains a space and PowerShell's provider cannot resolve one. Moving them
+  // out of TEMP also moved them out from under the OS's own cleanup, so the
+  // installer now leans on this sweep to be the thing that reclaims them after
+  // a run that was killed before its own delete. They are files rather than
+  // directories, which the prefix match above had never been given.
+  it("reclaims the installer's leftover scratch files, not just directories", () => {
+    const installDir = join(root, "cli");
+    mkdirSync(join(installDir, "cli", "dist", "cli"), { recursive: true });
+    writeFileSync(join(installDir, "cli", "dist", "cli", "main.js"), "// live\n");
+    writeFileSync(join(root, ".cli-staging-333.zip"), "PK");
+    writeFileSync(join(root, ".cli-staging-pull-333.log"), "pulling...\n");
+
+    sweepUpgradeOrphans(root, installDir);
+
+    expect(existsSync(join(root, ".cli-staging-333.zip"))).toBe(false);
+    expect(existsSync(join(root, ".cli-staging-pull-333.log"))).toBe(false);
+    expect(readFileSync(join(installDir, "cli", "dist", "cli", "main.js"), "utf-8")).toBe("// live\n");
+  });
+
+  // The zip must not be named `.cli-backup-*`: that prefix is a recovery
+  // candidate, so a leftover would be renamed *over* ~/.ix/cli on the next
+  // upgrade and replace the install with a zip file. This pins the reason for
+  // the name install.ps1 actually uses.
+  it("does not treat a stray file as an install worth recovering", () => {
+    const installDir = join(root, "cli");
+    writeFileSync(join(root, ".cli-staging-444.zip"), "PK");
+
+    sweepUpgradeOrphans(root, installDir);
+
+    expect(existsSync(installDir)).toBe(false);
+  });
+
   it("is a no-op when IX_HOME does not exist yet", () => {
     expect(() => sweepUpgradeOrphans(join(root, "nope"), join(root, "nope", "cli"))).not.toThrow();
   });
--- a/scripts/install/install.ps1
+++ b/scripts/install/install.ps1
@@ -25,10 +25,18 @@ $NodeMinMajor = 22
 # Installer scratch files live under $IxHome, never under $env:TEMP.
 #
 # #349: on a profile at `C:\Users\Win 10`, Windows hands the installer a TEMP
-# path in 8.3 short form and the run dies after extraction with
-# `An object at the specified path C:\Users\WIN10~1 does not exist.`
-# The reporter confirmed that pointing TEMP and TMP at a path without a space
-# lets the same install finish, so the short TEMP path is what breaks it.
+# path in 8.3 short form and the run dies after extraction. The reporter
+# confirmed that pointing TEMP and TMP at a path without a space lets the same
+# install finish, so the short TEMP path is what breaks it.
+#
+# The issue quotes the error verbatim as
+#   An object at the specified path C:\Users\WIN10\~1 does not exist.
+# -- note the backslash before `~1`. Probably a copy artifact, since
+# `C:\Users\WIN10~1` is the 8.3 alias of `C:\Users\Win 10` and a literal `~1`
+# directory makes no sense, but it has not been confirmed with the reporter and
+# the 8.3 reading depends on that one character. Quoted as written rather than
+# normalised, because the whole point of the block below is which parts of this
+# are established.
 #
 # $IxHome comes from USERPROFILE, which is the long form. $Staging was already
 # there, so moving these two takes the script off TEMP completely.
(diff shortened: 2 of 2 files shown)


KageBinary (commented):
The diagnosis is right and the fix is at the right altitude — routing scratch through `$IxHome` (long form, from `USERPROFILE`) sidesteps the whole short-name class rather than expanding 8.3 paths at each call site.

Checked the ordering, since writing into `$IxHome` only works if it exists by then:

- `$pullLog` (:209) — `New-Item -Force -Path $ComposeDir` at :197 creates `$IxHome\backend`, so the parent exists.
- `$tmp` (:287) — `New-Item -Force -Path $IxBin` at :290 runs before `curl.exe` at :295.

Both fine. `-LiteralPath` also keeps `.cli-staging-<pid>.zip` and the `.cli-staging-<pid>` directory (:330) from colliding on the `Remove-Item -Recurse` at :331.

The prefix choice is load-bearing and the second test pins exactly why: `sweepUpgradeOrphans` calls `recover(".cli-backup-", installDir, ...)` at `upgrade.ts:612`, so a leftover zip under that prefix really would get renamed over `~/.ix/cli` and replace the install with a zip file. Good call documenting the reason in the test rather than just the name.

One note, not blocking: the installer now leaves these files in `IX_HOME` *while it is still using them*, and `ix upgrade` sweeps that prefix on startup. A concurrent `ix upgrade` can delete the zip in the window between `curl` (:295) and `Expand-Archive` (:340). Narrow, requires concurrency, and it is the cost of the sweep-based cleanup you deliberately opted into — worth a comment at the `$tmp` assignment if you touch this again.

All checks green. Approving.

josephismikhail (approved):
Independent review passes on your commit and on each of my four follow-ups. Merging — but read the note on #349 below, because I changed the auto-close.

**Your code change and both tests stand.** Mutation-checked: making the sweep directory-only kills `reclaims the installer's leftover scratch files, not just directories`; making `.cli-staging-` a recovery candidate kills `does not treat a stray file as an install worth recovering`. One mutation, one death, each the test that claims it — and the second mutation's stdout (`Recovered the previous install from .cli-staging-444.zip`) independently proves your `.cli-backup-` warning is real, since `recover()` never checks the entry is a directory.

**The mechanism in the comment did not survive checking, so I rewrote it.** Not to be pedantic — it lands as permanent in-repo documentation and would send the next reader the wrong way:

- `Expand-Archive` resolves *through* the provider (`Resolve-Path` in `Microsoft.PowerShell.Archive`), so "curl.exe and Expand-Archive go straight to Win32" isn't the split. (The 5.1 module is UTF-16, so a naive grep returns 0 hits and looks like confirmation.)
- The provider resolves 8.3 fine on 5.1 **and** 7 — built a real `WIN10~1` path and ran the lot: `Test-Path`, `Get-Item`, `Expand-Archive`, `Remove-Item` all succeed.
- Your own transcript shows `Test-Path -LiteralPath $tmp` and `Get-Item -LiteralPath $tmp` **succeeding** — they run before `Expand-Archive`, and the paste reaches `[ok] Extraction complete`. So "the first provider cmdlet to touch the temp path failed" can't be what happened.
- That error string is `FileSystemProviderStrings.ItemDoesNotExist`; 24 path-taking cmdlets were run against a missing path on 5.1 and none produce it.
- #349 quotes it as `C:\Users\WIN10\~1`, with a backslash. Almost certainly a copy artifact — confirmed `WIN10~1` really is the 8.3 alias of `Win 10` — but the 8.3 reading rests on that character and the PR body normalised it silently.

**The fix is still right**, which is why this is merging: the reporter confirmed that moving TEMP off a spaced path makes the same install succeed. The likelier reading is 8.3 alias *creation* disabled on that volume with a stale short TEMP, which this repairs identically. Only the explanation changes.

**I also fixed a leak your change introduced** — five exits left the zip in `~/.ix`, and the sweep is a weaker backstop than it looks: it runs only from `ix upgrade`, only when an update exists, so a *first* install that dies at the download strands a multi-MB partial with no `ix` to ever reclaim it.

That fix then took three rounds, all in the recovery path, all caught by review rather than by me:

1. Reading `$script:tmp` was wrong — README ships this as `irm | iex`, so the body runs in the **caller's** scope. Before the assignment those names are the *user's* variables, and `$pullLog` is assigned inside the `else` of `if (Test-Healthy)`, so on every re-run it's never assigned and all twelve `Write-Err` sites were exposed. It deleted the user's files and didn't reclaim the real scratch. Same class as your `$Matches` note.
2. Rebuilt from `$IxHome` + `$PID`, but left `Test-Path` unguarded — under `ErrorActionPreference = Stop` a detached `IX_HOME` drive turns a clean `[error] …` exit into that plus a spurious `Ix installer failed. / Cannot find drive.` trap block.
3. Wrapped the whole body. Necessary, not tidier: isolating the statements shows **`Join-Path`** is what throws on the detached drive, before `Test-Path` is reached — so guarding `Test-Path` alone would not have fixed it.

Verified end-to-end on Windows PowerShell 5.1 and pwsh 7: user files survive, real scratch reclaimed, `$Staging` and the `.cli-backup-<pid>` copy of the previous CLI untouched, no throw across 15 pathological `IX_HOME` values. Encoding unmoved — 1203 non-ASCII, no BOM, 16 parse errors on 5.1, same as base.

**I downgraded `Fixes #349` to `Refs #349`** so merging doesn't auto-close it. That matches your own "still needs a real Windows run to confirm end to end" — the diagnosis isn't confirmed, and the reporter should close it from the affected machine.

Two follow-ups, neither for this PR: `ix upgrade` still stages through `os.tmpdir()` (`upgrade.ts:771`, `:1048`), which on Windows is `%TEMP%` verbatim — unfixed if this class is real. And `install.ps1` has no BOM with 1203 non-ASCII bytes, so `powershell -File` gives 16 parse errors on 5.1; pre-existing and invisible to the documented `irm | iex` path, which the server saves with `charset=utf-8`.

PR #356, opened 2026-08-09
fix(status): keep staleness scoped to each workspace

### LINKED ISSUE #353: [bug] ix status uses another workspace timestamp and hides stale files
## Description

## Steps to reproduce

   ```sh
   git init -q /tmp/ix-status-a
   git init -q /tmp/ix-status-b
   ```

   ```sh
   cd /tmp/ix-status-a
   ix map . --silent
   ```

   `staleFiles` is `1`.

   ```sh
   cd /tmp/ix-status-b
   ix map . --silent
   cd /tmp/ix-status-a
   ```

## Expected behavior

## Actual behavior

## CLI output

Before mapping B:

```json
{
  "currentRev": 1,
  "staleFiles": 1,
}
```

After mapping B:

```json
{
  "currentRev": 2,
  "staleFiles": 0,
  "sampleChangedFiles": []
}
```

## Environment

- OS: macOS 26.6.1
- Node: v26.7.0
- Ix version: 0.8.1

Hiro-Chiba:
Fixes #353.

## Summary

Keep staleness metadata in the existing workspace-root-specific ingest cache so mapping another workspace cannot make the current workspace appear up to date.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Extend the per-workspace mtime cache with the last successful ingest time and revision while retaining compatibility with legacy cache files.
- Make `status` and symbol freshness checks use that local baseline instead of graph-wide patch and stats reads.
- Preserve the previous baseline after parse or commit failures and fall back to ingest time for files not represented in the cache.
- Add regression coverage for two workspaces, legacy caches, unmapped workspaces, uncached files, and failed maps.

## Validation

- `cd ix-cli && npm test` (48 files, 631 tests, parser smoke passed)
- `npm run typecheck`
- `npm run build`

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/stale-workspace.test.ts
+++ b/ix-cli/src/cli/__tests__/stale-workspace.test.ts
@@ -1,0 +1,190 @@
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import * as fs from "node:fs";
+import * as os from "node:os";
+import * as path from "node:path";
+
+import type { IxClient } from "../../client/api.js";
+import { ingestMtimeCachePath } from "../config.js";
+import { loadIngestBaseline } from "../ingest-baseline.js";
+import { persistIngestBaselineIfClean } from "../commands/ingest.js";
+import { detectStaleFiles } from "../stale.js";
+
+let home: string;
+let savedHome: string | undefined;
+let savedProfile: string | undefined;
+
+const noPatchReads = {
+  listPatches: async () => {
+    throw new Error("global patches must not be read");
+  },
+  stats: async () => {
+    throw new Error("global stats must not be read");
+  },
+} as unknown as IxClient;
+
+function writeSource(root: string, name: string, source: string): string {
+  fs.mkdirSync(root, { recursive: true });
+  const filePath = path.join(root, name);
+  fs.writeFileSync(filePath, source);
+  return filePath;
+}
+
+function moveMtimeForward(filePath: string): number {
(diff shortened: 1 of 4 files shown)


KageBinary (approved):
Reviewed and verified locally (fork PRs don't get a CI run here, so `CI Passed` will never appear on this branch): `tsc --noEmit` clean, `eslint src` clean, full suite green (631 tests), and it merges onto `main` alongside #358/#359 with no conflicts.

The design is the right one. Moving the baseline into `ingest-baseline.ts` with an explicit `lastIngestAt` in the record — rather than leaning on the cache file's own mtime — survives backups, rsync and stray `touch`es. And injecting a client whose `listPatches`/`stats` *throw* is a much better test than asserting around the call: it proves the global read is gone rather than that it happened to return something unused.

Two follow-ups I'll open rather than hold this up:

1. `isFileStale` has the identical bug and isn't covered here — it's called from `explain.ts`, `locate.ts`, `read.ts` and `explain/facts.ts`, all still comparing against the newest patch anywhere on the machine.
2. This drops the `!opts.force` guard on persisting the baseline while keeping the comment above it that describes that guard. I think dropping it is correct now that the baseline is the source of truth for staleness rather than just a perf cache — a clean `--force` map *should* update it — but the comment needs to stop describing a guard that isn't there.

PR #357, opened 2026-08-09
fix(map): suppress progress outside terminals

### LINKED ISSUE #354: [bug] ix map emits carriage-return progress frames when stderr is not a TTY
## Description

`ix map` starts an 80 ms progress interval whenever the format is `text`, even when stderr is redirected and is not a TTY. Every refresh is retained as another carriage-return frame in CI logs and agent tool output.

Current `main` checks only `opts.format === 'text'` in `ix-cli/src/cli/commands/ingest.ts`; it does not check `process.stderr.isTTY`.

## Steps to reproduce

1. Create a one-file repository.

   ```sh
   mkdir /tmp/ix-progress-repro
   git init -q /tmp/ix-progress-repro
   printf 'export const value = 1;\n' > /tmp/ix-progress-repro/index.js
   ```

2. Redirect both output streams and count carriage returns.

   ```sh
   cd /tmp/ix-progress-repro
   ix map . >stdout.log 2>stderr.log
   wc -c stdout.log stderr.log
   LC_ALL=C tr -cd '\r' <stderr.log | wc -c
   ```

## Expected behavior

When stderr is not a TTY, Ix should skip intermediate progress frames and emit only a final summary or normal line-oriented diagnostics.

## Actual behavior

Even a sub-second one-file map produced 653 bytes on stderr with 9 carriage-return frames. The interval runs every 80 ms, so redirected output grows continuously with map duration.

## CLI output

```text
      99 stdout.log
     653 stderr.log
     752 total
       9
```

`ix map --silent` and machine-readable formats avoid the interval, but default text output does not detect the non-interactive destination.

## Environment

- OS: macOS 26.6.1
- Node: v26.7.0
- Ix version: 0.8.1

Hiro-Chiba:
Fixes #354.

## Summary

Only render the ingest progress interval when text output is going to an interactive stderr terminal. Redirected and agent-captured output remains line-oriented instead of retaining carriage-return refresh frames.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Gate the text progress interval on `process.stderr.isTTY === true`.
- Add regression coverage for interactive, redirected, undefined-TTY, JSON, and LLM output modes.

## Validation

- `cd ix-cli && npm test` (48 files, 627 tests, parser smoke passed)
- `npm run typecheck`
- `npm run build`
- Redirected local CLI smoke test completed with exit 0, zero stderr bytes, and zero carriage returns

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/ingest-progress.test.ts
+++ b/ix-cli/src/cli/__tests__/ingest-progress.test.ts
@@ -1,0 +1,16 @@
+import { describe, expect, it } from "vitest";
+
+import { shouldRenderIngestProgress } from "../commands/ingest.js";
+
+describe("shouldRenderIngestProgress", () => {
+  it("renders text progress only when stderr is an interactive terminal", () => {
+    expect(shouldRenderIngestProgress("text", true)).toBe(true);
+    expect(shouldRenderIngestProgress("text", false)).toBe(false);
+    expect(shouldRenderIngestProgress("text", undefined)).toBe(false);
+  });
+
+  it("does not render progress for machine-readable formats", () => {
+    expect(shouldRenderIngestProgress("json", true)).toBe(false);
+    expect(shouldRenderIngestProgress("llm", true)).toBe(false);
+  });
+});
--- a/ix-cli/src/cli/commands/ingest.ts
+++ b/ix-cli/src/cli/commands/ingest.ts
@@ -351,6 +351,13 @@ function renderProgressLine(phase: string, current: number, total: number): stri
   return `  ${phase.padEnd(8)}  ${bar}  ${pctStr}  ${count}`.padEnd(PROG_LINE_WIDTH);
 }
 
+export function shouldRenderIngestProgress(
+  format: string,
+  stderrIsTTY: boolean | undefined,
+): boolean {
+  return format === 'text' && stderrIsTTY === true;
+}
+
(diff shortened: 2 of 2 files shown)


KageBinary:
Superseded by #359, which is now merged — thank you for this, and sorry it went the other way.

Your diagnosis was right and #359 is the same fix; it just also covers three more places the same defect lives, which is why we took that one:

| site | gate before |
|---|---|
| ingest bar | `format === 'text'` — the one you found |
| `ix map` bar | `!machineFormat && !silent` |
| `ix reset` spinner | nothing at all — animated even when redirected |
| `printUpdateNotice` | unconditional `\r` + 80 spaces + `\r`, on ordinary commands with no bar to erase |

One thing worth flagging from your version: gating `interval` also drops the `Ingested in Xs` summary, because the teardown block is `if (interval) { erase…; summary }`. Your issue asked for "only a final summary or normal line-oriented diagnostics" — so #359 splits those two, keeping the summary on the format and only the redrawing frames on the TTY check.

Your `shouldRenderIngestProgress(format, stderrIsTTY)` signature is arguably the nicer shape — passing the TTY in rather than reading it inside. Noted for next time we touch it.

Your other two (#356, #358) are both merged as-is.

PR #358, opened 2026-08-09
fix(view): report the running visualizer port

### LINKED ISSUE #355: [bug] ix view reports the requested port for an already-running server on another port
## Description

When System Compass is already running, `ix view -p <different-port> start` exits successfully and prints the newly requested port instead of the port used by the existing process. The printed URL is unreachable.

The current implementation persists the PID and scope, but not the launch port. The already-running branch formats its URL from the current invocation's `-p` value.

## Steps to reproduce

1. Stop any existing visualizer and start one on port 19123.

   ```sh
   ix view stop
   ix view -p 19123 start --no-open
   curl -sS -o /dev/null -w '%{http_code}\n' http://127.0.0.1:19123/
   ```

2. Start again with another port.

   ```sh
   ix view -p 19124 start --no-open
   echo $?
   curl -sS http://127.0.0.1:19124/
   ```

## Expected behavior

Ix should either report the actual URL of the existing server, `http://localhost:19123`, or reject the mismatched port and explain how to restart.

## Actual behavior

The second command exits `0` and reports `http://localhost:19124`, but port 19124 refuses the connection. The original server remains available on port 19123.

## CLI output

```text
[ok] Visualizer started (PID 65764)
  http://localhost:19123

  http://localhost:19124

second command exit: 0
HTTP on 19123: 200
HTTP on 19124: connection refused
```

## Environment

- OS: macOS 26.6.1
- Node: v26.7.0
- Ix version: 0.8.1

Hiro-Chiba:
Fixes #355.

## Summary

Persist the visualizer's launch port and report that port when `ix view start` finds an existing process, instead of formatting an unreachable URL from the new invocation's `-p` value.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Store the launch port alongside the existing PID and scope state.
- Recover the port from the generated server script for processes started by an older CLI.
- Avoid printing a guessed URL when legacy state cannot reveal the port.
- Remove PID, scope, and port state together for stale processes and `ix view stop`.

## Validation

- `cd ix-cli && npm test` (48 files, 631 tests, parser smoke passed)
- `npm run typecheck`
- `npm run build`
- Real-process smoke test: started on 19123, repeated start with 19124 reported 19123, and stop removed all state files

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/view-running-port.test.ts
+++ b/ix-cli/src/cli/__tests__/view-running-port.test.ts
@@ -1,0 +1,153 @@
+import { EventEmitter } from "node:events";
+import { existsSync, mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { Command } from "commander";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const mocks = vi.hoisted(() => ({
+  execSync: vi.fn(),
+  spawn: vi.fn(),
+  createConnection: vi.fn(),
+}));
+
+vi.mock("child_process", () => ({
+  execSync: mocks.execSync,
+  spawn: mocks.spawn,
+}));
+
+vi.mock("net", () => ({
+  createConnection: mocks.createConnection,
+}));
+
+let ixHome: string;
+let originalIxHome: string | undefined;
+
+const pidFile = () => join(ixHome, "compass.pid");
+const scopeFile = () => join(ixHome, "compass.scope");
+const portFile = () => join(ixHome, "compass.port");
+const serverScriptFile = () => join(ixHome, "tmp", "compass-server.js");
+
+beforeEach(() => {
+  vi.resetModules();
+  vi.clearAllMocks();
+
+  originalIxHome = process.env.IX_HOME;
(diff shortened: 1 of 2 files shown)


KageBinary (approved):
Reviewed and verified locally (fork PRs don't get a CI run here): `tsc --noEmit` clean, `eslint src` clean, full suite green (631 tests), merges onto `main` alongside #356/#359 with no conflicts.

Picking this over my own #360, and the deciding factor is the legacy recovery — parsing `const PORT = N;` out of the old `compass-server.js` and migrating it into the port file means anyone with a visualizer already running when they upgrade gets a real answer instead of being told to restart. Mine just gave up in that case. `removeCompassState()` and the `pid <= 0` hardening in `readAlivePid` are both improvements on what was there.

The tests are the other reason: driving the real command against a temp `IX_HOME` with `spawn`/`net` mocked, asserting output *and* file state, catches wiring mistakes that my pure-function tests would have sailed straight past.

Two small things from #360 I will port as a follow-up rather than block on:

1. When `-p` does not match the running port, say so. Right now the correct URL prints but the user never learns their `-p` was ignored — worth using `getOptionValueSource("port")` so an explicit `-p 8080` warns and a bare `ix view` stays quiet.
2. `ix view status` could print the URL now that the port is recorded.

PR #360, opened 2026-08-09
fix(view): report where the visualizer is actually serving, not what was asked for

### LINKED ISSUE #355: [bug] ix view reports the requested port for an already-running server on another port

KageBinary:
Fixes #355.

## The problem

`ix view -p <other-port>` against an already-running server exits 0 and prints the port it was just *asked* for. That URL refuses connections while the real server keeps serving on the port it was launched with, and nothing in the output suggests the link is wrong.

The PID and the scope were both persisted at launch. The port was not — so the already-running branch had nothing to report and echoed the request back as though it were an answer. `-p` asks where to *start* a server, and in that branch there is nothing to start.

## The fix

Persist the port beside the PID and report that. It's written at the same moment as the pid, so the two can only ever describe the same process — deriving it later from a live socket would reintroduce exactly the guess this replaces.

**Printing the right URL alone wasn't enough.** That would fix the dead link but leave the user believing `-p` had moved the server, which is the other half of the report ("exits successfully"). A mismatch now says so and gives the two commands that would move it.

That warning fires only when the port was genuinely requested, using `getOptionValueSource` rather than a comparison against the default. Someone who explicitly types `-p 8080` at a server running elsewhere is asking precisely the question the warning answers — and `opts().port` reads `8080` whether or not they typed it.

**An instance started before the port file existed** is still running and still serving, and there is no honest way to say where. It says that, rather than guessing. One restart records it.

## Two things that followed

- `ix view status` printed only a PID. Now that the port is known, it can answer the question people run it to ask.
- The stale-pid cleanup and `stop` both clear the new file, so a reused pid can't inherit a previous instance's port.

## Verification

The decision is a pure function, so the bug is under test rather than the file read — inline in the command action it could only have been tested by launching a detached server. Five cases, including the two silence cases and the unknown-port case.

Run against a real server as well, following the issue's steps:

```text
[ok] Visualizer started (PID 100158)
  http://localhost:19123          <- serving here

$ ix view -p 19124 start --no-open
[ok] Visualizer is already running (PID 100158)
  http://localhost:19123
[!] You asked for port 19124, but it is serving on 19123.
    Run 'ix view stop' then 'ix view -p 19124' to move it.

HTTP on 19123: 200
HTTP on 19124: connection refused
```

It used to print `http://localhost:19124` there. 630 tests pass, `tsc --noEmit` clean.

### DIFF
--- a/ix-cli/src/cli/__tests__/view-running-port.test.ts
+++ b/ix-cli/src/cli/__tests__/view-running-port.test.ts
@@ -1,0 +1,48 @@
+import { describe, expect, it } from "vitest";
+
+import { runningInstanceLines } from "../commands/view.js";
+
+// `ix view -p <other-port>` against an already-running server exited 0 and
+// printed the port it had just been *asked* for, which refuses connections,
+// while the real server kept serving on the port it was launched with. The PID
+// and scope were both persisted; the port was not, so this branch had nothing
+// to report and echoed the request back as though it were an answer.
+describe("runningInstanceLines", () => {
+  it("reports where the server actually is, not what was asked for", () => {
+    const lines = runningInstanceLines(19123, 19124, true);
+    expect(lines[0]).toBe("  http://localhost:19123");
+    // The requested port must never appear as a URL — that is the whole bug.
+    expect(lines.some((l) => l.includes("http://localhost:19124"))).toBe(false);
+  });
+
+  it("explains the mismatch instead of silently ignoring the flag", () => {
+    // Silently printing the right URL would fix the broken link but leave the
+    // user believing -p had moved the server, which is the other half of the
+    // report: the command "exits successfully".
+    const lines = runningInstanceLines(19123, 19124, true);
+    expect(lines.join("\n")).toContain("You asked for port 19124");
+    expect(lines.join("\n")).toContain("ix view stop");
+  });
+
+  it("stays quiet when the running port is the one that was asked for", () => {
+    expect(runningInstanceLines(8080, 8080, true)).toEqual(["  http://localhost:8080"]);
+  });
+
+  it("stays quiet when no port was asked for", () => {
+    // The default is 8080, so a plain `ix view` against a server on 19123 must
+    // not nag: the user expressed no preference. This is why the caller passes
+    // getOptionValueSource rather than comparing against the default — someone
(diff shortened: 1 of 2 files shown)


KageBinary:
Closing in favour of #358 by @Hiro-Chiba, which is merged.

Theirs recovers the port of an already-running visualizer by parsing `const PORT = N;` out of the old `compass-server.js` and migrating it forward, so anyone who upgrades with a server already up gets a real URL instead of this PR's "not recorded, restart it". Their tests also drive the actual command against a temp `IX_HOME`, which would have caught a missing `writeFileSync(PORT_FILE, …)` — these pure-function tests would not have.

Two pieces from here are still worth having and will follow as a separate PR:

1. The `-p` mismatch warning. #358 prints the running port but says nothing about the port you asked for, so `ix view -p 19124` against a server on 19123 silently gives you 19123. `runningInstanceLines` plus `getOptionValueSource("port") === "cli"` distinguishes an explicit `-p 8080` from the 8080 default, so the warning fires only when someone actually asked.
2. `ix view status` printing the URL.

PR #361, opened 2026-08-09
fix(status): decide staleness from this workspace's ingest, not the newest patch

### LINKED ISSUE #353: [bug] ix status uses another workspace timestamp and hides stale files

KageBinary:
Fixes #353.

## The problem

`ix status --root <workspace>` scans the right directory, then compares every file against the newest patch in the **entire backend**. Mapping an unrelated repo advances that timestamp, so a genuinely modified file in the first workspace falls behind it and is reported up to date — `staleFiles` goes 1 → 0 with nothing in that workspace having changed.

## Why the suggested fix isn't available

The issue proposes passing a workspace to `listPatches`. That isn't possible today, and it's worth recording why, because it looks like a one-line change:

- `GET /v1/patches` accepts **only** `limit` (`PatchRoutes.scala:31`). Verified against a live backend: a real workspace id, a nonsense one, and none at all return byte-identical results, and the `X-Ix-Workspace` header the Compass proxy uses is ignored on this route too.
- More fundamentally, **a patch document has no workspace field at all** — not omitted from the query, absent from the schema. I dumped one from a live backend to check.

So scoping the patch log would mean writing the field at commit time, a backend release, and a decision about the existing patches that would still have no workspace. That's a backend change in another repo. This is fixable in the CLI without any of it.

## The fix

The per-root ingest mtime cache is already exactly the missing record: written at the end of each ingest *for that root*, keyed on the root, holding every file's mtime as of that ingest.

Staleness now comes from it, using **the same test ingest applies** when deciding what to re-parse — exact mtime inequality, absent-means-new. So `ix status` and `ix map` can no longer disagree about whether a given file is stale, which they previously could in both directions.

`lastIngestAt` becomes that cache file's own mtime — the one timestamp on the machine that means "this workspace, last ingest".

`currentRev` **stays global on purpose.** It's the graph's revision counter, shared by every workspace. Reporting it was never the bug; reporting another workspace's staleness was. Worth flagging since the issue mentions the revision too.

## Two things that came with it

- **`isFileStale` had the same defect**, and is used by `explain`, `locate`, `read` and `facts`. It no longer takes a client at all — it was doing an HTTP fetch of the newest patch on *every call, once per file*, to run the same cross-workspace comparison.
- **The cache loader moved to `config.ts`**, beside the path helper whose comment already says all consumers must agree on it. One implementation instead of a private copy in `ingest.ts` plus a second reading of the same file elsewhere.

## Verification

Tests reproduce the issue's exact sequence — map A, modify, map B, re-check A — and I ran them against the previous implementation to confirm they catch the bug rather than merely describing it:

| | old | new |
|---|---|---|
| stays stale after an unrelated workspace is mapped | ✗ | ✓ |
| lastIngestAt comes from this workspace | ✗ | ✓ |
| root with no ingest record | ✗ | ✓ |
| file added since ingest | ✗ | ✓ |
| modified file is stale | ✓ | ✓ |

The fake client stubs `stats()` purely so it can run against the old code, which called it and discarded the result — otherwise every case would fail on a missing method rather than on the behaviour under test.

630 tests pass, `tsc --noEmit` clean, no new lint warnings.

### DIFF
--- a/ix-cli/src/cli/__tests__/stale-workspace-scope.test.ts
+++ b/ix-cli/src/cli/__tests__/stale-workspace-scope.test.ts
@@ -8,20 +8,28 @@ import { ingestMtimeCachePath } from "../config.js";
 import type { IxClient } from "../../client/api.js";
 
 let home: string;
-let realHome: string | undefined;
+const saved: Record<string, string | undefined> = {};
+
+// HOME *and* USERPROFILE: os.homedir() reads USERPROFILE on Windows and HOME on
+// POSIX, so overriding only one sends the Windows run at the real profile — where
+// ~/.ix may not exist and the seed write fails with ENOENT before any assertion
+// runs. Setting both keeps every platform inside the temp directory.
+const HOME_VARS = ["HOME", "USERPROFILE"] as const;
 
 beforeEach(() => {
-  // The cache path is derived from homedir() at call time, so pointing HOME at a
-  // temp directory keeps these tests off the developer's real ~/.ix.
-  realHome = process.env.HOME;
   home = mkdtempSync(join(tmpdir(), "ix-stale-home-"));
-  process.env.HOME = home;
+  for (const key of HOME_VARS) {
+    saved[key] = process.env[key];
+    process.env[key] = home;
+  }
   mkdirSync(join(home, ".ix"), { recursive: true });
 });
 
 afterEach(() => {
-  if (realHome === undefined) delete process.env.HOME;
-  else process.env.HOME = realHome;
+  for (const key of HOME_VARS) {
+    if (saved[key] === undefined) delete process.env[key];
+    else process.env[key] = saved[key];
+  }
   rmSync(home, { recursive: true, force: true });
 });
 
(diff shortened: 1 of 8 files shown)


KageBinary:
Closing in favour of #356 by @Hiro-Chiba, which is merged.

Theirs has the better core: an explicit `lastIngestAt` written into the baseline record, where this PR derived it from the cache file's own mtime — which any backup, rsync or stray `touch` invalidates. It also removes the client from `detectStaleFiles` entirely rather than keeping one `listPatches` call for the global rev, and its test injects a client whose `listPatches`/`stats` throw, which proves the cross-workspace read is gone instead of asserting around it.

The one thing here that #356 does not cover is worth keeping, and will follow as a separate PR: **`isFileStale` has the identical bug.** It still calls `listPatches({ limit: 1 })` per file, so `ix explain`, `ix locate`, `ix read` and `explain/facts.ts` all continue comparing against the newest patch anywhere on the machine. Making it synchronous and baseline-backed also drops an HTTP round-trip per file on those paths.

PR #362, opened 2026-08-09
fix(view): say when -p was not honoured, and show the URL in status

KageBinary:
Follow-up to #358 (thanks @Hiro-Chiba), picking up the two pieces I flagged when closing #360.

## `-p` is still silently ignored

#358 fixed the real bug — the already-running branch printed an unreachable URL built from the port just requested. It reports the running port now, but says nothing about the request:

```
$ ix view -p 19124 start
  http://localhost:19123
```

The URL is right. But you asked for 19124, got 19123, and nothing said so — you find out by noticing the number changed. Now:

```
  http://localhost:19123
[!] You asked for port 19124, but it is serving on 19123.
    Run 'ix view stop' then 'ix view -p 19124' to move it.
```

Keyed on whether `-p` was **typed** (`getOptionValueSource("port") === "cli"`), not on whether it differs from 8080. Comparing against the default would stay silent for someone who explicitly passed `-p 8080` to a server running elsewhere — exactly the case they typed the flag to find out about — and would nag anyone who typed nothing.

## `status` can answer its own question now

The port is recorded as of #358, so:

```
$ ix view status
[ok] Visualizer is running (PID 65764)
  http://localhost:19123
```

Silent rather than guessing for an instance old enough to have neither a port file nor a server script to recover one from — a confidently wrong URL is what #358 removed and this must not reintroduce it through the warning path.

## Tests

Extends #358's `view-running-port.test.ts` rather than adding a competing file: five more cases through their `runView` harness (mismatch warns, no-flag stays quiet, status shows the URL, status stays silent when unknown), plus three unit cases on the extracted `runningInstanceLines`.

Full suite green — 647 tests, `tsc --noEmit` clean, `eslint src` clean.

### DIFF
--- a/ix-cli/src/cli/__tests__/view-running-port.test.ts
+++ b/ix-cli/src/cli/__tests__/view-running-port.test.ts
@@ -180,11 +180,18 @@ describe("view running port state", () => {
     expect(await runView(["status"])).toContain("http://localhost:19123");
   });
 
-  it("omits the URL from status when the port is unknown", async () => {
+  it("says the port is unknown from status instead of printing nothing", async () => {
     seedRunningState();
     vi.spyOn(process, "kill").mockReturnValue(true);
 
-    expect(await runView(["status"])).not.toContain("http://localhost");
+    const output = await runView(["status"]);
+
+    // Still no guessed URL — that is the bug #358 removed. But silence left
+    // someone who ran status *for* the URL with nothing to act on, while
+    // `ix view` in the identical state told them how to recover.
+    expect(output).not.toContain("http://localhost");
+    expect(output).toContain("unknown");
+    expect(output).toContain("ix view stop");
   });
 });
 
--- a/ix-cli/src/cli/__tests__/view-running-port.test.ts
+++ b/ix-cli/src/cli/__tests__/view-running-port.test.ts
@@ -150,4 +150,65 @@ describe("view running port state", () => {
     expect(existsSync(scopeFile())).toBe(false);
     expect(existsSync(portFile())).toBe(false);
   });
+
+  it("says the requested port was not honoured", async () => {
+    seedRunningState(19123);
+    vi.spyOn(process, "kill").mockReturnValue(true);
+
+    const output = await runView(["-p", "19124", "start", "--no-open", "--all"]);
+
+    // Printing only the correct URL leaves the user believing -p moved the
+    // server. It did not, and the difference is one digit.
+    expect(output).toContain("You asked for port 19124");
+    expect(output).toContain("it is serving on 19123");
+  });
+
+  it("stays quiet when no port was asked for", async () => {
+    seedRunningState(19123);
+    vi.spyOn(process, "kill").mockReturnValue(true);
+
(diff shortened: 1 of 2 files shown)


KageBinary (commented):
Extracting `runningInstanceLines` as a pure function is the right move — inline in the command action this branch could only be exercised by launching a detached server, which is why the gap survived #358 in the first place.

`getOptionValueSource("port") === "cli"` is the correct predicate, and the comment explains why the obvious alternative is wrong in *both* directions: comparing against the default stays silent for someone who explicitly typed `-p 8080` against a server running elsewhere (exactly the case they typed the flag to discover), and nags anyone who typed nothing.

The `runningPort === null` test earns its place — asserting that no `http://localhost` appears at all is a direct regression guard on the confidently-wrong URL #358 removed. Easy to reintroduce that through a new warning path; now it fails loudly.

`status` reporting the URL is the small one that will get used the most.

All checks green. Approving.

josephismikhail (approved):

**Your `getOptionValueSource` approach checks out empirically, not just by reading.** Against the installed commander 15.0.0 source: `.default()` stamps `'default'` (`command.js:679`), the argv path stamps `'cli'` (`:713`), and `'env'` is only registered when `.env()` is used (`:719`) — which this isn't. 12 argv orderings probed, and every genuinely typed `-p` returns `cli`, including `--port=19124`, `-p19124`, `-p` after the subcommand, and the `-p 8080` case your comment calls out. `main.ts` does no argv rewriting and no config layer supplies a port, so the source is only ever `cli` or `default`. The warning cannot silently never fire.

**The wiring is genuinely tested**, which was my main worry given the extracted-pure-function shape. `runView` builds a real `new Command()`, calls the production `registerViewCommand`, and parses real argv — nothing stubs commander or the option source. Forcing the source comparison to fail breaks a test:
`× says the requested port was not honoured — expected '[ok] Visualizer is already running (P…' to contain 'You asked for port 19124'`
All five load-bearing lines mutation-checked; each kills exactly the test that claims it.

**I pushed one commit.** `ix view` and `ix view status` disagreed in precisely the state that motivated #358: with a live PID and no recoverable port, `ix view` printed the advisory and a recovery command, `status` printed the PID line and stopped. Not a #358 regression — neither guesses a URL, and `readRunningPort` funnels everything through `parsePort`, so a corrupt port file yields `null`, never a bad port. But silence is the wrong answer for someone who ran `status` *for* the URL, and it was unpinned: the old assertion was only `not.toContain("http://localhost")`, which the advisory satisfies too, so it could have drifted either way.

`status` now routes through `runningInstanceLines` so the two branches cannot drift. It takes no `-p`, so the mismatch branch is unreachable from there — `portWasRequested: false` means the `0` is never read, confirmed against every reachable `runningPort` value. The old assertion is retained, not replaced. Bug restored: `× says the port is unknown from status instead of printing nothing`, exactly one failure, so nothing else depended on the silence.

One cosmetic thing I left alone: when the port and scope advisories both fire, they print two recovery commands that each silently undo the other's intent (one drops `--all`, the other drops `-p`). Uncommon state, and worth a follow-up rather than widening this PR.

PR #363, opened 2026-08-09
refactor(stale): drop the client the staleness checks no longer use

### LINKED ISSUE #353: [bug] ix status uses another workspace timestamp and hides stale files

KageBinary:
Follow-up to #356 (thanks @Hiro-Chiba). **Not a bug fix — I said in review that `isFileStale` still had the cross-workspace bug, and I was wrong. #356 fixed both functions.** This is the cleanup that fix left behind.

`detectStaleFiles` and `isFileStale` both still take an `IxClient` and ignore it (`_client`), and both are still `async` despite no longer doing any I/O.

## Why bother

#356 fixed #353 by moving staleness onto the workspace-local ingest baseline — `listPatches({ limit: 1 })` carries no workspace, so it answered with whatever repo was mapped most recently anywhere on the machine.

Keeping an unused client parameter leaves the door open to reaching for it again. Removing it makes the fix **structural rather than a matter of discipline**: there is no client in scope, so the local-only property holds because it is the only thing the signature permits.

That's the same argument the PR's own test made by injecting a client whose `listPatches`/`stats` throw. That stub goes away here — a parameter that doesn't exist can't be called, which is a stronger guarantee than one that throws when it is.

## Also drops the fake async

Both read a local file now. `isFileStale` runs **per result** on `ix explain`, `ix locate` and `ix read`, so the `await` it used to need was a backend round-trip per file. Those call sites lose the `await` with it — `read.ts`'s `checkStale` wrapper included.

Six call sites updated: `status.ts`, `read.ts` (×3), `explain.ts`, `locate.ts`, `explain/facts.ts`.

## Verification

`tsc --noEmit` clean, `eslint src` clean, full suite green (640 tests). No behaviour change — the staleness logic itself is untouched.

### DIFF
--- a/ix-cli/src/cli/__tests__/stale-workspace.test.ts
+++ b/ix-cli/src/cli/__tests__/stale-workspace.test.ts
@@ -3,7 +3,6 @@ import * as fs from "node:fs";
 import * as os from "node:os";
 import * as path from "node:path";
 
-import type { IxClient } from "../../client/api.js";
 import { ingestMtimeCachePath } from "../config.js";
 import { loadIngestBaseline } from "../ingest-baseline.js";
 import { persistIngestBaselineIfClean } from "../commands/ingest.js";
@@ -12,15 +11,6 @@ import { detectStaleFiles } from "../stale.js";
 let home: string;
 let savedHome: string | undefined;
 let savedProfile: string | undefined;
-
-const noPatchReads = {
-  listPatches: async () => {
-    throw new Error("global patches must not be read");
-  },
-  stats: async () => {
-    throw new Error("global stats must not be read");
-  },
-} as unknown as IxClient;
 
 function writeSource(root: string, name: string, source: string): string {
   fs.mkdirSync(root, { recursive: true });
@@ -69,7 +59,7 @@ describe("workspace-scoped staleness", () => {
     ).toBe(true);
 
     moveMtimeForward(fileA);
-    const beforeB = await detectStaleFiles(noPatchReads, rootA);
+    const beforeB = detectStaleFiles(rootA);
     expect(beforeB).toEqual({
       lastIngestAt: ingestedA.toISOString(),
       currentRev: 11,
@@ -88,7 +78,7 @@ describe("workspace-scoped staleness", () => {
       ),
     ).toBe(true);
 
-    expect(await detectStaleFiles(noPatchReads, rootA)).toEqual(beforeB);
+    expect(detectStaleFiles(rootA)).toEqual(beforeB);
   });
 
   it("loads the legacy mtime-only cache and uses its own file timestamp", async () => {
@@ -108,7 +98,7 @@ describe("workspace-scoped staleness", () => {
     fs.utimesSync(cachePath, legacyTimestamp, legacyTimestamp);
     moveMtimeForward(filePath);
 
-    const result = await detectStaleFiles(noPatchReads, root);
+    const result = detectStaleFiles(root);
     expect(result.currentRev).toBe(0);
     expect(result.lastIngestAt).toBe(fs.statSync(cachePath).mtime.toISOString());
     expect(result.staleFiles).toBe(1);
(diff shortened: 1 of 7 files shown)


KageBinary (commented):
Agreed with the framing in the doc comment: after #353 moved staleness onto the workspace-local baseline, an unused `IxClient` parameter is not dead weight, it is an invitation to reach for the backend again. Removing it makes the local-only property hold by construction instead of by discipline — the signature no longer permits the bug.

Dropping `async` is the part that actually shows up for users: `isFileStale` is called per result in `explain`, `locate` and `read`, so this used to be an `await` per file on a path that now only stats the local baseline.

Checked that the call sites are complete — `explain.ts`, `locate.ts`, `read.ts`, `status.ts`, `explain/facts.ts` — and Lint & Typecheck passing confirms nothing still passes a client.

Nit, take it or leave it: the `await` removal left redundant parens in four assertions, e.g. `expect((detectStaleFiles(root)).staleFiles)`.

All checks green. Approving.

josephismikhail (approved):
Independent review pass (separate agent, own worktree) — no blockers.

**Verified, not assumed:**
- All **13** call sites on `origin/main` were updated; re-grepped the whole repo including `docs/`, `scripts/`, `core-ingestion/`, `.github/`, and the `@ix/pro` symlink under `node_modules`. Zero stragglers, zero string/reflective references. (Control: the same sweep finds `IxClient` in 604 files, so it wasn't a silently-empty search.)
- A stale caller **cannot** compile: probed `detectStaleFiles(client, "/tmp")` and `isFileStale(client, p)` both plain and `as any` — TS2345/TS2554 on all four. The "client silently lands in `filePath`" scenario is unrepresentable, because `maxSamples` is a `number`.
- async→sync is safe: no `.then`/`.catch`/`Promise.all` on either symbol at base or head; every base call site already had its `await` **inside** the `try`, so the same `catch` that swallowed rejections now swallows throws. No un-awaited truthiness site existed, so there's no always-true-promise inversion.
- The `_client` really was inert, and nothing reaches the backend transitively (`loadIngestBaseline` is fs-only; `resolveWorkspaceRoot` is fs + `git rev-parse`).
- Confirmed this is the **vestige removal, not the fix** — `ebeee6c` (#356) is what fixed #353, and it renamed `client`→`_client` in the same hunk.

**On dropping `noPatchReads`:** agreed it's an acceptable trade. That stub had already gone tautological after #356 — it asserted a negative about a parameter nothing read. The test that actually encodes #353 (ingest A, ingest B newer, assert A unchanged) survives and needs no fake client. I checked it isn't vacuous: made `detectStaleFiles` ignore its `root` argument and the file went **5 failed / 1 passed**.

**Gates:** `tsc --noEmit` clean · `eslint src` 0 errors (38 warnings, all pre-existing) · `knip` exit 0 · full suite **638 passed, 2 skipped**.

Two things worth a follow-up, neither caused by this PR and neither blocking:
- `read.ts:185` resolves `--root` explicitly but `isFileStale` → `resolveWorkspaceRoot()` with no argument, so `ix read --root /other/repo` compares against the *cwd's* baseline. Identical on main.
- `explain.ts:117`, `locate.ts:98`, `facts.ts:182` pass a raw repo-relative `source_uri` into an `existsSync`, so they silently report "not stale" when cwd isn't the workspace root. `read.ts` is the only caller that converts to absolute first.

One nit on the new comment at `stale.ts:81`: "without one there is nothing to reach for" isn't literally true — `stale.ts` already imports from `./config.js`, which exports `createClient()`. The argument holds for a better reason worth stating instead: `createClient()` is async, so any use forces both functions back to `async` and breaks all 13 call sites at compile time.

PR #368, opened 2026-08-10
feat(skill): ship the ix agent skill and the HTTP API reference

Alot1z:
## What

Ships the self-contained **ix agent skill** that teaches an LLM agent to drive the `ix` CLI — build the graph, explain symbols, trace flows, measure impact — instead of grepping and guessing. Plus a backend HTTP API reference generated from `IxClient` and checked against the live backend.

Originally authored by @Alot1z; scoped down and fixed in `0c9087c` (see the commit message and [this comment](https://github.com/ix-infrastructure/Ix/pull/368#issuecomment-5235488710)).

## Contents

- **`skills/ix/`** — `SKILL.md`, `references/` (command routing, output formats, troubleshooting), `scripts/` (bootstrap for bash + PowerShell)
- **`docs/api/`** — HTTP API reference (`README.md`) + OpenAPI 3.0 spec (`openapi.yaml`, 39 paths / 13 schemas)
- **`scripts/install-skill.sh`** — deploy to `~/.claude/skills/ix` and `~/.agents/skills/ix`
- **README** — skill install section; **CLAUDE.md** — project context and verified gotchas; **.gitignore** — `/dist/`, session artifacts

## Not in scope any more

The Compass fit-view patch and its React port were removed in `0c9087c` — they patch compiled artifacts of this repo and of Compass, and Compass has its own repo. They belong in a `system-compass` PR, where the F key is genuinely new. The branch history preserves them.

## Verification

- Skill installs and re-installs cleanly; refuses to clobber a same-named skill it did not write; `--force` overrides. All four paths tested.
- `bootstrap.sh` dry-run green with install/backend/map/compass skipped.
- `bash -n` on both shell scripts; OpenAPI spec parses; every path base segment exists in the backend routes.
- `ix-cli` source untouched — no build or test impact.

## Findings kept from the original PR

- **`ix reset` is global** — takes no workspace_id and wipes every workspace in the shared backend.
- **`registerPatchesCommand` is defined but never registered** — now tracked as #371.


KageBinary (changes requested):
Thanks for this — and welcome. This is a lot of careful work, and unusually, everything you claimed in the PR description checks out when I ran it myself rather than just reading it. That is rarer than it should be, so I want to be specific about what I verified before getting to the changes I'm asking for.

## What I verified as working

- Your stated checks all pass: `node --check fit-view.js`, `bash -n` on all three shell scripts, and the OpenAPI spec parses cleanly (39 paths, 13 schemas). Every path's base segment exists in the backend's Scala routes, so the API reference is grounded, not invented.
- `apply.sh` is genuinely idempotent. I ran it three times against a fixture; the third pass printed "already patched" for all three steps and the `<script>` tag count in `index.html` stayed at 1.
- It works against the **real shipped Compass**. I copied `~/.ix/cli/compass` (v0.2.0) to a scratch dir and ran the patch: the minified marker `` {keys:[`0`],label:`Reset zoom & center`} `` really is present in `KeyboardHelp-CLjfSVTU.js`, the F entry inserted correctly, and `node --check` confirms the patched chunk still parses. Same for the injected `view.js`.
- `registerPatchesCommand is defined but never registered` — **correct**, and a real find. `patches.ts:6` exports it and nothing imports it, so `ix patches` is an unknown command today. That deserves its own issue; I'll open one.

So the mechanism works. My concerns are about layer and packaging, not craft.

## Blocking

**1. It patches the same two inline styles React owns.** `SystemMapCanvas.tsx:858-871` in the Compass source renders:

```jsx
<div style={{ transform: `translate(${snappedPan.x}px, ${snappedPan.y}px)`, transformOrigin: "0 0" }}>
  <div className="relative ix-crisp-canvas" style={{ zoom: snappedZoom, ... }}>
```

`fit-view.js` writes `wrap.style.transform` and `canvas.style.zoom` from outside React — the exact two properties driven by `snappedPan` / `snappedZoom` state. Every subsequent render restores React's values. The 60 ms "verify the target stuck and re-apply" timer is a band-aid on that race rather than a fix. Related: the background grid at `:855` uses `backgroundPosition: snappedPan`, which the patch doesn't update, so after pressing F the grid stays put while the nodes move.

Worth checking too whether the auto-frame is still needed: Compass #57 (`fix/bounded-layout-and-viewport-refit`, merged and bundled in v0.9.1) fixed the empty-first-paint case directly — the comment at `SystemMapCanvas.tsx:450` describes the same symptom you're solving.

**2. Two of three patch steps silently no-op on stock Linux.** This box has `python3` and no `python` (standard on Debian/Ubuntu/Fedora). Running `apply.sh`:

```
index.html: added fit-view.js script tag
apply.sh: line 40: python: command not found
keyboard help: skipped (python unavailable)
apply.sh: line 68: python: command not found
view.js server: skipped (python unavailable)
Compass fit-view patch applied. Restart the visualizer ... picks up the real /__ix/remap endpoint
--- exit code: 0 ---
```

It exits 0 and tells you to restart for an endpoint it never installed. `bootstrap.sh:137` compounds it by redirecting to `/dev/null` and reporting success. Swapping in a `python` → `python3` shim made all three steps apply perfectly, so this is purely the interpreter name plus the `|| { echo "skipped"; }` swallowing a real failure.

**3. `sed -i` with no suffix is GNU-only** (`apply.sh:33`) and fails on macOS, which the README section and `SKILL.md` both claim to support. The repo already has both portable forms: `scripts/disconnect.sh` uses `sed -i.bak`, `scripts/release.sh` uses `sed -i ''`.

**4. The `/__ix/remap` endpoint should be source, not a patched artifact.** This repo owns `ix-cli/src/cli/commands/view.ts`; step 4 string-injects the handler into the *compiled* `~/.ix/cli/cli/dist/cli/commands/view.js`, which every `ix upgrade` overwrites. In `serverScript()` it would be typechecked, tested, and permanent. Two things to fix on the way in: `serverScript` ends in `server.listen(PORT)` with no host, so it binds `0.0.0.0` — the handler would be reachable from the whole LAN, and from any web page in the user's browser, since a simple-content-type POST isn't preflighted. It should bind `127.0.0.1` and check `Sec-Fetch-Site` / `Origin` before spawning a 30-minute `ix map`.

**5. `CLAUDE.md` ships an absolute path from your machine:** `python "C:/Users/jacob/.agents/skills/skill-creator/scripts/package_skill.py"`. This is the file every agent working in the repo loads. The README's copy of the same command already uses `$HOME`, so the two disagree.

**6. `CLAUDE.md` documents two paths that can't exist** — `dist/ix.zip` and `.freebuff/run.md` are both gitignored by this same PR, but the layout table presents them as repo contents.

## Smaller

- `.gitignore`: `dist/` is unanchored so it matches at any depth, and `core-ingestion/dist/` + `ix-cli/dist/` were already covered. `/dist/` is what you want. (Nothing tracked is affected today — I checked.)
- `upstream/compass-fit-view/` is 723 lines of React that nothing here builds, typechecks, or tests. It belongs in the Compass repo as a PR.
- `install-skill.sh` `rm -rf`s `~/.claude/skills/ix` and `~/.agents/skills/ix` with no prompt, which silently destroys local edits to a same-named skill.

## Suggested split

This is five unrelated changes in one commit, and they're ready at very different times:

1. **`docs/api/`** — closest to mergeable as-is. I'd take this on its own today.
2. **`skills/ix/` + `install-skill.sh` + README** — mergeable once (2), (3) and (5) are fixed.
3. **`/__ix/remap`** — a small PR against `view.ts` with the bind + origin fix.
4. **fit-view** — to the Compass repo, after checking what #57 already covers.
5. The `registerPatchesCommand` and global-`ix reset` findings — issues, and thank you for both.

Happy to review each of those as they come. If you'd like to start with just #1, I think that lands quickly.

KageBinary:
Pushed the fixes to this branch rather than sending you round another loop — everything below is on `0c9087c`, and the original commit is untouched underneath it so nothing you wrote is lost.

**Scoped it to the two pieces that belong in this repo:** the `ix` skill and the API reference. The API docs I would have taken as-is; they hold up (39 paths, 13 schemas, every path grounded in the backend routes).

**Moved out**, because the layer was wrong rather than the code:

- `skills/ix/scripts/compass-patch/` — the reasons are in the commit message, but the short version is that this repo owns `view.ts`, so patching its *compiled* output is fighting the build, and `fit-view.js` was writing the exact two inline styles React drives from `snappedZoom`/`snappedPan`.
- `upstream/compass-fit-view/` — nothing here builds React.

**Both are worth landing in `system-compass`, and I would review that PR.** The F-key fit view is genuinely new there — Compass binds `0` for reset-zoom and nothing binds `F`. Your `upstream/` port is most of the work already. What I would drop is the auto-frame-on-load: Compass #57 fixed the empty-first-paint upstream and it ships in v0.9.1, so that part is solved. The F key is not.

**Two bugs found while testing your scripts, both fixed here:**

- `version_ge` in `bootstrap.sh` was inverted. `sort -C` succeeds when input is *already ordered*, so want-then-got asks "is want <= got" — the `-r` asked the reverse. Node 22.22.2 against a floor of 22 printed "Ix requires Node >= 22", and Node 18 passed silently. I hit it dry-running your bootstrap on this machine. Checked across 22.22.2 / 24 / 22.0.0 / 20.11.1 / 18.4.0.
- `install-skill.sh` `rm -rf`d the destination unconditionally. `ix` is a short name, and a hand-written skill that happened to share it went without a prompt or a backup. It now refuses unless the destination carries `name: ix`, with `--force` to override. Tested all four paths: clean install, re-install, squatted name, and `--force`.

Also: `C:/Users/jacob/...` out of CLAUDE.md, the two gitignored paths out of its layout table, `/dist/` anchored, and the `POST /__ix/remap` section in the API docs now says what actually happens (falls through to the SPA handler, returns index.html with a 200) rather than describing the patched behaviour.

**Kept, with credit:** the global-`ix reset` warning and `registerPatchesCommand` never being registered. The second is now #371 — a good catch, it had been sitting there unnoticed because `patches` is in `PRO_COMMANDS`, so OSS users get "requires Ix Pro" instead of an unknown-command error.

Re-reviewing now. Thanks for a genuinely careful first contribution — the fact that everything in your description checked out when I ran it is not the norm.

KageBinary (approved):
Re-reviewed after `0c9087c`. Every blocking item is resolved — the two that were architectural by removal, the rest by fix — and the full matrix is green now that the held workflow runs are approved (`CI Passed`, E2E, Linux/macOS/Windows on Node 22 and 24).

What lands: the skill, the API reference, and two real findings about this codebase. That is a good first contribution.

@Alot1z — the fit-view work is not rejected, just addressed to the wrong repo. If you open it against `system-compass` I will review it there. Your `upstream/compass-fit-view/` port is most of the way; drop the auto-frame (Compass #57 covers it) and keep the F key, which nothing binds today.

Approving. Merging per the queue.

PR #372, opened 2026-08-10
feat(llm): implement --format llm for the five commands that faked it

KageBinary:
## What

`explain`, `read`, `status`, `doctor` and `savings` advertised `text|json|llm` and rendered `text` when given `llm`. `docs/llm-format.md` documented that fallback as intentional — commands whose output is prose or verbatim source route to "the most compact existing form".

That holds for the *payload* and not for what surrounds it. `explain`'s prose is a rendering of `facts` + `role` + `importance`, so an agent can have the inputs and skip the paragraph. `read`'s source was arriving under a 4-column line-number gutter and two ANSI escapes per line. And `explain` is the first call most plugins make on any question, so the fallback was costing the most tokens on the hottest path.

## Measured

Against `--format json`, which the help text sent agents to as the workaround:

| command | json | llm | saved |
|---|---:|---:|---:|
| explain | 1849 | 829 | **55%** |
| doctor | 682 | 470 | 31% |
| savings --detail | 1010 | 693 | 31% |
| read (40 lines) | 1661 | 1585 | 5% |
| status | 184 | 186 | −1% |

`read` gains 23% against `text` instead — 601 of those bytes are gutter and ANSI on a 40-line read.

**`status` is a wash on size and is in anyway**, for the explicit `stale=true|false` field: that is the question the command gets called to answer, and `text` only implied it through a warning line. Its unreachable-backend path now emits the spec's `error code=...` record so a consumer parses the failure the same way it parses a result.

## Notes

- **`read`'s content block is the one place the one-record-per-line invariant is relaxed.** The preceding `content lines=<n>` record makes the block self-delimiting, so a line of source can never be mistaken for a record.
- Still on text, now documented rather than silent: `diff --content` (verbatim hunks) and `ingest` (hidden implementation-detail command).
- Found while writing this: the `doctor` status ternary. A warning there is `{ok: false, warn: true}`, so checking `warn` on the `ok` branch reports every warning as a hard failure — the opposite of what `hasFailure` concludes two lines down. Caught by the test, fixed before it shipped.

## CLAUDE.md

The paragraph on this was wrong in both directions: it claimed the deprecated `query` advertises `llm` (it offers only `text|json`) and missed `doctor`, `savings` and `ingest`. It also never mentioned that `config`, `init`, `reset`, `upgrade`, `view`, `watch` and `docker` take no `--format` at all — `--format llm` there is an `unknown option` error, not a fallback, which is the one case where an agent following the doc's "pass it unconditionally" advice breaks.

That paragraph sits between the `IX-MEMORY` markers, which say not to hand-edit. **There is no generator for it in this repo** (only `scripts/disconnect.sh` touches the markers, and only to strip them), so I corrected it in place — but whatever regenerates it needs the same fix or this reverts on the next run.

## Test

51 files / 661 tests green, plus 13 new in `llm-tier5.test.ts` covering the records, zero-count survival, quoting of a multi-line signature, and the `--detail` gate.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -22,10 +22,8 @@ llm` to it silently renders text.
 
 `--format llm` is **not** accepted at all by `config`, `init`, `reset`,
 `upgrade`, `view`, `watch` and `docker`, which take no `--format`; passing it is
-an `unknown option` error, not a fallback. Do not pass it blind.
-
-These commands take no `--format` at all: `config`, `init`, `reset`, `upgrade`,
-`view`, `watch`. (`ingest` does accept it, despite being an action command.)
+an `unknown option` error, not a fallback. Do not pass it blind. (`ingest` does
+accept `--format`, despite being an action command.)
 
 **Pro features.** Some commands below are marked **[Pro]**, as are the whole
 **Planning** and **Workflows** sections — every command in those two tables is
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -15,10 +15,14 @@ Use the `ix` CLI exclusively.
 - `--format json` — use when chaining results between commands, or when you need
   to pull a specific field out of the response.
 
-**Four commands advertise `llm` in their help but do not implement it**, and fall
-back to human-oriented text without an error: `explain`, `read`, `status`, and the
-deprecated `query`. Use `--format json` on those if you need to parse the result.
-Everywhere else in the routing tables, `llm` behaves as described.
+Every command that accepts `--format` implements `llm`, with two exceptions that
+fall back to text without an error: `diff --content` (verbatim hunks) and
+`ingest`. The deprecated `query` accepts only `text|json` — passing `--format
+llm` to it silently renders text.
+
+`--format llm` is **not** accepted at all by `config`, `init`, `reset`,
+`upgrade`, `view`, `watch` and `docker`, which take no `--format`; passing it is
+an `unknown option` error, not a fallback. Do not pass it blind.
 
 These commands take no `--format` at all: `config`, `init`, `reset`, `upgrade`,
 `view`, `watch`. (`ingest` does accept it, despite being an action command.)
--- a/docs/llm-format.md
+++ b/docs/llm-format.md
@@ -95,8 +95,31 @@ Renderers shipped: Tier 1 (`map`, `subsystems`, `impact`, `smells`,
 `overview`) plus `stats`; Tier 2 (`inventory`, `rank`, `depends`, `trace`,
 `contains`, `callers`, `callees`, `imports`, `imported-by`); Tier 3 (`search`,
 `text`, `history`, `patches`); Tier 4 (`entity`, `locate`, `diff`,
-`conflicts`). Commands whose output is verbatim source or prose (`read`,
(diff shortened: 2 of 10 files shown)


josephismikhail (approved):

**The renderers themselves are solid.** Record shape matches the shipped Tier 1–4 convention (header record named after the command, then repeated child records), keys are snake_case throughout with no camelCase leaking, and every field except `read`'s content block goes through `llmLine` → `llmField` → `llmQuote`. Probed the free-text fields directly with embedded newlines, quotes, tabs and backslashes — doctor check names and details, status paths with spaces and quotes, raw-explain docstrings — all correctly escaped, nothing spans two lines. Every docs claim was verified by running the built CLI rather than reading: all seven of `config`/`init`/`reset`/`upgrade`/`view`/`watch`/`docker` do exit 1 with `unknown option`, and `query` really does offer only `text|json`. Three of the strongest test claims mutation-checked and each bites.

**One blocker found and fixed, in a commit I pushed.** `printLlmLines` skipped falsy entries, and a blank source line is `""`. So `read --format llm` counted lines in `content lines=<n>` that never reached stdout, and a consumer honouring the count — the entire point of the self-delimiting block — read the following records as source. Wrong on almost every real input, not an edge case: `ix read src/cli/llm.ts --format llm` declared **91** and emitted **81**. Correct only for a file with no blank lines and no trailing newline.

Changed to `!= null`. Safe because `printLlmLines` had zero callers on `main` before this branch, and I enumerated every path into it: all seven call sites go through renderers that build lines exclusively from `llmLine("<literal>", …)`. The only three `llmLine` calls with a variable kind (`impact.ts:548`, `trace.ts:331`, `format.ts:490`) don't use `printLlmLines`, and the only `push("")` calls are text renderers behind `subsystems`.

Worth naming the trap, because it's the recurring one here: **your test was correct and still passed against the bug.** `expect(lines.slice(2)).toEqual([…, ""])` asserts on `renderReadLlm`'s return value, which was always right — the defect lived in render *composed with* print. So `outputResult` is now exported and the new tests drive the real print path across blank lines, an empty file, CRLF, no-trailing-newline and blank-lines-only.

Also added the `{ok:true, warn:true}` row for `doctor` — the one shape your ternary's ordering actually decides, and previously uncovered. It pins `status=ok`, matching what the glyph line renders for the same record; `warn` would have been wrong. And folded the CLAUDE.md paragraph that still said "these commands take no `--format`" into your new one, since yours is a strict superset (it adds `docker`) and the two contradicted each other two paragraphs apart.

Bug restored on both fixes: 6 targeted failures — 5 framing, 1 doctor. `"no trailing newline"` correctly does *not* fail, since it's the one input the old code got right. tsc/eslint/knip clean, 667 passing.

Not fixed, flagged for follow-up: **the `--format llm` dispatch itself is still untested for all five commands.** Reverting the `"llm"` comparison in each leaves the full suite green. The renderers are covered; the branch that reaches them is not, which is exactly the regression this PR exists to fix. Closing that needs client mocking per command, and I'd rather leave it to you than bolt on a structural test that gives false confidence. Also unfixed: `read`'s not-found and `explain`'s resolve-failure branches still emit human prose to stderr and nothing on stdout, so `ix read bogus --format llm` gives an agent zero bytes and exit 0 — `diff.ts:471` already has the `llmError` precedent.

PR #373, opened 2026-08-10
ci(release): give the auto-generated brew PR a conventional title

KageBinary:
`pr-title.yml` runs `action-semantic-pull-request` against a fixed type list — `feat fix docs refactor test ci build perf chore revert`. The release workflow opens its formula PR as `brew: update formula for vX`, and `brew` is not on that list, so **the Conventional title check has failed on arrival for every release**.

It is not one of the ruleset's required checks (the ruleset requires only `CI Passed`), so the PR stays technically mergeable and the red X reads as background noise instead of something with a fixable cause. #367 is the current instance — I retitled it by hand to `chore(brew): …` to clear it; this stops the next release from reproducing it.

`scripts/release.sh` prints the same string as a copy-paste instruction for the manual path, so it drifts the same way. Fixed alongside.

Verified: `bash -n scripts/release.sh` and a YAML parse of `release.yml`.

### Separate, not fixed here

#367's workflow runs also arrived in `action_required` and sat there — held pending manual approval, which is why `CI Passed` never reported and the merge was blocked. I approved the six runs by hand and they went green. That will recur on every release too: the PR is created with `GITHUB_TOKEN` as `github-actions[bot]`, which is what puts its runs behind the approval gate. Fixing it properly means opening that PR as a real actor (app token or PAT), which needs a secret and is your call to make.


josephismikhail (approved):

**Your change is correct and complete.** `chore` is in `pr-title.yml`'s type list, and there is no `scopes`/`requireScope`/`subjectPattern` config that could reject the title for a different reason — verified against the action source at the pinned SHA, where all four of those checks are strictly opt-in. `grep` for `brew:` repo-wide returns only the comment itself, the tap is this repo (no `--repo` on `gh pr create`, `Formula/ix.rb` lives here), and nothing parses commit subjects — no commitlint, no `.github/release.yml`, and the only consumer (`github/transform.ts:141`) takes the subject as a display string. `bash -n` clean, YAML re-parses, one argv element with the parens intact.

**I pushed one commit** correcting the comment, because it lands as permanent in-repo documentation and three claims in it did not hold:

- *"every release"* — `pr-title.yml` arrived in #282 on 2026-06-08. Nine of the eleven formula PRs (#38→#268, latest 2026-06-05) predate it. Only #364 and #367 were ever in scope.
- *"failed on arrival"* — it does not run on arrival. The PR is opened by `github-actions[bot]`, so its `pull_request` runs land in `action_required`. On #364 the title check **never ran at all** — there was no red X. Direct evidence on #367: runs `created_at 23:16:07`, `run_started_at 03:05:05` — queued 3h49m awaiting approval.
- *"written off as noise each time"* — n=1, and it wasn't written off; you renamed it and it went green 94 seconds later. That manual rename is the toil this removes, and it's the stronger argument.

The comment now also records the approval hold, since that is what actually blocks the formula PR and this doesn't fix it.

Two things I got slightly wrong in my own commit message, for the record: the rename was 94 seconds after the red X (3h50m after opening), not "within six minutes" — six minutes is the rename→merge gap. And the ruleset carries `required_approving_review_count: 1` as well as the workflow-approval hold, so there are two holds, not one. Neither affects the landed comment.

PR #375, opened 2026-08-10
fix(ingest): resolve JS and TS calls across parse batches

### LINKED ISSUE #374: [bug] JS and TS calls disappear across 500-file parse batches
## Description

`ix map` resolves JavaScript and TypeScript calls differently depending on which 500-file parse batch contains the caller and callee. A caller parsed before its imported definition loses its `CALLS` edge, while an equivalent caller in the definition's batch resolves correctly.

This makes `ix callers`, `ix callees`, and `ix impact` undercount production usage and can produce a low-risk assessment for a widely used symbol.

## Steps to reproduce

1. Create a fresh TypeScript repository with 515 tracked files.
2. Put `src/a-consumer.ts` near the start of `git ls-files`:

```ts
import { crossBatchTarget } from "src/z-target";

export function runCrossBatchTarget(): number {
  return crossBatchTarget();
}
```

3. Add 510 filler TypeScript files so that `src/z-target.ts` is outside the first 500-file batch:

```ts
export function crossBatchTarget(): number {
  return 42;
}
```

4. Add `src/zz-consumer.ts` after the target with the same import and call pattern.
5. Configure `baseUrl` as `.`, then run:

```sh
ix map . --silent
ix impact crossBatchTarget --kind function --depth 2 --format json
ix callees runCrossBatchTarget --kind function --format json
```

## Expected behavior

Both `runCrossBatchTarget` and the same-batch consumer are callers of `crossBatchTarget`. The early consumer's `callees` result contains `crossBatchTarget`.

## Actual behavior

Only the consumer in the target's batch is reported. The earlier consumer has no resolved callee.

The path-only global index identifies `src/z-target.ts`, but JavaScript and TypeScript definitions are not included in `fileQKeys` or `fileHasSymbol`. `resolveEdges` therefore cannot verify the imported destination and drops the cross-batch `CALLS` edge.

## CLI output

```json
{
  "summary": {
    "callers": 1,
    "callees": 0
  },
  "callerList": [
    {
      "kind": "function",
      "name": "runSameBatchTarget"
    }
  ]
}
```

`ix callees runCrossBatchTarget` returns zero results.

## Environment

- OS: macOS 26.6.1
- Node: v26.7.0
- Ix version: 0.9.1

Hiro-Chiba:
Fixes #374.

## Summary

Pre-scan JavaScript and TypeScript definitions into the repository-wide resolution index so streamed parse batches can resolve calls to imported symbols defined in later batches.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Include JavaScript and TypeScript source files in the existing resolution pre-scan.
- Build their symbol index from parser-produced entities so it stays aligned with normal ingestion.
- Add a core regression test where the caller is parsed without the imported definition in its batch.
- Pin the CLI pre-scan extension set with coverage for all supported JavaScript and TypeScript extensions.

## Validation

- `cd ix-cli && npm test` (50 files, 649 tests, parser smoke passed)
- `npm run typecheck`
- `npm run build`
- `core-ingestion/src/__tests__/resolveEdges.test.ts` (33 tests passed)
- Anonymous 515-file end-to-end map: both the early-batch and same-batch callers resolve after the change; only the same-batch caller resolved before it
- The complete `core-ingestion` suite has six pre-existing failures identical on `origin/main`: two fixtures require the private `memory-layer` tree, one snapshot is stale, and three existing Rust/Scala expectations differ from current parser output

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/core-ingestion/src/__tests__/resolveEdges.test.ts
+++ b/core-ingestion/src/__tests__/resolveEdges.test.ts
@@ -256,6 +256,78 @@ describe('resolveEdges', () => {
     });
   });
 
+  it('resolves a TypeScript call when the imported definition is outside the parse batch', () => {
+    const caller = fileResult(
+      '/repo/consumer.ts',
+      SupportedLanguages.TypeScript,
+      [entity('runCrossBatchTarget', SupportedLanguages.TypeScript)],
+      [
+        { srcName: 'consumer.ts', dstName: 'target', predicate: 'IMPORTS', importRaw: './target' },
+        { srcName: 'runCrossBatchTarget', dstName: 'crossBatchTarget', predicate: 'CALLS' },
+      ],
+    );
+    caller.importBindings = [
+      { pkg: './target', local: 'crossBatchTarget', imported: 'crossBatchTarget' },
+    ];
+
+    const targetPath = '/repo/target.ts';
+    const globalIndex = buildGlobalResolutionIndex(
+      ['/repo/consumer.ts', targetPath],
+      new Map([[targetPath, 'export function crossBatchTarget() {}\n']]),
+    );
+
+    expect(resolveEdges([caller], undefined, globalIndex)).toContainEqual({
+      srcFilePath: '/repo/consumer.ts',
+      srcName: 'runCrossBatchTarget',
+      dstFilePath: targetPath,
+      dstName: 'crossBatchTarget',
+      dstQualifiedKey: 'crossBatchTarget',
+      predicate: 'CALLS',
+      confidence: 0.9,
+    });
+  });
+
+  it('uses caller-supplied parse results instead of re-parsing', () => {
+    // The CLI parses prescan sources on its worker pool and hands the results
+    // in, so the index costs no main-thread parsing. Proven by supplying a
+    // parse result whose entity does not appear in the source text: if the
+    // index re-parsed, `fromPool` could not be resolved.
+    const targetPath = '/repo/target.ts';
+    const caller = fileResult(
+      '/repo/consumer.ts',
+      SupportedLanguages.TypeScript,
+      [entity('run', SupportedLanguages.TypeScript)],
+      [
+        { srcName: 'consumer.ts', dstName: 'target', predicate: 'IMPORTS', importRaw: './target' },
+        { srcName: 'run', dstName: 'fromPool', predicate: 'CALLS' },
+      ],
+    );
+    caller.importBindings = [{ pkg: './target', local: 'fromPool', imported: 'fromPool' }];
+
+    const poolResult = fileResult(
+      targetPath,
+      SupportedLanguages.TypeScript,
+      [entity('fromPool', SupportedLanguages.TypeScript)],
+      [],
+    );
+
+    const globalIndex = buildGlobalResolutionIndex(
+      ['/repo/consumer.ts', targetPath],
+      new Map([[targetPath, '// the pool result is authoritative, not this text\n']]),
+      new Map([[targetPath, poolResult]]),
+    );
+
+    expect(resolveEdges([caller], undefined, globalIndex)).toContainEqual({
+      srcFilePath: '/repo/consumer.ts',
+      srcName: 'run',
+      dstFilePath: targetPath,
+      dstName: 'fromPool',
+      dstQualifiedKey: 'fromPool',
+      predicate: 'CALLS',
+      confidence: 0.9,
+    });
+  });
+
   it('resolves Elixir alias-qualified calls through the implicit short alias', () => {
   const caller = fileResult(
     '/repo/lib/my_app/accounts.ex',
--- a/core-ingestion/src/index.ts
+++ b/core-ingestion/src/index.ts
@@ -2641,6 +2641,14 @@ export interface GlobalResolutionIndex {
 export function buildGlobalResolutionIndex(
   filePaths: string[],
   sources?: Map<string, string>,
+  // Parse results the caller already has. The parser-derived language blocks
+  // below (PHP, JS/TS, R, SAS) call parseFile synchronously on the main thread,
+  // which is fine for a handful of files and not fine for a whole repository's
+  // worth of TypeScript. The CLI owns a worker pool; letting it parse in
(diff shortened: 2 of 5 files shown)


KageBinary (changes requested):
Thanks for contributing! The diagnosis here is correct and the index construction is careful — my one concern is the cost, which I think is larger than the 515-file validation repo could show.

## The part that's right

The JS/TS block mirrors the in-batch construction in `resolveEdges` exactly — same `kind === 'file' || kind === 'module'` filter, same `qualifiedKey(e)`, same name-keyed list of qkeys. That is the thing that usually drifts in this file (it is why the R and SAS blocks carry those "mirrors the qkMap construction at resolveEdges" comments), and you got it right, so cross-batch and in-batch resolution will agree on `dstQualifiedKey`. No correctness objection.

## The concern: this is the first prescan language where the prescan is the dominant cost

`buildGlobalResolutionIndex` calls `parseFile`, which is **synchronous tree-sitter on the main thread**. The streaming parse loop below it uses `ensureParsePool()` — a worker pool. So every JS/TS file now gets parsed twice: once serially in the prescan, once in parallel in the pool.

That trade was fine for the existing prescan set. Go uses a regex scan specifically to stay cheap, and R/SAS are niche and usually a small slice of a repo. JS/TS is the common case, and for a TypeScript repo it is ~all of the files.

Measured on this repo's own TS (176 files, 1.5 MB, `parseFile` via `core-ingestion/dist`):

```
SYNC prescan parse of all JS/TS: 1.02 s for 176 files  (~5.8 ms/file)
```

Extrapolating: ~5.8 s per 1,000 JS/TS files, ~29 s at 5,000 — single-threaded, before the parse pool starts.

**The incremental path is where this really bites.** In Path A:

```js
const remainingIndexFiles = filePaths.filter(fp => needsIndexPrescan(fp) && !changedSet.has(fp));
```

`filePaths` is the whole repo, so every *mtime-clean, unchanged* JS/TS file is read from disk and then synchronously parsed. Before this change `.ts` was not in `INDEX_PRESCAN_EXTENSIONS`, so `sources` and `remainingIndexFiles` were both empty and the cost was ~zero. After it, a one-file edit in a 5,000-file TS repo pays the full read + full serial parse on every `ix map`. That is the fast path this cache exists to protect.

At 515 files this is ~3 s and reads as acceptable, which I suspect is why it did not surface in your validation.

## Options, roughly in order of preference

1. **Run the prescan through the existing worker pool.** `ensureParsePool()` already exists in `ingest.ts` and is already parallel. It would need `buildGlobalResolutionIndex` to accept pre-parsed entities (or an async variant) rather than calling `parseFile` itself, since the pool lives in the CLI and this function is in `core-ingestion` — a bit of plumbing, but it removes both the double-parse and the serialization.
2. **Reuse the streaming loop's parse output.** The pool parses these files moments later anyway; only the ordering (index-before-batch) forces the duplicate.
3. **Narrow what gets prescanned.** Only files that are actually import targets of something in the changed set need to be in the index, which on an incremental map is a small fraction of the repo.
4. **Regex scan like Go.** Cheapest, but you deliberately avoided it for the drift reason above, and I think you were right to.

I do not think the fix has to be in this PR — but I would not want to land the JS/TS prescan without a plan for the incremental case, since that is the most common thing users do.

Happy to look again once you have picked a direction.

KageBinary:
Rebased onto `main` — #380 and #384 landed since you opened this, so it was behind.

Authorship is preserved on your commit; only the base moved. Nothing else in the PR was touched, and the review comments above still stand as written.

KageBinary:
Rebased onto `main` and pushed the fix for the blocking item, so this is unblocked from my side.

**Note the conflict**: #382 landed while this was open and touched the same three places — `INDEX_PRESCAN_EXTENSIONS`, the parser-derived block in `buildGlobalResolutionIndex`, and `resolveEdges.test.ts`. I merged both rather than picking a side: the prescan set is now `.go .php .r .sas` + the six JS/TS extensions, and both cross-batch tests are kept.

## The perf fix

Your design decision — derive the index from real parse results rather than a regex, so it cannot drift from what the in-batch path extracts — is right, and I did not want to trade it away. The problem was only *where* the parsing happened: a synchronous main-thread loop inside `buildGlobalResolutionIndex`.

`buildGlobalResolutionIndex` now takes an optional `preParsed` map, and the CLI parses prescan sources on **the worker pool it is already running** for the streaming loop, handing the results in. Absent an entry, each language block falls back to parsing in-index, so tests and library callers behave exactly as before.

Measured on this repo, 180 JS/TS files:

```
index WITHOUT preParsed (main-thread parse): 1.24 s
index WITH preParsed (pool did the work):    0.001 s
```

The parse work still happens — it moved to the pool and spreads across cores instead of blocking. Applied to PHP, R and SAS too, since all four blocks had the same shape.

Added a test that supplies a parse result whose entity does not appear in the source text: it can only resolve if `preParsed` is genuinely used rather than re-parsed.

## Verification

- `core-ingestion`: **35 passed** in `resolveEdges.test.ts`; full suite 280 passed with the same 6 pre-existing failures you documented, no new ones
- `ix-cli`: **698 passed** across 52 files, `npm run typecheck` clean, lint 0 errors / 38 pre-existing warnings

## Still open, and worth its own issue rather than this PR

An incremental map still parses every prescanned file, not just the changed ones — it is parallel now instead of serial, but a one-file edit in a 5,000-file TypeScript repo still does 5,000 parses. The real fix is caching the per-file index keyed on mtime so unchanged files are never re-parsed, which needs cache invalidation and is a bigger change than belongs here. I'd rather land this and file that separately than hold a correct fix behind an optimisation.

I amended your commit rather than stacking, since the fix is threaded through the same lines; you are still the author and I added myself as co-author. Shout if you would rather it were split.

KageBinary (approved):
Blocking items resolved — see the comment above for what I pushed and why. Approving.

Verified after the changes: full `ix-cli` suite green, typecheck clean, lint 0 errors, and `core-ingestion` still at exactly the six pre-existing failures you documented on `origin/main`.

PR #378, opened 2026-08-10
fix(ingest): remove stale graph entities

### LINKED ISSUE #377: [bug] Remapping keeps deleted and renamed symbols in the graph
## Description

After a successful incremental `ix map`, symbols removed by a rename or file deletion remain queryable in the graph. A deleted file can report a clean workspace even though its old nodes and incident edges are still present.

## Steps to reproduce
1. Create a small TypeScript project with `src/provider.ts` exporting `removedTarget` and `src/caller.ts` importing and calling it.
2. Run `ix map . --silent` and confirm `ix callers removedTarget --format json` returns the caller.
3. Either rename `removedTarget` in both files or delete `src/provider.ts`.
4. Run `ix map . --silent` again.
5. Run `ix locate removedTarget --format json`.

## Expected behavior

Entities no longer emitted by the current source should be removed with their incident edges. If a deleted provider file is later recreated, surviving callers should be re-ingested so their dependency edges are restored.

## Actual behavior

The old symbol remains resolvable after a rename, so repeated renames accumulate stale graph entities. Deleting the provider also leaves its old symbol and caller edge behind even though the map finishes successfully.

## CLI output

After deleting the provider and remapping with Ix 0.9.1:

```text
$ ix map . --silent
map: 3 files ...

$ ix locate removedTarget --format json
"resolvedTarget": { "kind": "function", "name": "removedTarget", "path": "src/provider.ts" }
```

## Environment
- OS: macOS 26.6.1
- Node: 26.7.0
- Ix version: 0.9.1

Hiro-Chiba:
Fixes #377.

## Summary

Remove graph entities that disappear after a source edit or file deletion, and restore surviving dependency edges when a deleted file returns.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Compare the previous patch manifest with the new patch and emit explicit node and incident-edge deletions for entities no longer present.
- Detect mapped files that have disappeared from disk and commit deterministic deletion patches through the single-patch endpoint.
- Track surviving dependent source files and re-ingest them when a deleted provider is recreated.
- Keep the workspace-local ingest baseline backward compatible and avoid advancing it after incomplete cleanup.
- Add regression coverage for stale detection, deletion reconciliation, retry safety, and dependent recovery.

## Validation

- `cd ix-cli && npm test` (51 files, 656 tests, parser smoke passed)
- `npm run typecheck`
- `npm run build`
- `npm run lint` (0 errors; 38 existing warnings)
- `core-ingestion/src/__tests__/patchBuilder.test.ts` (2 tests passed)
- Anonymous end-to-end fixture: the deleted symbol and its caller edge disappear after remapping; recreating the provider restores the caller edge
- The complete `core-ingestion` suite has six pre-existing failures identical on `origin/main`: two fixtures require the private `memory-layer` tree, one snapshot is stale, and three existing Rust/Scala expectations differ from current parser output

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/core-ingestion/src/__tests__/patchBuilder.test.ts
+++ b/core-ingestion/src/__tests__/patchBuilder.test.ts
@@ -1,6 +1,11 @@
 import { describe, expect, it } from 'vitest';
 
-import { extractorName, PREVIOUS_EXTRACTORS } from '../patch-builder.js';
+import {
+  buildDeletionPatch,
+  extractorName,
+  PREVIOUS_EXTRACTORS,
+  sourcePatchIdCandidates,
+} from '../patch-builder.js';
 
 describe('patch-builder extractor policy', () => {
   it('tracks the immediate predecessor when the extractor version is bumped', () => {
@@ -16,5 +21,30 @@ describe('patch-builder extractor policy', () => {
     expect(PREVIOUS_EXTRACTORS).toContain(previous);
     expect(PREVIOUS_EXTRACTORS).not.toContain(current);
     expect(PREVIOUS_EXTRACTORS.every(version => /^tree-sitter\/\d+\.\d+$/.test(version))).toBe(true);
+  });
+
+  it('builds a deterministic tombstone for every supported prior patch id', () => {
+    const filePath = 'src/removed.ts';
+    const sourceHash = 'previous-source-hash';
+    const workspaceId = 'deadbeef';
+    const candidates = sourcePatchIdCandidates(filePath, sourceHash, workspaceId);
+    const ops = [{ type: 'DeleteNode', id: '00000000-0000-0000-0000-000000000001' }];
+
+    const patch = buildDeletionPatch(filePath, sourceHash, 'baseline-1', workspaceId, ops);
+    const retry = buildDeletionPatch(filePath, sourceHash, 'baseline-1', workspaceId, ops);
+    const nextDeletion = buildDeletionPatch(filePath, sourceHash, 'baseline-2', workspaceId, ops);
+
+    expect(new Set(candidates).size).toBe(PREVIOUS_EXTRACTORS.length + 2);
+    expect(patch.patchId).toBe(retry.patchId);
+    expect(patch.patchId).not.toBe(nextDeletion.patchId);
+    expect(patch.source).toMatchObject({
+      uri: filePath,
+      extractor: extractorName(),
+      sourceType: 'code',
+      workspaceId,
+    });
+    expect(patch.source.sourceHash).not.toBe(sourceHash);
+    expect(patch.ops).toEqual(ops);
+    expect(patch.replaces).toEqual(candidates);
   });
 });
--- a/core-ingestion/src/patch-builder.ts
+++ b/core-ingestion/src/patch-builder.ts
@@ -95,6 +95,20 @@ export function extractorName(): string {
 /** Previous extractor versions — their patches are superseded when re-ingesting. */
 export const PREVIOUS_EXTRACTORS = ['tree-sitter/1.23', 'tree-sitter/1.22', 'tree-sitter/1.21', 'tree-sitter/1.20', 'tree-sitter/1.19', 'tree-sitter/1.18', 'tree-sitter/1.17', 'tree-sitter/1.16', 'tree-sitter/1.15', 'tree-sitter/1.14', 'tree-sitter/1.13', 'tree-sitter/1.12', 'tree-sitter/1.11', 'tree-sitter/1.10', 'tree-sitter/1.9', 'tree-sitter/1.8', 'tree-sitter/1.7', 'tree-sitter/1.6', 'tree-sitter/1.5', 'tree-sitter/1.4', 'tree-sitter/1.3', 'tree-sitter/1.2', 'tree-sitter/1.1'];
 
+/** Patch ids that may own the active graph entities for a stored source hash. */
+export function sourcePatchIdCandidates(
+  filePath: string,
+  sourceHash: string,
+  workspaceId: string,
+): string[] {
+  const { computePatchId, legacyPatchId } = makeIds(workspaceId);
+  return [
+    computePatchId(filePath, sourceHash, extractorName()),
+    ...PREVIOUS_EXTRACTORS.map(extractor => computePatchId(filePath, sourceHash, extractor)),
+    legacyPatchId(filePath, sourceHash),
+  ];
+}
+
 // nodeId / edgeId / chunkId / computePatchId / legacyPatchId are produced by
 // makeIds(workspaceId) above so every id is scoped to its workspace.
 
@@ -377,6 +391,42 @@ export interface MultiRepoContext {
   repoId?: string;
   /** filePath (workspace-relative, repo-prefixed) -> that file's repo workspace_id. */
   repoWorkspaceOf?: (filePath: string) => string | undefined;
+}
+
+/** Build a patch that retires every entity previously emitted for a deleted file. */
+export function buildDeletionPatch(
+  filePath: string,
+  previousSourceHash: string,
+  deletionToken: string,
+  workspaceId: string,
+  ops: PatchOp[],
+  multiRepo?: MultiRepoContext,
+): GraphPatchPayload {
+  const extractor = extractorName();
+  const deletionHash = crypto
+    .createHash('sha256')
+    .update(`deleted:${previousSourceHash}:${deletionToken}`)
+    .digest('hex');
+  const { computePatchId } = makeIds(workspaceId);
+
+  return {
+    patchId: computePatchId(filePath, deletionHash, extractor),
+    actor: 'ix/ingestion',
+    timestamp: new Date().toISOString(),
+    source: {
+      uri: filePath,
+      sourceHash: deletionHash,
+      extractor,
+      sourceType: sourceType(filePath),
+      workspaceId,
+      ...(multiRepo?.systemId ? { systemId: multiRepo.systemId } : {}),
+      ...(multiRepo?.repoId ? { repoId: multiRepo.repoId } : {}),
+    },
+    baseRev: 0,
+    ops,
+    replaces: sourcePatchIdCandidates(filePath, previousSourceHash, workspaceId),
+    intent: `Deleted ${nodePath.basename(filePath)}`,
+  };
 }
 
 export function buildPatchWithResolution(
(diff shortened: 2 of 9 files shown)


KageBinary (changes requested):
Thanks for contributing — and thanks for taking on the hardest of the four. Stale-entity cleanup is a real gap and the overall shape here (compare the previous patch manifest, emit explicit deletions, track dependents so a recreated file heals its callers) is the right one. The tombstone `patchId` derivation is nicely done: deterministic per `(previousHash, deletionToken)` so a retry is idempotent but a later deletion is not, which is exactly the property you want.

There is one bug I think is a blocker, plus a few things worth reconsidering.

## Blocker — `ix map` after `ix ingest` deletes chunk nodes and never recreates them

`reconcileRemovedEntities` runs **before** `stripMapModeOps`, at both call sites:

```js
patch = await reconcileRemovedEntities(client, patch, sourcePatchIdCandidates(...));
if (mapMode) patch = stripMapModeOps(patch);
```

`stripMapModeOps` filters out chunk `UpsertNode` ops but leaves `DeleteNode` ops untouched. And `chunkId` hashes the start line:

```js
chunkId: (filePath, chunkKind, name, startLine) =>
  deterministicId(`${ns}${normalizePath(filePath)}:chunk:${chunkKind}:${name ?? 'file_body'}:${startLine}`)
```

So any edit that shifts lines changes the chunk ids. Sequence: user runs `ix ingest` (chunks created), edits a file, runs `ix map`. Reconcile sees the old chunk ids missing from the new patch and emits deletions; the strip then removes the replacement upserts.

I confirmed this on your branch with the real `reconcileRemovedEntities` and a copy of `stripMapModeOps` — previous patch `[file-node, chunk-L1, chunk-L10]`, new patch upserting `[file-node, chunk-L1, chunk-L12]`:

```
DELETED : [ 'chunk-L10' ]
UPSERTED: [ 'file-node' ]
=> chunk nodes deleted but NOT recreated: [ 'chunk-L10' ]
```

Before this PR `ix map` simply left chunks alone (stale, but present). Now the cheaper command silently destroys embedding/search data for every file it touches, and only a full `ix ingest` puts it back. Easiest fix is to skip reconciliation entirely in `mapMode`, or to filter chunk-node ids out of `removedNodeIds` when `mapMode` is set — `ix map` never owned that data and shouldn't be deleting it.

## Please reconsider — dropping the `mtimes.size === 0` guard

```js
- if (!ingestCompletedCleanly(parseErrors, commitErrors) || mtimes.size === 0) return false;
+ if (!ingestCompletedCleanly(parseErrors, commitErrors)) return false;
```

I understand why: an emptied repo has to be able to persist an empty baseline. But there is no `filePaths.length === 0` guard anywhere in `ingestFiles`, so *any* run that discovers zero files — an over-broad ignore rule, a bad `cwd`, a discovery glitch — now overwrites a good baseline with an empty one and forces a full re-ingest next time.

The `!fs.existsSync(filePath)` check on `deletedPaths` is a good safeguard and does stop this from turning into mass deletion, which I was glad to see. But the baseline still gets wiped. Gating on `mtimes.size > 0 || deletedPaths.length > 0` would keep the guard while still covering the case you actually need.

## Please reconsider — hash-lookup failure now aborts the whole map

```js
knownHashes = await loadExistingHashes(..., deletedPaths.length > 0);
```

With any deleted file present, a transient `getSourceHashes` failure throws instead of degrading to an empty map, so one network blip fails the entire ingest rather than just the deletion work. Failing closed on *deletion* is right — deleting from an unreliable baseline is how you lose data. But the scope is wider than the risk. Consider catching it and skipping deletion reconciliation for this run (leaving `deletedFiles` in the baseline so the next run retries) while letting the ordinary ingest proceed.

## Perf — per-file round-trips inside the patch loop

`reconcileRemovedEntities` now runs for every file with a `previousHash`, and it is serial I/O in what used to be a pure-CPU loop:

- `sourcePatchIdCandidates` returns **25** ids (current + 23 `PREVIOUS_EXTRACTORS` + legacy), tried one at a time until one is not a 404. Common case is 1 call, but the first `ix map` after a CLI upgrade walks the list for *every* changed file.
- then one `client.entity(nodeId)` per removed node, also serial.

A refactor that removes 50 symbols across 20 files pays 50 sequential `entity()` calls on top of the `getPatch` walk. Worth batching, or at least running the candidate probe concurrently.

## Smaller things

- **Op ordering.** `DeleteNode` ops are emitted before the `UpsertEdge` ops, and the `DeleteEdge` ops for incident edges come *after* the `DeleteNode`s. If the backend cascades incident-edge removal on node delete, those trailing `DeleteEdge`s target rows that are already gone — is that a no-op server-side, or an error? Worth confirming, since it decides whether the ordering is load-bearing.
- **Baseline write amplification.** `onCommitted` calls `saveIngestBaseline` once per committed deletion, rewriting the whole JSON each time. Deleting a directory of 500 files means 500 full rewrites. Durability per-file is a reasonable goal, but a debounce or a single write after the batch would be cheaper.
- **`readStoredPatchEntities` `data.ops` branch** treats `DeleteNode` ids from a previous patch as live entities. Harmless (they re-delete), but it means a tombstone-follows-tombstone sequence re-emits deletions for already-removed nodes.

## Also

The branch is currently conflicting with `main` and will need a rebase.

The test coverage is genuinely good — `planDeletedFileRecovery`, the retry-idempotency assertion on the tombstone id, and the "fails closed when the previous patch has no entity manifest" case are all the right things to pin. Once the map-mode chunk issue is sorted I would be happy to take another pass.

KageBinary:
Rebased onto `main` — this one had real conflicts, so flagging exactly what I changed while resolving them. Your commit and authorship are intact; the two edits below are the only things that differ from what you pushed.

**`main` changed `isFileStale` out from under this PR.** It is now synchronous and takes only a path — the `IxClient` parameter is gone, because it reads a local baseline file rather than the patch log, and it is called per result on `explain`/`locate`/`read` where the `await` used to cost a backend round-trip per file.

Your version still had the old `async (client, filePath)` shape, so the two collided head-on. I kept **`main`'s signature** and **your behaviour**:

```ts
export function isFileStale(filePath: string): boolean {
  const workspaceRoot = path.resolve(resolveWorkspaceRoot());
  const baseline = loadIngestBaseline(workspaceRoot);
  if (!baseline) return false;

  const absolutePath = path.isAbsolute(filePath)
    ? path.resolve(filePath)
    : path.resolve(workspaceRoot, filePath);
  if (!fs.existsSync(absolutePath)) {
    return baseline.files.has(absolutePath) || baseline.files.has(filePath);
  }
  ...
}
```

So your two changes survive — the workspace-relative path resolution, and a deleted-but-mapped file reporting stale instead of the old blanket `if (!fs.existsSync) return false`. That early return was the bug this PR exists to fix, and dropping it is what `main`'s rewrite would otherwise have quietly undone.

The matching test edits: `stale-workspace.test.ts` dropped the now-unused `IxClient` import and the `noPatchReads` stub, and the three call sites in your new "reports a mapped file that was deleted after ingest" case moved to the sync signature.

Verified after the rebase: `npm run typecheck` clean, full `ix-cli` suite **693 passed** (52 files), `patchBuilder.test.ts` passing, lint 0 errors / 38 pre-existing warnings. All four `isFileStale` callers already use the sync form.

The review feedback above is unchanged — in particular the map-mode chunk deletion, which is still the blocking item.

KageBinary:
Rebased onto `main` (clean this time — #382/#384/#386/#387 landed since) and pushed fixes for the blocking item and two of the "please reconsider" points, as `d827cc9`. Your commits are intact; mine sits on top.

## 1. The blocker — map mode was destroying chunk data

`stripMapModeOps` drops every chunk `UpsertNode` but leaves `DeleteNode` alone, and `chunkId` hashes the chunk's start line. So any edit that shifts lines made every downstream chunk look removed, and `ix map` after an `ix ingest` deleted the file's chunks without recreating them — the cheap command silently destroying what the expensive one built.

`reconcileRemovedEntities` now takes a `mapMode` flag and leaves chunk nodes and `CONTAINS_CHUNK`/`NEXT` edges alone. The kind comes from the `client.entity()` lookup **you were already making** for incident edges, so it costs no extra round trip — that turned out to be the neat part of your design, not something I had to work around.

Three tests: chunks untouched in map mode, still deleted outside it, and chunk-carrying edges spared while a `CALLS` edge on the same node is still removed.

## 2. Empty-baseline guard — restored, but scoped

You dropped `mtimes.size === 0` so that deleting the last mapped file could persist an empty baseline. That case is real, but the guard was load-bearing for a different one: there is no `filePaths.length === 0` check anywhere in `ingestFiles`, so *any* discovery failure — over-broad ignore rule, wrong cwd — now overwrote a good baseline with an empty one.

It now asks **why** the map is empty: with deletions recorded this run, empty is a real state and is persisted; without them, discovery is assumed to have gone wrong and the baseline is kept. Your "persists an empty baseline after the last mapped file is deleted" test still passes unchanged, and I added its mirror image.

## 3. Fail-closed lookup — kept, narrowed

Failing closed before deleting is right; deleting from an unreliable baseline is how data gets lost. But letting the throw escape aborted the entire map over one transient blip, taking the ordinary ingest of every changed file with it. The deletions are now dropped for that run — they stay in the baseline and retry next time — and the rest proceeds.

## Verification

- `ix-cli`: **710 passed** across 53 files (was 693), typecheck clean, lint 0 errors / 38 pre-existing warnings
- `core-ingestion`: 279 passed with the same 6 pre-existing failures, no new ones

## Left alone deliberately

- **Per-file round trips.** `sourcePatchIdCandidates` returns 25 ids tried serially, plus one `client.entity()` per removed node. Real, but batching it is a separate change and I did not want to bury the correctness fixes underneath it.
- **Op ordering.** `DeleteNode` still precedes the trailing `DeleteEdge`s for incident edges. I could not answer from the CLI side whether the backend cascades incident-edge removal on node delete — if it does, those trailing deletes target rows that are already gone. Worth confirming before anyone calls the ordering intentional.
- **Baseline write amplification.** `onCommitted` rewrites the whole baseline per deleted file. Bounded, and per-file durability is a defensible trade.

I'm happy with this now aside from those. If you'd rather take the round-trip batching yourself, say so and I'll leave it; otherwise I can follow up separately.

KageBinary (approved):
Blocking items resolved — see the comment above for what I pushed and why. Approving.

Verified after the changes: full `ix-cli` suite green, typecheck clean, lint 0 errors, and `core-ingestion` still at exactly the six pre-existing failures you documented on `origin/main`.

PR #380, opened 2026-08-10
fix(resolve): preserve same-kind ambiguity

### LINKED ISSUE #379: [bug] --kind silently selects among duplicate same-kind symbols
## Description

Passing `--kind` can silently select the first result when multiple exact-name candidates have the same requested kind. Because resolution completes before ambiguity handling, an accompanying `--pick` is ignored as well.

## Steps to reproduce
1. Create two TypeScript files that each export a class named `DuplicateService`.
2. Give the first class one member and the second class two members so the selected target is observable.
3. Run `ix map . --silent`.
4. Run `ix impact DuplicateService --kind class --format json`.
5. Run `ix impact DuplicateService --kind class --pick 2 --format json`.

## Expected behavior

`--kind class` should filter out other kinds but remain ambiguous when multiple classes still match. Adding `--pick 2` should select the second class.

## Actual behavior

Both commands silently select the first class and report one member. The second command ignores `--pick 2`.

Without `--kind`, the same fixture correctly reports both candidates:

```text
Ambiguous symbol "DuplicateService":
  1. class ... DuplicateService in src/alpha/DuplicateService.ts
  2. class ... DuplicateService in src/beta/DuplicateService.ts
```

## CLI output

```json
{
  "resolvedTarget": {
    "kind": "class",
    "name": "DuplicateService"
  },
  "summary": {
    "members": 1
  }
}
```

The output remains the same with `--kind class --pick 2`, although the second class has two members.

## Environment
- OS: macOS 26.6.1
- Node: 26.7.0
- Ix version: 0.9.1 and current `main`

Hiro-Chiba:
Fixes #379.

## Summary

Keep exact-name resolution ambiguous when `--kind` still leaves multiple candidates, so commands do not silently analyze the wrong symbol and `--pick` remains effective.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Remove the resolver shortcut that treated any requested kind as sufficient disambiguation.
- Preserve the existing single-candidate, score-gap, preferred-kind, and path-based resolution rules.
- Add regression coverage for duplicate exact-name classes with `--kind` and `--kind --pick 2`.

## Validation

- `cd ix-cli && npm test` (50 files, 650 tests, parser smoke passed)
- `npm run typecheck`
- `npm run build`
- `npm run lint` (0 errors; 38 existing warnings)
- Resolver-focused tests (2 files, 25 tests passed)
- Anonymous end-to-end fixture: `--kind class` now reports both matching classes; `--kind class --pick 2` selects the second class and reports its two members

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/search-ranking.test.ts
+++ b/ix-cli/src/cli/__tests__/search-ranking.test.ts
@@ -193,6 +193,66 @@ describe("resolveEntityFull: method preferred over chunk for PascalCase names", 
   });
 });
 
+describe("resolveEntityFull: same-kind ambiguity", () => {
+  const duplicateClasses = [
+    {
+      id: "alpha-service-id",
+      name: "DuplicateService",
+      kind: "class",
+      provenance: { sourceUri: "src/alpha/DuplicateService.ts" },
+    },
+    {
+      id: "beta-service-id",
+      name: "DuplicateService",
+      kind: "class",
+      provenance: { sourceUri: "src/beta/DuplicateService.ts" },
+    },
+  ];
+
+  it("stays ambiguous when --kind matches multiple exact-name candidates", async () => {
+    const mockClient = {
+      search: vi.fn().mockResolvedValue(duplicateClasses),
+    } as any;
+
+    const result = await resolveEntityFull(
+      mockClient,
+      "DuplicateService",
+      ["class"],
+      { kind: "class" },
+    );
+
+    expect(result.resolved).toBe(false);
+    if (!result.resolved) {
+      expect(result.ambiguous).toBe(true);
+      if (result.ambiguous) {
+        expect(result.result.candidates.map(candidate => candidate.id)).toEqual([
+          "alpha-service-id",
+          "beta-service-id",
+        ]);
+      }
+    }
+  });
+
+  it("honors --pick after --kind leaves multiple candidates", async () => {
+    const mockClient = {
+      search: vi.fn().mockResolvedValue(duplicateClasses),
+    } as any;
+
+    const result = await resolveEntityFull(
+      mockClient,
+      "DuplicateService",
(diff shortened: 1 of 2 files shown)


KageBinary (approved):
Thanks for contributing! This one is my favourite of the four — small, and it fixes a real correctness bug rather than papering over it.

The premise is right: `--kind` is a **filter**, not a **tiebreaker**. The old shortcut treated "the user named a kind" as consent to pick arbitrarily among everything matching that kind, so two same-named classes resolved to whichever sorted first — and because it returned `"exact"` as the resolution mode, the output actively asserted confidence it did not have. Silently analysing the wrong symbol is worse than asking.

I checked the things a change like this usually breaks, and they hold:

- **`--pick` still reaches the user.** `buildAmbiguous` already emits `"Use --pick <n> or --path to disambiguate."`, so the new ambiguous path is actionable rather than a dead end.
- **The earlier exits still cover the single-match case.** `--kind` narrowing to one candidate returns via the `unique.length === 1` branch with `resolutionMode(best, opts)`, so the ordinary "kind disambiguates" flow is unchanged.
- **No accidental auto-pick downstream.** With the shortcut gone, `--kind` queries now fall through to the preferred-kind and container-kind rules. For two same-kind candidates the kind ranks are equal, so the `topTierKindRanks.every(...)` guard does not fire, and `CONTAINER_KINDS.has(bestKind) && !CONTAINER_KINDS.has(secondKind)` is false when both kinds are identical. Ambiguity is preserved, which is what your test asserts.

Both new tests are testing the actual behaviour change (ambiguity preserved; `--pick 2` selects by index and carries the right path), not just the code path.

Approving. No changes requested.

PR #382, opened 2026-08-10
fix(ingest): resolve PHP calls through typed receivers

### LINKED ISSUE #381: [bug] PHP member calls lose receiver types
## Description

PHP member calls discard their receiver before graph construction. Calls such as `$this->domainService->create()` and `$this->repository->create()` both become a single bare `create` relationship, so same-named methods are deduplicated and the graph can miss or misresolve their targets.

PHP namespace imports also retain their full backslash-separated path, so they do not match the imported class file during cross-file resolution. On an incremental map, unchanged PHP target definitions are not pre-indexed, which can leave newly qualified calls unresolved.

## Steps to reproduce

1. Create a tracked `UseCase.php` file:

```php
<?php

interface DomainService
{
    public function create(): void;
}

interface Repository
{
    public function create(): void;
}

final class UseCase
{
    public function __construct(
        private DomainService $domainService,
        private Repository $repository,
    ) {}

    public function create(): void
    {
        $this->domainService->create();
        $this->repository->create();
    }
}
```

2. Run `ix map . --silent`.
3. Inspect the parsed or stored `CALLS` relationships for `UseCase.create`.
4. Move the interfaces to unchanged imported files, edit only `UseCase.php`, and map again.

## Expected behavior

`UseCase.create` has distinct calls to `DomainService.create` and `Repository.create`, and those targets remain resolved after an incremental map.

## Actual behavior

The parser emits one deduplicated relationship:

```text
UseCase.create -> create
```

The receiver types are lost. Imported or unchanged targets may therefore be missing, ambiguous, or represented by an unresolved destination.

## CLI output

The affected graph reports no distinct callee for either receiver, despite both method calls being present in the source.

## Environment

- OS: macOS 26.6.1
- Node: v26.7.0
- Ix version: 0.9.1

Hiro-Chiba:
Fixes #381.

## Summary
Resolve PHP member calls through declared receiver types and keep those targets stable across incremental maps.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Capture receivers for regular and nullsafe PHP member calls.
- Resolve constructor-promoted properties, typed properties, typed parameters, and `$this` calls to qualified method names.
- Normalize PHP namespace imports to their imported class names.
- Pre-index PHP definitions for cross-batch and incremental resolution.
- Add parser, resolver, and ingest discovery regressions.

## Validation
- `cd ix-cli && npm test` (50 files, 649 tests, parser smoke passed)
- `npm run typecheck`
- `npm run build`
- `npm run lint` (0 errors; 38 pre-existing warnings)
- Targeted PHP and resolver tests (36 tests passed)
- Synthetic namespace-based map followed by a caller-only incremental map; all three `CALLS` edges remained attached to existing method nodes

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/core-ingestion/src/__tests__/queries.php.test.ts
+++ b/core-ingestion/src/__tests__/queries.php.test.ts
@@ -1,0 +1,102 @@
+import { describe, expect, it } from 'vitest';
+
+import { parseFile } from '../index.js';
+
+describe('PHP queries', () => {
+  it('normalizes namespace imports to the imported class name', () => {
+    const result = parseFile(
+      '/repo/UseCase.php',
+      `<?php
+namespace App;
+
+use Vendor\\Contracts\\DomainService;
+      `,
+    );
+
+    expect(result).not.toBeNull();
+    expect(result!.relationships).toContainEqual({
+      srcName: 'UseCase.php',
+      dstName: 'DomainService',
+      predicate: 'IMPORTS',
+      importRaw: 'Vendor\\Contracts\\DomainService',
+    });
+  });
+
+  it('resolves calls through typed properties and method parameters', () => {
+    const result = parseFile(
+      '/repo/UseCase.php',
+      `<?php
+interface DomainService
+{
+    public function create(): void;
+}
+
+interface Repository
+{
+    public function create(): void;
+}
+
+final class UseCase
+{
+    private AuditLogger $auditLogger;
+
+    public function __construct(
+        private DomainService $domainService,
+        private Repository $repository,
+    ) {}
+
+    public function create(Logger $logger): void
+    {
+        $this->domainService->create();
+        $this->repository?->create();
+        $this->auditLogger->write();
+        $logger->write();
+        $this->finish();
+    }
+
+    private function finish(): void {}
+}
+      `,
+    );
+
+    expect(result).not.toBeNull();
+    expect(result!.relationships).toEqual(
+      expect.arrayContaining([
+        { srcName: 'UseCase.create', dstName: 'DomainService.create', predicate: 'CALLS' },
+        { srcName: 'UseCase.create', dstName: 'Repository.create', predicate: 'CALLS' },
+        { srcName: 'UseCase.create', dstName: 'AuditLogger.write', predicate: 'CALLS' },
+        { srcName: 'UseCase.create', dstName: 'Logger.write', predicate: 'CALLS' },
+        { srcName: 'UseCase.create', dstName: 'UseCase.finish', predicate: 'CALLS' },
+      ]),
+    );
+    expect(result!.relationships).not.toContainEqual({
+      srcName: 'UseCase.create',
+      dstName: 'create',
+      predicate: 'CALLS',
+    });
+    expect(result!.relationships).not.toContainEqual({
+      srcName: 'UseCase.create',
+      dstName: 'UseCase.create',
+      predicate: 'CALLS',
+    });
+  });
+
+  it('preserves bare-name fallback when the receiver type is unknown', () => {
+    const result = parseFile(
+      '/repo/Runner.php',
+      `<?php
+function run($service): void
+{
+    $service->create();
+}
+      `,
+    );
+
+    expect(result).not.toBeNull();
+    expect(result!.relationships).toContainEqual({
+      srcName: 'run',
+      dstName: 'create',
+      predicate: 'CALLS',
+    });
+  });
+});
--- a/core-ingestion/src/__tests__/queries.php.test.ts
+++ b/core-ingestion/src/__tests__/queries.php.test.ts
@@ -99,4 +99,83 @@ function run($service): void
       predicate: 'CALLS',
     });
   });
+
+  it('resolves nullable declared types', () => {
+    // `?Service` parses as (optional_type (named_type (name))), not named_type.
+    // Matching only the latter dropped every nullable dependency to a bare name.
+    const result = parseFile(
+      '/repo/Nullable.php',
+      `<?php
+final class Nullable
+{
+    private ?Service $service;
+
+    public function __construct(private ?Repository $repository) {}
+
+    public function run(?Logger $logger): void
+    {
+        $this->service->create();
+        $this->repository->find();
+        $logger->write();
+    }
+}
+      `,
+    );
+
+    expect(result).not.toBeNull();
+    expect(result!.relationships).toEqual(
+      expect.arrayContaining([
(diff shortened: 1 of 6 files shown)


KageBinary (changes requested):
Thanks for contributing! This is careful work — the receiver-type plumbing is well matched to how the existing Python/Go/C++ paths do it, and the `paramTypeMap` scope keys line up correctly with `findEnclosingFunction` (`${className}.${methodName}` on both sides), which is the detail I expected to be wrong and wasn't.

I ran your tests plus a set of probes on your branch. The core mechanism works. Two query gaps are worth closing before this lands, because both are ordinary modern PHP.

## 1. Nullable types silently fall back to bare names

The queries match `type: (named_type (name) @_typed_var_type)`, but `?Service` does not parse as `named_type` — it is `optional_type` wrapping it:

```
private ?Service $s;   → (property_declaration type: (optional_type (named_type (name))) ...)
private Service $s;    → (property_declaration type: (named_type (name)) ...)
```

So on your branch:

```php
class C {
  private ?Service $s;
  public function go(): void { $this->s->create(); }
}
```
resolves to `C.go -> create` rather than `C.go -> Service.create`. Given `?Type` is idiomatic for any nullable dependency, I would expect this to be a large share of real member calls.

## 2. Plain functions are not covered, only methods

The typed-parameter pattern is rooted at `(method_declaration ...)`, so a top-level `function` gets nothing:

```php
function run(Service $svc): void { $svc->create(); }   // → run -> create
```

Your third test pins the bare-name fallback using an *untyped* param, so it passes either way and doesn't catch this.

## Suggested fix — verified on your branch

Accept `optional_type` alongside `named_type`, and alternate the parameter pattern over both declaration forms:

```
; Typed properties: private Service $s / private ?Service $s
(property_declaration
  type: [(named_type (name) @_typed_var_type)
         (optional_type (named_type (name) @_typed_var_type))]
  (property_element
    name: (variable_name (name) @_typed_var_name))) @_typed_var_scope

; Constructor-promoted properties: private ?Service $service
(property_promotion_parameter
  type: [(named_type (name) @_typed_var_type)
         (optional_type (named_type (name) @_typed_var_type))]
  name: (variable_name (name) @_typed_var_name)) @_typed_var_scope

; Typed parameters on methods and plain functions: function run(?Service $s)
[(method_declaration
   parameters: (formal_parameters
     [(simple_parameter
       type: [(named_type (name) @_typed_param_type)
              (optional_type (named_type (name) @_typed_param_type))]
       name: (variable_name (name) @_typed_param_name))
      (property_promotion_parameter
       type: [(named_type (name) @_typed_param_type)
              (optional_type (named_type (name) @_typed_param_type))]
       name: (variable_name (name) @_typed_param_name))]))
 (function_definition
   parameters: (formal_parameters
     (simple_parameter
       type: [(named_type (name) @_typed_param_type)
              (optional_type (named_type (name) @_typed_param_type))]
       name: (variable_name (name) @_typed_param_name))))] @_typed_param_scope
```

No handler changes needed — `childForFieldName('name')` works on `function_definition` too, and with no enclosing class the scope becomes the bare function name, which is exactly what `findEnclosingFunction` returns for a top-level function.

With that applied on your branch:

- `function run(Service $svc)` → `run -> Service.create` ✅
- `private ?Service $s` → `C.go -> Service.create` ✅
- your three PHP tests and `resolveEdges.test.ts` still pass ✅
- full `core-ingestion` suite: 6 failures, the same six you documented as pre-existing on `origin/main` — no new ones ✅

## Deliberately left alone

- **Union types** (`Service|Repo $x`) parse as `union_type` and still fall back to the bare name. That seems right — there is no single receiver type to attribute the call to, so bare is the honest answer. Worth a one-line comment saying so, since the next reader will wonder.
- **Chained and static receivers** (`$this->repo->find()->hydrate()`, `A::make()->run()`, `self::$log->write()`) fall back via `phpReceiverIsSupported = false`. Correct — resolving them needs return-type inference, which is well beyond this PR.

## Minor

`.php` joins the prescan set here, which has the same incremental-map cost I raised on #375 — every unchanged `.php` file gets re-read and re-parsed serially on each `ix map`. The blast radius is much smaller than the JS/TS case in that PR, so I would not block on it, but whatever you land there should probably cover this too.

Happy to re-review once the query is updated.

KageBinary:
Rebased onto `main` — #380 and #384 landed since you opened this, so it was behind.

Authorship is preserved on your commit; only the base moved. Nothing else in the PR was touched, and the review comments above still stand as written.

KageBinary:
Pushed the fix from my review as `9c74d47`, since it was the only thing blocking this and it was already written and tested.

**What changed** — only `core-ingestion/src/queries.ts` and `queries.php.test.ts`. Your commit is untouched.

- `property_declaration`, `property_promotion_parameter` and the parameter query now accept `optional_type` alongside `named_type`, so `?Service` resolves.
- The parameter pattern alternates over `[(method_declaration) (function_definition)]`, so top-level functions get typed parameters too.

No handler change was needed, exactly as expected: `childForFieldName('name')` works on `function_definition`, and with no enclosing class the scope falls back to the bare function name — which is what `findEnclosingFunction` already returns for a top-level function.

**Union types stay bare deliberately.** `Service|Repo` has no single receiver type to attribute the call to, so picking one member arbitrarily would be worse than the fallback. I added a test pinning that so it reads as a decision rather than an oversight.

**Verified**
- 3 new tests; the two covering the fixed gaps fail without the query change (`× resolves nullable declared types`, `× resolves typed parameters on plain functions`)
- `queries.php.test.ts` + `resolveEdges.test.ts`: 39 passed (was 36)
- full `ix-cli` suite: 686 passed / 51 files; `npm run typecheck` clean
- `core-ingestion`: 6 failures, the same six you documented as pre-existing on `origin/main` — no new ones

One thing worth flagging for anyone editing this file later: `PHP_QUERIES` is a template literal, so backticks inside those query comments terminate the string. I used them in a first draft and got `Expected ':' but found ';'` from a completely unrelated-looking line. The comments now use plain text.

This clears the blocking review item. The remaining note is the minor one about `.php` joining the prescan set, which is the same incremental-map cost I raised on #375 and is better dealt with there.

KageBinary (approved):
Blocking item resolved by `9c74d47` — nullable types and plain-function parameters both resolve now, with regression tests that fail without the query change. Verified the full `ix-cli` suite (686) and typecheck clean, and `core-ingestion` still at exactly the six pre-existing failures you documented.

Approving. The remaining prescan-cost note is shared with #375 and belongs there, not here.

PR #390, opened 2026-08-11
fix(cli): register the patches command instead of stubbing it

### LINKED ISSUE #371: ix patches is registered nowhere — registerPatchesCommand is dead code
Found by @Alot1z while writing the agent skill (#368).

\`ix-cli/src/cli/commands/patches.ts:6\` exports \`registerPatchesCommand\`, but nothing imports it — it is absent from \`ix-cli/src/cli/register/oss.ts\`. So \`ix patches\` is an unknown command today.

It is also listed in \`PRO_COMMANDS\` in \`oss.ts\`, which means \`registerProStubs\` currently installs the "requires Ix Pro" stub for it. Users on Pro presumably get the real one; users on OSS get the Pro message rather than an unknown-command error, which is why this has gone unnoticed.

Decide which it is:
- If \`patches\` is meant to be OSS, register it in \`oss.ts\` and drop it from \`PRO_COMMANDS\`.
- If it is meant to be Pro-only, delete \`patches.ts\` from this repo so the stub is the only implementation.

Note the file already carries a complete \`--format <fmt>\` declaration advertising \`llm\`, with no llm renderer behind it — see the \`--format llm\` coverage issue.

KageBinary:
## The bug

`ix-cli` has always contained a complete `ix patches` — client call, `--limit`, `--format text|json|llm`, and `formatPatches` already has a working `llm` branch (`format.ts:242`).

It was never imported by `register/oss.ts`. Meanwhile `patches` sat in `PRO_COMMANDS`, so `registerProStubs` installed the "requires Ix Pro" sentinel over the top of it.

So OSS users got a paywall message for a command this repo ships. Nobody noticed because the stub is a *plausible* answer — had it fallen through to commander's `unknown command` the way `goals` did (#384), it would have been reported long ago.

Found by @Alot1z while writing the agent skill (#368).

## The fix

Register it, and drop it from `PRO_COMMANDS`.

## The part worth reviewing

`@ix/pro` registers a `patches` too. Commander **throws** on a duplicate command name, and Pro's `tryRegister` wraps registration in a `try`/`catch` that swallows it — so on a Kartr install the OSS registration (which runs first, from `registerOssCommands`) is the one that survives, and Pro's is silently skipped.

That is the outcome we want: this version also supports `--format llm`, which Pro's does not. But it means the two implementations' behaviour now matters, so this one uses the same `createClient()` factory Pro's does instead of constructing `IxClient` directly. They are identical today (`createClient` is literally `new IxClient(getEndpoint())`) — the point is that its docstring says Pro commands should prefer it "so auth and endpoint resolution can evolve in one spot", and `patches` is now a command both editions route through.

Verified the swallow rather than assuming it: commander 12.1.0 raises `cannot add command 'patches' as already have command 'patches'` from `_registerCommand`, and `tryRegister`'s `catch` covers the `fn(program)` call, not just the `import`.

## Tests

The stub test asserting `patches` reports Pro is replaced by one asserting it is **not** stubbed. A stub for a command that exists in OSS shadows the real implementation, and because `registerOssCommands` runs first the symptom is silence, not a crash — worth pinning so it cannot come back.

722 tests green. `ix patches --help` verified against a real build.

## Docs

CLAUDE.md said the wrong thing in three places: the gotcha claiming `ix patches` is an unknown command, the `[Pro]` marker on its routing row, and the "only the first three work without Pro" line above that table.

Fixes #371

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -47,8 +47,12 @@ ix map --silent                             # refresh the graph after code chang
   Pro probe diffs against it. Adding a command to `oss.ts` silently makes it
   OSS; removing one makes Pro own it. `registerProCommands` is async and MUST
   be awaited.
-- **`registerPatchesCommand` is defined but never registered** in `oss.ts` —
-  `ix patches` is an "unknown command" today.
+- **`ix patches` is OSS, not Pro** (#371). It is implemented here and registered
+  in `oss.ts`. `@ix/pro` also registers a `patches`; commander throws on the
+  duplicate and Pro's `tryRegister` swallows the throw, so the OSS one — which
+  registers first — wins on a Kartr install too. Do not re-add it to
+  `PRO_COMMANDS`: a stub for a command that exists in OSS shadows the real
+  implementation, and the failure is silent rather than a crash.
 - **`ix upgrade` wipes `~/.ix/cli/compass`.** The Compass assets ship only via
   `ix upgrade`, and re-running the installer re-extracts over them, so a
   re-install can leave `ix view` with no UI. `bootstrap.sh` re-runs `ix upgrade`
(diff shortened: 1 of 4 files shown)


josephismikhail (approved):
Verified the two claims this turns on, both hold.

**Pro shadowing.** `tryRegister`'s `try` covers the `fn(program)` call, not just the `import` (`Ix-pro/src/cli/register.ts:17-23`), so commander's duplicate-name throw is swallowed. `main.ts:69` runs `registerOssCommands` synchronously before the Pro probe at :74, so the OSS registration is the survivor. And `Ix-pro/src/cli/commands/patches.ts` is now identical to this one bar the `llm` in its `--format` description — it already imports `createClient` and `formatPatches` from `@ix/cli`, so switching to `createClient()` here makes them the same call, not merely a similar one.

**One gap, fixed in 89ca55e.** The new test pins that `patches` is absent from the stub list. That is the half that can't silently regress — the other half can. I deleted the `registerPatchesCommand(program)` line from `oss.ts` and the whole suite stayed green (722 passed) while `ix patches` went back to being commander's `unknown command`, i.e. the state #371 was filed about.

Added `registers patches as an OSS command` alongside it, asserting the command exists on the program after `registerOssCommands` and carries `--format`. It fails with `expected undefined to be defined` against that mutation.

Both halves are now pinned. 723 green.

PR #391, opened 2026-08-11
fix(upgrade): stop comparing compass versions across two unrelated series

### LINKED ISSUE #376: ix upgrade compares two unrelated version series to decide if Compass is stale
The bundled Compass is stamped with the **Ix** version; `ix-compass-dist` releases use their **own** series. `isNewer` compares them directly, so the check is only accidentally correct.

Verified on the shipped v0.9.1 linux tarball:

```
compass/.version = 0.9.1          # the Ix version, not a compass version
```

and the latest `ix-compass-dist` release is **v0.3.0** (2026-08-09).

`upgrade.ts` does `isNewer(compassLatest, compassCurrent)` — `isNewer("0.3.0", "0.9.1")` → `false`, so nothing happens and `ix view` keeps the correct bundled build. That is the right outcome, but only because Ix's numbers currently happen to be larger.

**The moment `ix-compass-dist` tags anything above the running Ix version** — v1.0.0, or just v0.9.2 — `isNewer` returns true and `ix upgrade` replaces a correct, newer bundled Compass with the dist build. That is the same failure v0.9.1 just fixed (#365/#366), reached by a different route: there the stamp was missing so `current` read as `none`; here the stamp is present but denominated in the wrong series.

Options:
- Stamp the bundle with the **compass** version the release was built from (v0.3.0 was built from `system-compass@7f98724`), so both sides of the comparison are the same series.
- Or stamp the source commit and compare identity rather than order — the question is really "is the installed bundle the one this CLI shipped with", not "is it numerically older".

Either way, `getInstalledCompassVersion` and `writeCache(compassLatest)` should not be feeding two different numbering schemes into one `isNewer`.

KageBinary:
## The bug

`compass/.version` is written by two producers in two different version series, and recorded no way to tell them apart:

| producer | writes | series |
|---|---|---|
| `release.yml` | the **Ix** version (`0.9.2`) | Ix |
| `install.sh` / `install.ps1` / the compass-upgrade path | the **ix-compass-dist** release (`0.3.0`) | compass |

`isNewer(compassLatest, compassCurrent)` compared whichever happened to be on disk against the dist series. That is correct only while Ix's numbers are the larger ones. **The first ix-compass-dist tag above the running Ix version** — a v1.0.0, or just a v0.9.2 — makes `isNewer` true and replaces a *newer* bundled compass with an older dist build.

Same class as #365/#366, reached from the other end: there the stamp was missing so `current` read as `none`; here the stamp is present but denominated in the wrong series.

## It is not only latent

`install.sh` stamped every install with the latest **dist** release number regardless of where the compass came from. The compass it installs always comes from the release tarball — so it mislabelled a release bundle as a dist build, and clobbered the correct stamp the tarball already carried.

Reproduced on a live v0.9.x install:

```
~/.ix/cli/compass/.version   → 0.2.0     # written by install.sh, from ix-compass-dist
~/.ix/.version-check.json    → compassLatest: 0.3.0
```

`ix` prints "Compass update available" and `ix upgrade` would replace a bundle built from system-compass `main` at v0.9.x release time with dist v0.3.0.

## The fix

The stamp declares its own provenance, and the comparison only runs where it means something:

```
source=release        source=dist
ix=0.9.3              version=0.3.0
compass=7f98724
```

**A release bundle is never replaced by a dist download.** It was built from `main` when the CLI was cut, so it is at least as new as any dist release published before it; one published after arrives with the next Ix release, which bundles `main` again.

The dist download stays the **repair** path for a missing or gutted bundle. That still fires: `getInstalledCompassVersion()` reports `0.0.0` when `index.html` is absent, which no real release number matches.

## Backwards compatibility

A bare number is a legacy stamp — everything shipped up to and including v0.9.2 — and reads as **dist-series**. That is what `install.sh` wrote, it is the majority of installs in the wild, and it leaves them exactly as correct as they are today. Release-bundled legacy stamps stay accidentally-correct until compass-dist passes the Ix version, i.e. no worse than before. New releases carry the explicit form and are immune.

## Why the commit, not a version

`system-compass`'s `package.json` version is not maintained against the dist release series — it reads `0.2.0` while dist is on `v0.3.0` (built from `7f98724`). So stamping the package version would have been a third wrong number. The commit is the only reliable identity the bundle has, and it is passed from the `compass` job to `build` as a job output.

## Both installers stop calling ix-compass-dist

Neither ever installs from there, so neither needs to know what it has published. `resolve_compass_version` / `Get-CompassLatestVersion` are removed. `ix upgrade` asks that question, at the one point where it can also tell which series the installed bundle belongs to.

Each installer now prefers the stamp the tarball shipped with, and only writes one if the bundle has none.

## Tests

`upgrade-compass-stamp.test.ts` — release form, dist form, bare legacy number, empty/malformed, CRLF (how a Windows bundle arrives), and a case pinning the inversion itself.

728 tests green. `bash -n` and `sh -n` clean on install.sh; release.yml parses and `needs.compass.outputs.sha` resolves in the `build` job.

Fixes #376

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -59,6 +59,13 @@ jobs:
   compass:
     if: ${{ github.event_name != 'pull_request' }}
     runs-on: ubuntu-latest
+    outputs:
+      # The system-compass commit this bundle was built from. Recorded in the
+      # bundle's .version stamp so a build is identifiable after the fact —
+      # system-compass's package.json version is not maintained against the
+      # ix-compass-dist release series (it read 0.2.0 while dist was on v0.3.0),
+      # so the commit is the only reliable identity the bundle has.
+      sha: ${{ steps.build-compass.outputs.sha }}
     steps:
       - name: Set up Node.js
         uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
@@ -70,6 +77,7 @@ jobs:
       # a fully green release that shipped an empty compass/ directory and a
       # broken `ix view` for every user. A compass failure must fail the release.
       - name: Build System Compass
+        id: build-compass
         env:
           COMPASS_TOKEN: ${{ secrets.COMPASS_TOKEN }}
         run: |
@@ -98,6 +106,9 @@ jobs:
               exit 1
             }
           cd system-compass
+          # Capture the source commit before building, so the stamp written in
+          # the package job can identify this bundle (see the job's `outputs`).
+          echo "sha=$(git rev-parse --short HEAD)" >> "$GITHUB_OUTPUT"
           npm ci --legacy-peer-deps
           npm run build
           if [ ! -f dist/index.html ]; then
@@ -256,6 +267,9 @@ jobs:
           # A PR dry-run skips the compass job, so an absent bundle is expected
           # there and only there.
           REQUIRE_COMPASS: ${{ github.event_name != 'pull_request' }}
+          # Empty on a PR dry-run (the compass job is skipped); the stamp falls
+          # back to `unknown`, which is fine because no bundle is staged either.
+          COMPASS_SHA: ${{ needs.compass.outputs.sha }}
         run: |
           STAGING="ix-${VERSION}-${TARGET}"
           # Compute the checksum with node (always present on every runner);
@@ -282,16 +296,25 @@ jobs:
             # Windows an unrelated tar failure masked it, and everywhere else the
             # swap succeeded and silently reverted the map to a June build.
             #
-            # The Ix version rather than a compass version: the bundle is built
-            # from compass main at release time, so it is always at least as new
-            # as any ix-compass-dist release cut before it. That makes the Ix
-            # version a monotonic stand-in, and isNewer() still offers a genuine
-            # dist upgrade published later at a higher number.
+            # `source=release` is load-bearing, not decoration. This bundle is
+            # built from system-compass main and has no ix-compass-dist release
+            # number, so the version recorded here is an *Ix* version. install.sh
+            # and the compass-upgrade path write the same file with a *dist*
+            # version. Until v0.9.2 the stamp was a bare number with no way to
+            # tell the two apart, so `isNewer(distLatest, stamp)` compared two
+            # unrelated series and was correct only while Ix's numbers happened
+            # to be the larger ones — the first dist tag above the running Ix
+            # version would have replaced this newer bundle with an older dist
+            # build (Ix#376). The marker lets the CLI skip the comparison
+            # entirely for release bundles, which is the only correct answer:
+            # a dist release published after this one arrives with the next Ix
+            # release, which bundles main again.
             #
-            # install.sh writes this same file after extracting, using the dist
-            # release number. Either value suppresses the spurious update; the
-            # stamp being present at all is what matters.
-            printf '%s' "$VERSION" > "$STAGING/compass/.version"
+            # Keep the key=value shape. parseCompassStamp() reads a bare number
+            # as dist-series for backwards compatibility, so dropping the marker
+            # silently reintroduces the bug rather than failing loudly.
+            printf 'source=release\nix=%s\ncompass=%s\n' \
+              "$VERSION" "${COMPASS_SHA:-unknown}" > "$STAGING/compass/.version"
             test -s "$STAGING/compass/.version" \
               || { echo "::error::failed to stamp $STAGING/compass/.version"; exit 1; }
             COMPASS_STAGED=1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -296,24 +296,32 @@ jobs:
             # Windows an unrelated tar failure masked it, and everywhere else the
             # swap succeeded and silently reverted the map to a June build.
             #
-            # `source=release` is load-bearing, not decoration. This bundle is
-            # built from system-compass main and has no ix-compass-dist release
-            # number, so the version recorded here is an *Ix* version. install.sh
-            # and the compass-upgrade path write the same file with a *dist*
-            # version. Until v0.9.2 the stamp was a bare number with no way to
-            # tell the two apart, so `isNewer(distLatest, stamp)` compared two
-            # unrelated series and was correct only while Ix's numbers happened
-            # to be the larger ones — the first dist tag above the running Ix
-            # version would have replaced this newer bundle with an older dist
-            # build (Ix#376). The marker lets the CLI skip the comparison
-            # entirely for release bundles, which is the only correct answer:
-            # a dist release published after this one arrives with the next Ix
-            # release, which bundles main again.
+            # The `+release.<sha>` marker is load-bearing, not decoration. This
+            # bundle is built from system-compass main and has no
+            # ix-compass-dist release number, so the version recorded here is an
+            # *Ix* version. install.sh and the compass-upgrade path write the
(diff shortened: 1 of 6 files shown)


josephismikhail (approved):
The diagnosis is right and the live reproduction settles it — `.version` really was denominated in the wrong series, and `source=` is the correct shape for the answer. Checked the parts that could bite:

- `build` does declare `needs: compass`, so `needs.compass.outputs.sha` resolves; on a PR dry-run compass is skipped and the `${COMPASS_SHA:-unknown}` fallback covers it.
- No leftover callers of `resolve_compass_version` / `Get-CompassLatestVersion` — both survive only as comments.
- `$Version` is assigned at `install.ps1:207`, well above the stamp block at :532.
- The `0.0.0` repair check correctly sits *ahead* of the source check, so a gutted release bundle is still repaired rather than pinned forever.

**One gap, fixed in c415461.** `shouldOfferCompassUpgrade` read the stamp itself, so the only thing the new test file could reach was `parseCompassStamp` — a pure function that was never the bug. I deleted the `readCompassStamp().source === "release"` line, which is the entire behavioural change here, and all 728 tests passed. The case named `pins the inversion the marker exists to prevent` survives it too: `expect(parseCompassStamp(...).source).toBe("release")` restates the parser rather than the decision.

Made the decision take the stamp as an argument (`shouldOfferCompassUpgradeFor`) and exported it. `installedCompass()` now returns the whole stamp rather than the version alone, which also drops a second read of `compass/.version` per call.

Five cases now cover it: the inversion, the dist-series comparison that must still happen, the repair path for a gutted bundle, the offline case, and a legacy bare stamp. Two mutations verified failing — deleting the source check, and swapping it ahead of the `0.0.0` check (that one is #365/#366 again: `ix view` never repairs).

733 green.

josephismikhail (changes requested):
Retracting my earlier approval. A deeper review turned up a blocking backward-compatibility break that I verified empirically, so this should not merge in its current form.

## Blocking: the new stamp format is unreadable by every already-shipped CLI

The shipped reader is `getTrackedVersion` (`upgrade.ts:199-206`) — `readFileSync(...).trim()`, i.e. **the whole file**. The upgrade sequence matters here: `ix upgrade` swaps in the new tree (including the new-format `compass/.version`), and then the *still-running old process* continues into its own compass block and reads that file with the old parser.

I ran the new format through `origin/main`'s own `isNewer`:

```
dist 0.3.0 vs release-stamp 0.9.4 -> isNewer=false
dist 0.3.0 vs release-stamp 1.0.0 -> isNewer=TRUE
dist 0.3.0 vs release-stamp 1.1.0 -> isNewer=TRUE
dist 0.3.0 vs legacy bare  0.9.4 -> isNewer=false
```

`splitVersion` yields `Number("source=release\nix=1") = NaN -> 0` for the major, so a 1.x release parses as `[0,0,0]`. Today (0.9.x) the damage is a garbled three-line version string in `[ok] Compass already on the latest version (...)`. **At the first 1.0.0 release, every CLI older than this one downloads ix-compass-dist over the newer bundled compass** — the exact #376 downgrade this PR exists to prevent, caused by the format change itself.

Suggested fix that keeps the whole design: make the stamp a **single-line semver with build metadata**, `0.9.4+release.7f98724`. `splitVersion` already does `v.split("+")[0]` with the comment "Build metadata (`+sha`) never participates in precedence" — so old CLIs read `0.9.4` and behave exactly as they do today, while `parseCompassStamp` reads the `+release.<sha>` suffix for provenance. Dist stays a bare number, which is already the legacy-compatible form.

## Also blocking

**`0.0.0` is overloaded** (`upgrade.ts:298`). The doc comment says only a missing `index.html` produces it, but `parseCompassStamp` also returns `{source:"release", version:"0.0.0"}` for a release stamp missing its `ix=` key — a `.version` truncated by a full disk or a kill during the non-atomic write. With the bundle perfectly serviceable, the repair branch fires ahead of the source check and downgrades it. Worth distinguishing "absent" from "malformed" explicitly rather than through the version number.

**`install.sh:1094` stamps `source=release` on a compass it did not install.** The stamp block runs even when the CLI install was skipped (`check_installed_version` == `$VERSION`, or `SKIP_CLI_INSTALL=1`). If `compass/` holds an unstamped *dist* bundle — reachable via `ix upgrade`'s own `stage="stamp"` failure, or Ctrl-C between extract and stamp — it is now permanently labelled `release`, and `shouldOfferCompassUpgradeFor` refuses every future compass update while printing "Compass is the build bundled with Ix <v>", which is false. The deleted code re-stamped with the dist number and stayed self-correcting.

**`install.sh:1093`'s `[ ! -s .version ]` never repairs a mislabelled stamp** — so the fix does not reach the installs the PR body calls the majority in the wild. Re-running the installer on an already-current version skips the extract, so no tarball stamp is laid down, and the bare `0.3.0` survives.

**`IX_VERSION=0.9.0` pins** stamp `source=release ix=0.9.0` onto a months-old bundle and exempt it from dist upgrades forever. `resolve_version` also falls back to `0.1.0` when the API is unreachable.

## Non-blocking

Gating the compass block behind `shouldOfferCompassUpgrade` also gates the orphan sweep inside it. `main` has exactly two sweep sites — one behind "CLI needs updating", one behind "compass needs updating" — and the second exists as the fallback for the first, per its own comment. After this change a user on the latest CLI with a release compass sweeps nothing. I traced it and the severe cases self-heal (a gutted compass reports 0.0.0 and repairs; an interrupted CLI upgrade leaves the CLI stale), so this is stranded orphan directories rather than breakage — but it undercuts #392's reason for sweeping downloads at all.

The diagnosis and the live reproduction in the PR body are right, and the provenance idea is the correct answer. It is the encoding that needs to change.

josephismikhail (approved):
Addressed in 3f70e75. Re-verified end to end.

**Stamp format.** Provenance now rides in semver build metadata on one line — `0.9.3+release.7f98724` for a release bundle, a bare number for dist. `splitVersion` already drops everything after `+`, so every already-shipped CLI reads `0.9.3` and behaves exactly as it does today. The design intent is unchanged; only the encoding moved.

**`present` replaces the `0.0.0` sentinel.** `installedCompass()` carries whether `index.html` is on disk, so a stamp truncated mid-write no longer masquerades as a missing bundle and triggers a downgrade of an intact one. The present-check still sits ahead of the source-check, so a gutted release bundle is repaired.

**install.sh only stamps what it installed.** Gated on a new `CLI_EXTRACTED` flag. The block runs even when the install was skipped, and in that case the compass on disk can be an unstamped dist bundle — labelling that `+release` froze it against every future update. When we did extract, the tarball's own stamp has already landed, which is also what repairs a mislabelled install. `install.ps1` needs no such gate: it has no already-current skip path.

**New producer drift test.** `upgrade-compass-stamp-producers.test.ts` reads `release.yml`, `install.sh` and `install.ps1`, renders the literal each one writes, and pushes it through the real `parseCompassStamp`. Three producers in three languages with no compile-time link to the reader is how this class of bug keeps recurring. Verified non-vacuous — restoring the multi-line form fails both properties it asserts.

Mutations re-run against the reworked code: deleting the release guard fails two cases, and reverting the stamp producer fails two more.

744 green. `bash -n` and `sh -n` clean on install.sh, release.yml parses, install.ps1 parses clean.

Left as-is deliberately: an `IX_VERSION` pin stamps `+release` on that release's bundle and exempts it from dist upgrades. That reads correct to me — the user pinned a release, and the repair path still fires if the bundle is gutted — but flagging it since the review raised it.

Issue #385, opened 2026-08-10
[bug] ix upgrade breaks Windows CLI when upgrading from 0.8.1 to 0.9.1


RMA1313:
## Description

Running `ix upgrade` on Windows broke the installed Ix CLI while upgrading from version `0.8.1` to `0.9.1`.

Before the upgrade, `ix` worked normally. After the upgrade completed, invoking `ix` failed because the generated wrapper at `~\.ix\bin\ix.cmd` pointed to a CLI path that no longer existed.

## Steps to reproduce

1. Start with a working Ix `0.8.1` installation on Windows.
2. Run:

   ```powershell
   ix upgrade
   ```
3. After upgrading to `0.9.1`, run:

   ```powershell
   ix
   ```

## Expected behavior

`ix upgrade` should upgrade the CLI in place and leave the `ix` command functional.

After upgrading, `ix --version` should report `0.9.1` and all existing commands should continue to work.

## Actual behavior

After upgrading from `0.8.1` to `0.9.1`, the `ix` command became unusable.

The wrapper under:

```text
C:\Users\Win 10\.ix\bin\ix.cmd
```

attempted to execute:

```text
C:\Users\Win 10\.ix\bin\..\cli\ix.cmd
```

but that target was no longer available after the upgrade.

Re-running the full Windows installer restored the CLI.

## CLI output

```text
PS C:\Users\Win 10\Documents\project\coin> ix

'"C:\Users\Win 10\.ix\bin\..\cli\ix.cmd"' is not recognized as an internal or external command,
operable program or batch file.
```

## Environment

* OS: Windows
* Node: v24.18.0
* Ix version before upgrade: 0.8.1
* Ix version being upgraded to: 0.9.1

KageBinary:
Thank you for the issue — this is an unusually good report. Naming the exact path the wrapper tried to execute is what made it diagnosable in one pass, and noting that reinstalling restored it confirmed the cause before anyone had to reproduce it.

## Unblocking you now

```powershell
irm https://ix-infra.com/install.ps1 | iex
```

That rewrites the launcher and restores the CLI, which matches what you already found. You should not need to uninstall first, and your graph and config are untouched — the breakage is only in `~\.ix\bin\ix.cmd`.

## What happened

You were on the wrong side of a one-way door.

[...]

PR #392, opened 2026-08-11
fix(upgrade): stage downloads under IX_HOME, not TEMP

KageBinary:
## Why

#349 is a Windows install that died after extraction because TEMP arrived in 8.3 short form (`C:\Users\WIN10\~1\...`) on a profile named `C:\Users\Win 10`. The reporter confirmed that pointing `TEMP`/`TMP` at a path without a space let the same install finish.

#352 moved `install.ps1`'s scratch off TEMP for that reason — and said so in its own comment:

> Note `ix upgrade` still stages through `os.tmpdir()` (upgrade.ts), which on Windows is TEMP verbatim. If this class is real, it is unfixed there.

This is that half. It is the worse of the two places for the failure to bite: a failed install leaves a machine with no Ix on it, but a **failed upgrade breaks an install that was working** — which is exactly the shape of #385.

## What changed

Both download directories move to `IX_HOME`, joining the staging and backup directories already there. `IX_HOME` comes from `USERPROFILE` in its long form. `mkdirSync(IX_HOME)` first, because a first upgrade after a manual install can reach this before anything has created it.

`sweepUpgradeOrphans` learns `.cli-download-` and `.compass-download-`. Leaving TEMP also leaves the OS's own cleanup, so without this an upgrade killed mid-download parks a multi-MB archive in `~/.ix` forever.

The prefixes are deliberately `-download-` and not `-backup-`: `recover()` renames the newest `.cli-backup-*` over the install directory, and an archive must never be a candidate for that. There is already a test pinning that reasoning for `install.ps1`'s zip; this adds the matching one.

## The part I nearly broke

**The sweep call had to move above the download.** It sat between the download and the extract, which was safe only while the scratch lived in `os.tmpdir()`. Once the sweep reclaims `.cli-download-*`, from that position it deletes the archive it is about to extract — breaking *every* upgrade on every platform, not just Windows.

Caught while tracing the ordering rather than by a test, so there is now a test for it. Above the download, the sweep also puts an interrupted install back even when the download then fails, which the old order could not do. The compass branch already had this ordering; the CLI branch was the odd one out.

## Tests

724 green. Two new cases: an interrupted download's scratch is reclaimed, and a download directory is not treated as an install worth recovering.

`npm run lint` — 0 errors (38 pre-existing warnings). The now-unused `tmpdir` import is dropped.

## Not verified

No Windows machine here, so the 8.3 path itself is still unreproduced — see the note on #349 about the mechanism never having been pinned down. What this does is take the upgrade path off TEMP entirely, which fixes the failure under every reading of it.

Refs #349

### DIFF
--- a/ix-cli/src/cli/__tests__/upgrade-archive-shape.test.ts
+++ b/ix-cli/src/cli/__tests__/upgrade-archive-shape.test.ts
@@ -5,6 +5,7 @@ import { tmpdir } from "node:os";
 
 import {
   installStagedTree,
+  prepareDownloadDir,
   soleChildDir,
   swapInStagedTree,
   sweepUpgradeOrphans,
@@ -324,5 +325,51 @@ describe("sweepUpgradeOrphans", () => {
     sweepUpgradeOrphans(root, installDir);
 
     expect(existsSync(installDir)).toBe(false);
+  });
+});
+
+/**
+ * The ordering that #349's fix turns on, and the one thing about it no test
+ * reached while it was written inline in the command action: the sweep must run
+ * BEFORE the download scratch is created. Once the sweep reclaims
+ * `.cli-download-*`, running it afterwards deletes the archive the extract is
+ * about to read — breaking every upgrade on every platform, not just the
+ * Windows one. Both call sites go through this function so the order cannot
+ * drift back apart.
+ */
+describe("prepareDownloadDir", () => {
+  it("reclaims an earlier run's scratch but keeps the one it just made", () => {
+    const installDir = join(root, "cli");
+    mkdirSync(join(root, ".cli-download-stale1"), { recursive: true });
+    writeFileSync(join(root, ".cli-download-stale1", "ix-0.9.1-linux-amd64.tar.gz"), "old");
+
+    const dir = prepareDownloadDir(root, installDir, ".cli-download-");
+
+    expect(existsSync(join(root, ".cli-download-stale1"))).toBe(false);
+    // Swept after the mkdtemp instead of before, this directory is gone too and
+    // the download that follows writes an archive nothing will extract.
+    expect(existsSync(dir)).toBe(true);
+
+    writeFileSync(join(dir, "ix-0.9.2-linux-amd64.tar.gz"), "gz");
+    expect(existsSync(join(dir, "ix-0.9.2-linux-amd64.tar.gz"))).toBe(true);
+  });
+
+  it("puts an interrupted install back before downloading anything", () => {
+    // The other half of sweeping first: the recovery happens even if the
+    // download then fails, which the old position could not do.
+    const installDir = join(root, "cli");
+    mkdirSync(join(root, ".cli-backup-999", "cli", "dist", "cli"), { recursive: true });
+    writeFileSync(join(root, ".cli-backup-999", "cli", "dist", "cli", "main.js"), "// prev\n");
+
+    const dir = prepareDownloadDir(root, installDir, ".cli-download-");
+
+    expect(readFileSync(join(installDir, "cli", "dist", "cli", "main.js"), "utf-8")).toBe("// prev\n");
+    expect(existsSync(dir)).toBe(true);
+  });
+
+  it("creates IX_HOME when a manual install never did", () => {
+    const ixHome = join(root, "fresh");
+    const dir = prepareDownloadDir(ixHome, join(ixHome, "cli"), ".compass-download-");
+    expect(existsSync(dir)).toBe(true);
   });
 });
--- a/ix-cli/src/cli/__tests__/upgrade-archive-shape.test.ts
+++ b/ix-cli/src/cli/__tests__/upgrade-archive-shape.test.ts
@@ -291,4 +291,38 @@ describe("sweepUpgradeOrphans", () => {
   it("is a no-op when IX_HOME does not exist yet", () => {
     expect(() => sweepUpgradeOrphans(join(root, "nope"), join(root, "nope", "cli"))).not.toThrow();
   });
+
+  // `ix upgrade` downloads into `.cli-download-*` / `.compass-download-*` under
+  // IX_HOME rather than os.tmpdir(), which on Windows is TEMP verbatim and is
+  // the path #349 died on. Leaving TEMP also left the OS's own cleanup, so a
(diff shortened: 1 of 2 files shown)


josephismikhail (approved):
Taking the upgrade path off TEMP is right, and the reasoning for `-download-` over `-backup-` is correct — `recover()` renames the newest `.cli-backup-*` over the install directory, so an archive must never match that prefix.

**One gap, fixed in c8754ba.** The PR calls the sweep-above-download move "the part I nearly broke" and says "there is now a test for it". There wasn't. The two new cases pin *which prefixes the sweep reclaims*, which is what makes the ordering matter, but not the ordering itself — that lived inline in the command action where no test reaches it. I moved the call back below the download and all 724 tests passed, on a change that breaks every upgrade on every platform because the sweep then deletes the archive it is about to extract.

Both call sites now go through `prepareDownloadDir(ixHome, installDir, prefix)`, which does mkdir → sweep → mkdtemp. The order is inside one testable function instead of duplicated across two sites as a comment. Three cases: an earlier run's scratch is reclaimed while the one just created survives, an interrupted install is restored before any download happens, and IX_HOME is created when a manual install never did. Inverting the order inside the helper fails all three.

Worth noting the compass branch got the same treatment, so it no longer sweeps and then separately mkdirs — one shape on both paths.

727 green, lint clean.

josephismikhail (changes requested):
Retracting my earlier approval — a deeper review found two correctness regressions introduced by moving the scratch into `IX_HOME`, one of which is data loss.

## Blocking: a failed download can delete the user's only working compass

`cleanupCompassSwap:615` tests `existsSync(join(compassDir, "index.html"))` and its comment is explicit about why: a `compass/` holding only a `.version` is a real state, and a bare `existsSync` "would call that 'back in place' and drop a backup the caller had just told the user to go and rescue."

`recover()` inside `sweepUpgradeOrphans` uses exactly that bare `existsSync(dest)`. So with `compass/` left as a `.version`-only husk and `.compass-backup-1234` holding the only real bundle, `recover()` declines to restore, and the loop below then deletes the backup.

On `main` that sweep ran *after* a successful download, so an offline run exited before reaching it and the backup survived for the next attempt. Hoisting the sweep above `curl` means **a run that installs nothing now destroys it** — the user sees only "[error] Failed to download", and `ix view` is permanently blank with no copy left on disk.

The two tests can't see this because they never construct the husk. Aligning `recover`'s test with `cleanupCompassSwap`'s (`index.html`, not bare `existsSync`) fixes it and makes the two agree.

## Blocking: concurrent `ix upgrade` deletes an in-flight download

The scratch moved from `os.tmpdir()`, where no sweep could match it, into `IX_HOME` with a swept prefix. Process A creates `.cli-download-AAAAAA` and starts a curl with a 300s timeout; process B — a second terminal, or `bootstrap.sh` re-running `ix upgrade` when it finds compass missing — calls `prepareDownloadDir` and deletes A's directory wholesale. On Linux/macOS curl's fd stays valid so A exits 0, then `tar -xzf` hits a missing path and A reports "Failed to extract CLI update" on a good download.

`.cli-staging-<pid>` has the same shape on main, but its window is the few seconds of extraction; this one is the whole download. Some ownership marker (pid in the name, skipped when that pid is live) would close both.

## Worth addressing

- `prepareDownloadDir` in the compass branch is the first filesystem write under `IX_HOME` and sits *outside* the try/catch that deliberately keeps compass failures non-fatal. On an unwritable `~/.ix` this turns "[!!] Compass extract failed / ix view keeps working" into a hard abort that also skips `writeCache`.
- Two bare `rmSync` calls (959, 1264) skip the retry treatment `rmQuiet` exists for — and this PR moves those directories into the user profile, where AV and indexer handles are the norm, which is precisely the case `rmQuiet`'s comment describes.

## One question before this lands

#349's reporter is on `C:\Users\Win 10` and their confirmed workaround was pointing TEMP at a space-free path. This moves the scratch from `C:\Users\WIN10~1\AppData\Local\Temp\...` (no space — 8.3 aliases never contain one) to `C:\Users\Win 10\.ix\.cli-download-...` (space). If the failure is space-driven rather than short-path-driven, the upgrade path now has the space where it previously didn't. "Refs #349" rather than "Fixes" is the right call, and the PR is honest that it's unreproduced — but "fixes the failure under every reading of it" doesn't hold under the space reading. Nothing in the suite sets `IX_HOME` to a path with a space.

Taking the upgrade path off TEMP is still right, and the `-download-` vs `-backup-` prefix reasoning is correct. It's the sweep's new reach that needs bounding.

josephismikhail (approved):
Addressed in 5432c49 and b4d9c00, and rebased onto main now that #391 has landed.

**The data-loss path.** `recover()` now takes the file that proves the thing is real — `index.html` for the compass, `cli/dist/cli/main.js` for the install — so it applies the same test `cleanupCompassSwap` already applies. The two no longer disagree about a husk, so a torn swap restores the backup instead of the sweep deleting it. Verified: reverting to the bare `existsSync` fails two of the new cases.

**The concurrency window.** Download directories carry their pid (`.cli-download-<pid>-<rand>`) and the sweep skips one whose process is still addressable — `process.kill(pid, 0)`, treating EPERM as live. Unparseable names are treated as dead so the pre-pid naming stays collectable, and the guard is scoped to download prefixes so staging/backup reclaim behaviour is untouched. Verified: removing the guard fails the live-process case.

**The two bare `rmSync` calls** are now `rmQuiet`, which is the whole reason that helper exists — and this PR is what moves those directories from TEMP into the user profile.

**One more, found while composing the two PRs** (b4d9c00): both sweep calls sit behind "something needs updating", and #391 tightened the compass gate to skip release-bundled compasses — which is every install from v0.9.3 on. So on a healthy current install neither would fire and nothing would ever be reclaimed, which defeats the reason for sweeping the downloads at all. There is now one unconditional sweep at the top of a real `ix upgrade`; the calls inside `prepareDownloadDir` stay, since that is where the ordering invariant lives.

753 green on the rebased result, lint clean.

**Still open, and deliberately not resolved here:** whether #349 is space-driven or short-path-driven. The scratch now sits under `C:\Users\Win 10\.ix\...`, which contains the space, where the TEMP path it replaces did not. `Refs #349` is the right framing until someone reproduces it — worth keeping the issue open rather than closing it on this.

PR #403, opened 2026-08-12
fix(config): respect workspace path boundaries

### LINKED ISSUE #402: [bug] Workspace lookup matches sibling path prefixes
## Description

Workspace lookup treats any configured root that is a string prefix of the current directory as an ancestor. For example, a workspace rooted at `/work/app` is selected while the CLI is running from the unrelated sibling `/work/app-copy`.

This scopes read commands to the wrong graph. A symbol mapped from `/work/app` can therefore appear in `ix stats`, `ix locate`, and other reads issued from `/work/app-copy` before that sibling has been registered.

## Steps to reproduce

1. Map a separate workspace first so it remains the default workspace.
2. Create two sibling directories such as `/work/app` and `/work/app-copy`.
3. Put a uniquely named function in `/work/app`, initialize it as a Git repository, and run `ix map /work/app --silent`.
4. From `/work/app-copy`, run `ix locate <unique-function> --kind function`.

## Expected behavior

`/work/app-copy` is not treated as a descendant of `/work/app`. Normal unmatched-workspace fallback behavior applies, so the unique function from `/work/app` is not returned.

## Actual behavior

The function from `/work/app` is returned while the current directory is `/work/app-copy`.

## CLI output

The sibling directory reports the mapped workspace's graph and resolves its unique symbol.

## Environment

- OS: macOS 15.6, arm64
- Node: 26.7.0
- Ix version: 0.9.3

The cause is `findWorkspaceForCwd` using `cwd.startsWith(root_path)` without checking a path-component boundary.

Hiro-Chiba:
Fixes #402

## Summary

Prevent workspace lookup from treating an unrelated sibling directory whose name shares a prefix as part of a registered workspace.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes

- Match workspace roots using path-component boundaries instead of raw string prefixes.
- Preserve nearest-workspace selection for nested workspaces.
- Add regression coverage for exact roots, descendants, sibling prefixes, and nested roots.

## Validation

- `npm test` (61 files, 843 tests, including parser smoke tests)
- `npm run typecheck`
- `npm run build`
- Targeted ESLint and Prettier checks
- Live synthetic reproduction with a distinct default workspace. A sibling path no longer resolves a symbol mapped only from the prefix-sharing workspace.

## Checklist

- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/config-workspace-boundary.test.ts
+++ b/ix-cli/src/cli/__tests__/config-workspace-boundary.test.ts
@@ -1,0 +1,35 @@
+import { describe, expect, it } from "vitest";
+
+import { selectWorkspaceForCwd, type WorkspaceConfig } from "../config.js";
+
+function workspace(rootPath: string, name = rootPath): WorkspaceConfig {
+  return {
+    workspace_id: name,
+    workspace_name: name,
+    root_path: rootPath,
+    default: false,
+  };
+}
+
+describe("workspace path matching", () => {
+  it("matches a workspace root and its descendants", () => {
+    const candidate = workspace("/work/app", "app");
+
+    expect(selectWorkspaceForCwd([candidate], "/work/app")).toBe(candidate);
+    expect(selectWorkspaceForCwd([candidate], "/work/app/src/features")).toBe(candidate);
+  });
+
+  it("does not match a sibling whose name only shares the root prefix", () => {
+    const candidate = workspace("/work/app", "app");
+
+    expect(selectWorkspaceForCwd([candidate], "/work/app-copy")).toBeUndefined();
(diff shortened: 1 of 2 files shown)


josephismikhail (approved):
Reviewed path-boundary semantics and callers; no blocking findings. Independently verified exact roots, descendants, sibling prefixes, nested roots, Windows case/trailing-separator/different-drive behavior, and fallback preservation. Full ix-cli validation on the PR head: 843 tests (840 passed, 3 skipped), parser smoke passed, typecheck clean, lint 0 errors (pre-existing warnings only).

Issue #103, opened 2026-03-29
[IX Error][ix] ix-inventory: inventory failed


KageBinary:
## Summary

Auto-reported by ix-claude-plugin.

## Context

- **Component:** `ix-inventory`
- **Type:** ix
- **Action:** `ix inventory '~/ix/ix-website'`

## Error

- **Message:** inventory failed
- **Exit code:** `1`

### Stderr

```
error: required option '--kind <kind>' not specified
```

## Metadata

- Plugin version: `2.3.0`
- Timestamp: 2026-03-29T20:47:17Z
- Retry: yes

---
> Sensitive data was redacted automatically.
>

KageBinary:
**Recurrence** — 2026-04-09T03:03:35Z

- Exit: `1`
- Action: `ix inventory 'src'`
### Stderr

```
error: required option '--kind <kind>' not specified
```

PR #404, opened 2026-08-12
fix(cli): drop the dead `plans` and `goals` Pro stubs

josephismikhail:
## Summary
Companion to **Ix-pro#109** (merged as `da55e53`), which collapsed `ix plans` into `ix plan list`. `PRO_COMMANDS` still stubbed the plural, so an OSS-only install answered *"The 'plans' command requires Ix Pro."* — an upsell for a command installing Pro would not provide.

It turned out to be **two** stale stubs, not one. `goals` has been dead since Ix-pro#103: Ix#327 correctly dropped the stub, and **Ix#384 re-added it** on the stated premise that Pro registers a plural for every singular — which was already false. That premise was encoded in the paired-rule test, which is why it came back.

Verified against Pro `main` (`da55e53`): neither `registerPlansCommand` nor `registerGoalsCommand` is registered. `tasks` is now the only plural Pro actually registers.

## Changes
- `register/oss.ts` — drop both stubs. The singular stubs still cover the subcommand form; they swallow operands, so `plan list` reaches the sentinel.
- `commands/workflows.ts` — add `plans` to `COLLAPSED_HELP_TOPICS`. Per the lesson from #396, dropping a stub breaks `ix help <plural>`, which resolves topics against `program.commands`.
- `pro-stub-message.test.ts` — rewrite the paired-rule block so it no longer *asserts* the false premise. It now carries only `task`/`tasks`, plus a collapsed-subcommand table for `goal`/`plan` that pins both halves: the singular must reach the sentinel, the plural must not answer at all. The comment says to verify registration before adding a row rather than inferring one from the singular.
- `help-topics.test.ts` — add the `plans` → `plan` forwarding row.
- `CLAUDE.md`, `skills/ix/references/output-formats.md` — stop naming `ix plans`.

## Restore-the-bug
Each guard verified by reintroducing what it exists to catch:

| restored bug | result |
|---|---|
| re-add the `plans` stub | red — `stubs the collapsed 'plan list' subcommand, not a 'plans' command` |
| re-add the `goals` stub | red — `stubs the collapsed 'goal list' subcommand, not a 'goals' command` |
| drop the `plans` help forwarder | red — `ix help plans forwards to the plan command` |

## Test plan
- [x] `pro-stub-message.test.ts` + `help-topics.test.ts` — 30/30 green
- [x] Full `ix-cli` suite — 795 passed. Two files fail at *collection* on a missing `@modelcontextprotocol/sdk` in my local `node_modules`; reproduces with this change reverted, so it is environmental, not from this PR.

## Note on ordering
Deliberately opened **after** Ix-pro#109 merged — dropping the stub first would have given OSS users "unknown command" for something Pro still had.

This one needs a review from someone else; the ruleset won't let me approve my own.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -106,7 +106,7 @@ accept `--format`, despite being an action command.)
 
 **Pro features.** Some commands below are marked **[Pro]**, as are the whole
 **Planning** and **Workflows** sections — every command in those two tables is
-Pro-only, including `plan`, `plans`, `task`, `tasks` and `workflow`. If any Pro
+Pro-only, including `plan`, `task`, `tasks` and `workflow`. If any Pro
 command prints `The '<name>' command requires Ix Pro.`, this install does not
 have them — skip that step, do not retry it, and do not mention it again for the
 rest of the session. Nothing outside those marks and those two sections is
@@ -193,7 +193,7 @@ Only the first four work without Pro.
 | Plan status | `ix plan status` | `ix plan status <id> --format json` |
 | Next actionable task | `ix plan next` | `ix plan next <id> --with-workflow --format json` |
 | Run next task workflow | `ix plan next` | `ix plan next <id> --run-workflow --stage discover --format json` |
-| List all plans | `ix plans` | `ix plans --format json` |
+| List all plans | `ix plan list` | `ix plan list --format json` |
 | List tasks | `ix tasks` | `ix tasks --status pending --plan <id> --format json` |
 | Task details | `ix task show` | `ix task show <id> --with-workflow --format json` |
 | Update task | `ix task update` | `ix task update <id> --status done --format json` |
--- a/ix-cli/src/cli/__tests__/help-topics.test.ts
+++ b/ix-cli/src/cli/__tests__/help-topics.test.ts
@@ -86,14 +86,17 @@ describe("collapsed plural help topics resolve", () => {
   it.each([
     ["bugs", "bug"],
     ["goals", "goal"],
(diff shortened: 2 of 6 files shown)


josephismikhail:
Closing — no longer needed, and deliberately not deferred.

This PR existed because Ix-pro#109 removed `ix plans` while `PRO_COMMANDS` kept stubbing it, so an OSS-only install would have upsold Pro for a command Pro no longer had. That removal is being reverted in Ix-pro#112: `ix plans` stays a live Pro command (alongside the new `ix plan list`). With the plural live again, the existing stub message *"The 'plans' command requires Ix Pro."* is true, so there is nothing to fix here.

**One real thing this PR found does survive, and it is not caused by #109.** The `goals` stub at `ix-cli/src/cli/register/oss.ts:46` is genuinely dead — Ix-pro removed the command in #103, #327 correctly dropped the stub, and #384 re-added it on the premise that Pro registers a plural for every singular, which was already false. That premise is still encoded in `pro-stub-message.test.ts`'s paired-rule table, which is why it came back once already. I verified against Pro `main`: `registerGoalsCommand` is not registered.

Leaving it as-is for now rather than reopening a scoped-down PR, since the ask was to keep the OSS repo untouched. Worth picking up separately — the fix is to drop the `goals` entry, add `goals` to `COLLAPSED_HELP_TOPICS` so `ix help goals` keeps resolving, and rewrite the paired-rule comment so the false premise stops regenerating the stub.

Branch left in place if anyone wants to revive it.

PR #406, opened 2026-08-13
fix: resolve relative imports by path

### LINKED ISSUE #405: [bug] Relative imports are dropped when duplicate file stems exist
## Description

Relative imports are resolved through a repository-wide basename/stem lookup. If an unrelated directory contains a file with the same name, the correct import becomes ambiguous and the resolver drops the corresponding `IMPORTS` and `CALLS` edges.

This affects JavaScript/TypeScript `./` and `../` imports and Python relative imports.

## Steps to reproduce

1. Create `a/caller.ts` that imports `target` from `./utils`.
2. Create `a/utils.ts` that exports `target` and build the resolution graph.
3. Add an unrelated `b/utils.ts` that also exports `target`.
4. Build the resolution graph again.

The same behavior can be reproduced with `from .utils import target` and duplicate `utils.py` files in Python.

## Expected behavior

The import path selects `a/utils.ts`, and the resolver emits `IMPORTS` and `CALLS` edges to that file regardless of unrelated same-named files.

## Actual behavior

The relative specifier is flattened to the module stem `utils`. Both files become candidates, the target is treated as ambiguous, and the valid edges from the caller are omitted.

## CLI output

```text
Before adding b/utils.ts: IMPORTS and CALLS resolve to a/utils.ts
After adding b/utils.ts:  no resolved IMPORTS or CALLS edge for the caller
```

## Environment

- OS: Darwin 25.6.0, arm64
- Node: v26.7.0
- Ix version: 0.9.3, main at 8b117da

Hiro-Chiba:
Fixes #405

## Summary
Resolve JavaScript, TypeScript, and Python relative imports against the importing file's path before using repository-wide name fallback.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- resolve `./` and `../` imports with TypeScript-compatible source priority and directory index support
- resolve Python relative modules, including `from . import module`, while preserving imported-symbol bindings
- prevent missing relative imports from connecting to unrelated same-named files
- add regression coverage for duplicate stems, parent imports, directory indexes, source priority, Python bindings, and cross-batch indexes

## Validation
- `cd core-ingestion && npx vitest run src/__tests__/queries.python.test.ts src/__tests__/resolveEdges.test.ts` (52 passed)
- `cd ix-cli && npm test` (60 files, 853 passed, 1 skipped; parser smoke passed)
- `cd ix-cli && npm run typecheck && npm run build && npm run lint` (0 lint errors; 41 existing warnings)
- `cd core-ingestion && npm test` (294 passed; the same 6 pre-existing failures reproduce on unmodified `origin/main`, including missing private Scala fixtures and stale expectations)

## Checklist
- [x] Tests pass for the changed paths; pre-existing core failures are documented above
- [x] Smoke tests pass

### DIFF
--- a/core-ingestion/src/__tests__/queries.python.test.ts
+++ b/core-ingestion/src/__tests__/queries.python.test.ts
@@ -35,6 +35,11 @@ def top_level():
       dstName: 'User',
       predicate: 'IMPORTS',
     });
+    expect(result!.importBindings).toContainEqual({
+      pkg: '.models',
+      local: 'User',
+      imported: 'User',
+    });
     expect(result!.relationships).toContainEqual({
       srcName: 'example.py',
       dstName: 'package.helpers',
@@ -63,6 +68,25 @@ def top_level():
       srcName: 'Service.run',
       dstName: 'value.save',
       predicate: 'CALLS',
+    });
+  });
+
+  it('preserves the module path for from-dot imports', () => {
+    const result = parseFile(
+      '/repo/pkg/caller.py',
+      'from . import utils\n\ndef caller():\n    return utils.target()\n',
+    );
+
+    expect(result!.relationships).toContainEqual({
+      srcName: 'caller.py',
+      dstName: 'utils',
+      predicate: 'IMPORTS',
+      importRaw: '.utils',
+    });
+    expect(result!.importBindings).toContainEqual({
+      pkg: '.utils',
+      local: 'utils',
+      imported: 'utils',
     });
   });
 
--- a/core-ingestion/src/__tests__/queries.python.test.ts
+++ b/core-ingestion/src/__tests__/queries.python.test.ts
@@ -90,6 +90,19 @@ def top_level():
     });
   });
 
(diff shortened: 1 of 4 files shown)


josephismikhail (approved):
Reviewed the relative-resolution implementation and reproduced four edge cases before merge. I added fixes and regression coverage in 2b07d81 for:

- suppressing relative self-import file edges
- case-insensitive path matching on Windows
- JS runtime-file priority when both `.js` and `.ts` exist
- query/hash suffixed module specifiers and Python `as` bindings

Validated on the updated branch with 60 focused/property tests, the 854-test CLI suite plus parser smoke, builds, typecheck, and lint (0 errors). The full core suite has the same 39 environment/baseline failures on the untouched base and this branch.

josephismikhail (approved):
Follow-up security review complete: replaced the flagged query/hash regex with bounded indexOf/slice parsing in 7d2ddbb. Focused/property tests (60) and the ingestion build still pass.

PR #408, opened 2026-08-13
fix: resolve local export aliases

### LINKED ISSUE #407: [bug] Same-repo default and aliased exports drop resolved edges
## Description

Within a single repository, JavaScript and TypeScript imports do not resolve through provider-side default or aliased export names. The parser already records most consumer import bindings and provider public names, but `resolveEdges` only performs a by-name lookup against the provider's local entity names.

As a result, valid `CALLS`, `EXTENDS`, and `REFERENCES` edges are omitted. The common `export default LocalName` form is also not recorded as a provider public name.

## Steps to reproduce

1. Create `provider.ts` with `function impl() {}` and `export { impl as publicFn }`.
2. Create `consumer.ts` with `import { publicFn } from "./provider"` and call `publicFn()`.
3. Parse both files and run `resolveEdges`.
4. Repeat with `function actualName() {}` / `export default actualName` and `import localName from "./provider"`.

## Expected behavior

The consumer call resolves to the provider's local entity (`impl` or `actualName`) while preserving the consumer-side name on the emitted edge. The same mapping should apply when the imported symbol is used in inheritance or a type reference.

## Actual behavior

No resolved edge is emitted for either fixture. In the identifier default-export form, the parser also omits the `default` to local-name mapping.

## CLI output

```text
aliasedExportCallEdges=0
defaultExportCallEdges=0
```

## Environment

- OS: Darwin 25.6.0, arm64
- Node: v26.7.0
- Ix version: 0.9.3, main at 8b117da

Hiro-Chiba:
Fixes #407

## Summary
Resolve same-repository JavaScript and TypeScript imports through provider-side default and aliased export names.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- preserve `export default LocalName` as a provider public-name mapping
- index provider public names for cross-batch resolution
- resolve `CALLS`, `EXTENDS`, and `REFERENCES` from consumer import bindings to provider local entities
- keep local shadows and ordinary same-file calls ahead of imported provider symbols
- avoid using public names to choose between ambiguous module candidates

## Validation
- `cd core-ingestion && npx vitest run src/__tests__/resolveEdges.test.ts` (44 passed)
- `cd ix-cli && npm test` (60 files, 853 passed, 1 skipped; parser smoke passed)
- `cd ix-cli && npm run typecheck && npm run build && npm run lint` (0 lint errors; 41 existing warnings)
- `cd core-ingestion && npm test` (290 passed; the same 6 pre-existing failures reproduce on unmodified `origin/main`, including missing private Scala fixtures and stale expectations)

## Checklist
- [x] Tests pass for the changed paths; pre-existing core failures are documented above
- [x] Smoke tests pass

### DIFF
--- a/core-ingestion/src/__tests__/resolveEdges.test.ts
+++ b/core-ingestion/src/__tests__/resolveEdges.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from 'vitest';
 
-import { buildGlobalResolutionIndex, resolveEdges, type FileParseResult, type ParsedEntity, type ParsedRelationship } from '../index.js';
+import { buildGlobalResolutionIndex, parseFile, resolveEdges, type FileParseResult, type ParsedEntity, type ParsedRelationship } from '../index.js';
 import { SupportedLanguages } from '../languages.js';
 
 function entity(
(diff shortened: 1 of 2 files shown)


josephismikhail (approved):
Reviewed and reproduced the reported same-repo alias/default-export failures. Before merge I also found and fixed three interactions in 2526f0f:

- provider alias lookup bypassed #406 path narrowing, so duplicate relative provider stems still dropped valid edges
- nested-function locals were treated as shadows in outer scopes
- parameter shadows were missed, and TypeScript value/type namespaces were conflated

Added regression coverage for relative provider disambiguation, nested and parameter shadows, imported type annotations, and separate value/type shadowing. Validated with 74 focused/property tests, the 854-test CLI suite plus parser smoke, builds, typecheck, and lint (0 errors). The full core suite has the same 39 baseline/environment failures as current main.

PR #412, opened 2026-08-13
fix(view): keep backend scope out of generated script

josephismikhail:
## Summary

- Make the generated Compass server script invariant.
- Pass runtime scope and paths through process arguments instead of embedding backend-derived data in executable JavaScript.
- Use the current Node executable for the detached server.
- Wait for the detached server before reporting it ready.
- Centralize and test the runtime argument contract, including scope headers and paths with spaces.

Closes the network-data-to-file CodeQL finding in ix view (security/code-scanning/50).

## Validation

- npm test: 856 passed, 3 skipped
- npm run build
- npm run typecheck
- npm run lint -- --quiet
- npm run knip
- npm audit --audit-level=high: 0 vulnerabilities
- bash scripts/ci/check-config-security.sh

### DIFF
--- a/ix-cli/src/cli/__tests__/view-running-port.test.ts
+++ b/ix-cli/src/cli/__tests__/view-running-port.test.ts
@@ -41,9 +41,20 @@ beforeEach(() => {
   writeFileSync(join(compassDir, "index.html"), "<!doctype html>");
 
   mocks.spawn.mockReturnValue({ pid: 4242, unref: vi.fn() });
+  // The same probe answers two opposite questions: before the spawn, "is this
+  // port already taken?" (it must not be), and after it, "is the server up
+  // yet?" (it must be). A mock that always refuses satisfies the first and
+  // hangs the second for the whole readiness budget, so model the transition:
+  // refuse until the server is spawned, accept once it has been.
   mocks.createConnection.mockImplementation(() => {
-    const connection = new EventEmitter();
-    queueMicrotask(() => connection.emit("error", new Error("not listening")));
+    // `end()` because the connect path closes the socket it just opened.
+    const connection = Object.assign(new EventEmitter(), { end: vi.fn() });
+    const listening = mocks.spawn.mock.calls.length > 0;
+    queueMicrotask(() =>
+      listening
+        ? connection.emit("connect")
+        : connection.emit("error", new Error("not listening")),
+    );
     return connection;
   });
 });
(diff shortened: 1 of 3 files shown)


KageBinary:
Reviewed and pushed one commit on top (3d67995).

The change itself is right — I checked the argv contract end to end and `argv[2..7]` line up with the spawn array in both the command and the test, empty strings are passed positionally so nothing shifts, and `|| null` preserves the old `JSON.stringify(null)` semantics. No `shell: true`, so paths with spaces are still fine. I also confirmed it keeps the `MAP_ROOT` guard intact — remap still refuses when it is null and pins `cwd` to it — so the "detached server inherits the launch cwd" hazard stays closed.

**What I added.** The new guard

```js
if (!DIST || !Number.isInteger(PORT) || ... || !MAP_MAIN) {
  throw new Error("invalid ix view server arguments");
}
```

throws inside a process spawned `detached: true, stdio: "ignore"`, and the CLI only checked `child.pid` before printing `[ok] Visualizer started`, writing a PID file and opening a browser. So a server that died on launch reported success and opened a tab on a dead port. That failure mode is new — before this PR there was no argv contract to get wrong.

`view-server.test.ts` already has `waitForPort` for precisely this reason, and even says so: *"a server that dies on startup would otherwise surface as a timeout in waitForPort with no indication of why ... invisibly, because stdio is ignored."* My commit gives the command the same check.

Deliberately **non-fatal**: 8s is a heuristic, so on a slow machine killing the server or failing the command would be worse than the mis-report. The PID/scope/port files are written either way, so `status`/`stop` still work on whatever did start.

One knock-on: `view-running-port.test.ts` mocked `createConnection` to always refuse. That answers the pre-spawn "is this port already taken?" check but hangs a post-spawn readiness probe for its full budget, so the mock now models the transition (refuse before spawn, accept after).

Local: 856 tests pass, typecheck clean, lint 0 errors (41 warnings, the existing baseline), knip exit 0 — same as without the change.

One thing not addressed, flagging rather than changing: argv is visible in the process table, so `workspaceId`, `systemId` and the paths are now readable by other local users via `ps`, where before they sat in a file. Identifiers and paths rather than secrets, so I don't think it blocks this — but it is a change in exposure.

KageBinary (approved):
Reviewed the argv contract end to end and ran the full suite locally: 856 tests, typecheck clean, lint 0 errors, knip 0. Approving the security fix; note I pushed 3d67995 on top (readiness check), so a second pair of eyes on that commit specifically is welcome.

PR #414, opened 2026-08-13
fix(config): create Ix home before saving config

### LINKED ISSUE #413: [bug] ix config set fails when ~/.ix does not exist
## Description

On a fresh home directory, `ix config set` fails before writing its first configuration. `saveConfig()` writes an atomic staging file beside `~/.ix/config.yaml`, but assumes the parent `~/.ix` directory already exists.

## Steps to reproduce

1. Create an empty temporary home directory.
2. Run `HOME=<temp> USERPROFILE=<temp> ix config set format json`.

## Expected behavior

Ix creates `~/.ix/config.yaml`, stores the requested value, and keeps the file private (`0600` on POSIX).

## Actual behavior

The command fails with `ENOENT` while opening `~/.ix/config.yaml.<pid>.tmp`, and no config is written.

## CLI output

```text
Error: ENOENT: no such file or directory, open '<temp>/.ix/config.yaml.<pid>.tmp'
```

## Environment

- OS: macOS 26.6.1, arm64
- Node: v26.7.0
- Ix: 0.9.3, main at `ab823e3`

Hiro-Chiba:
Fixes #413

## Summary
Create the Ix home directory when `saveConfig()` performs the first config write, while preserving the existing private atomic replacement path.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Create `~/.ix` recursively before writing the same-directory staging file.
- Keep the existing `0600` staging mode, atomic rename, and post-rename permission tightening.
- Cover a fresh home, file permissions, atomic replacement, staging cleanup, and preservation of extension and user-owned fields.

## Validation
- `npx vitest run src/cli/__tests__/config-file-mode.test.ts` (5 passed)
- `npm test` (61 files; 861 passed, 1 skipped; parser smoke passed)
- `npm run typecheck`
- `npm run build`
- Fresh-home CLI probe: `ix config set format json` wrote a `0600` config with no staging file left behind
- `git diff --check`
- `npm run format:check` still reports the same 123 pre-existing files as unformatted on both `origin/main` and this branch

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/config-file-mode.test.ts
+++ b/ix-cli/src/cli/__tests__/config-file-mode.test.ts
@@ -44,6 +44,13 @@ describe("saveConfig persistence", () => {
     expect(fs.readdirSync(cfgDir())).toEqual(["config.yaml"]);
   });
 
+  it.skipIf(!posix)("creates the config directory 0700", () => {
+    // The 0600 on the file is undone by a 0755 directory beside it: the names
+    // in ~/.ix are themselves a disclosure, and this is the call that creates it.
+    saveConfig({ endpoint: "http://localhost:8090", format: "text" });
+    expect(fs.statSync(cfgDir()).mode & 0o777).toBe(0o700);
+  });
+
   it.skipIf(!posix)("creates the config 0600", () => {
     saveConfig({ endpoint: "http://localhost:8090", format: "text" });
     expect(fs.statSync(cfgPath()).mode & 0o777).toBe(0o600);
--- a/ix-cli/src/cli/__tests__/config-file-mode.test.ts
+++ b/ix-cli/src/cli/__tests__/config-file-mode.test.ts
@@ -2,6 +2,7 @@ import { afterEach, beforeEach, describe, expect, it } from "vitest";
 import * as fs from "node:fs";
 import * as os from "node:os";
 import * as nodePath from "node:path";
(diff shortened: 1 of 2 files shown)


KageBinary:
Thanks — this is a clean fix, and deleting the `mkdirSync` from `beforeEach` is the part that makes it real. That line was masking the bug, so the new fresh-home case actually tests something.

I pushed one hardening commit on top (`3a24e8c`): `mkdirSync(configDir, { recursive: true, mode: 0o700 })`.

The file is deliberately written `0600` because it holds credentials — Pro's `instances` carry a tunnel JWT and a long-lived IdP refresh token. A directory created at the default umask (typically 0755) undoes part of that, since anyone on the host can list what sits beside it. `mode` applies only to directories the call creates, so an existing `~/.ix` keeps whatever mode it already had.

Added a test for the directory mode alongside your `0600` one, under the same `posix` guard.

Approving once CI is green.

KageBinary (approved):
CI green on `3a24e8c`. Small, correct, and the test change is the part that matters — removing the `mkdirSync` from `beforeEach` is what turns the new fresh-home case into a real regression test. Directory mode hardened to 0700 on top. Ready to merge whenever @ianhock wants it.

PR #421, opened 2026-08-13
fix(watch): refresh through canonical map ingestion

### LINKED ISSUE #420: [bug] ix watch bypasses canonical ingestion and leaves stale graph data
## Description

`ix watch` commits per-file patches directly instead of using the canonical ingestion pipeline. Watched changes can therefore leave removed entities in the graph, omit cross-file resolved edges, ignore deleted files, and leave the architectural map stale even though watch reports the file as ingested.

## Steps to reproduce

1. Create `provider.ts` with an exported `target` function.
2. Create `consumer.ts` with a `caller` function that imports and calls `target`.
3. Run `ix map . --silent`, then start `ix watch`.
4. Rename `caller` to `callerTwo` and wait for watch to report the ingest.
5. Run `ix callers target`, `ix locate caller`, and `ix locate callerTwo`.
6. Delete `consumer.ts` and query the graph again.

## Expected behavior

Watch converges to the same graph as canonical ingestion followed by map. The old caller is removed, the new caller has a resolved edge to `target`, deleted-file entities disappear, and map data is refreshed.

## Actual behavior

The old caller remains, the new caller has no resolved call edge, and deleting the source file produces no watch update and leaves its entity in the graph. The newly committed entity can also report no map data.

## CLI output

```text
[watch] changed: src/consumer.ts
[watch] ingested: src/consumer.ts → rev 2

ix callers target     -> caller
ix locate caller      -> old entity still resolves
ix locate callerTwo   -> new entity resolves with stale=true and hasMapData=false
```

## Environment

- OS: macOS 26.6.1, arm64
- Node: v26.7.0
- Ix: 0.9.3, main at `ab823e3`

Hiro-Chiba:
Fixes #420

## Summary
Refresh watched workspaces through the canonical map ingestion path so watch produces the same graph data as a normal full ingest and map.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Replace watch's direct per-file patch commits with serialized canonical workspace refreshes.
- Preserve full entity, edge, chunk, claim, replacement, deletion, baseline, and region updates behind the existing map lock.
- Coalesce changes that arrive during a refresh and retry lock contention without losing the trailing update.
- Detect deletions in both recursive watch and polling modes while keeping `--path` limited to the watched event scope.
- Preserve migration baseline handling and TypeScript runtime loaders for the development CLI without forwarding debugger flags.

## Validation
- `npx vitest run src/cli/__tests__/watch-dedup.test.ts src/cli/__tests__/map-auto-skip.test.ts` (28 passed)
- `npm test` (61 files; 874 passed, 1 skipped; parser smoke passed)
- `npm run typecheck`
- `npm run build`
- Isolated-backend E2E: renaming `caller` removed the old entity and produced the new cross-file call edge; deleting the source removed the caller entity and edge
- `ix map . --silent` against an isolated backend (217 files)
- `git diff --check`
- `npm run format:check` still reports repository-baseline formatting failures; the changed watch test passes targeted Prettier checking

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/map-auto-skip.test.ts
+++ b/ix-cli/src/cli/__tests__/map-auto-skip.test.ts
@@ -1,5 +1,17 @@
-import { afterEach, describe, expect, it } from 'vitest';
-import { shouldSkipAutoMap } from '../commands/map.js';
+import * as fs from 'node:fs';
+import * as os from 'node:os';
+import * as path from 'node:path';
+import { hostname } from 'node:os';
+import { Command } from 'commander';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import {
+  applyRequestedMapCoalesceExitCode,
+  mapModeForIngest,
+  registerMapCommand,
+  requestedMapCoalesceExitCode,
+  shouldSkipAutoMap,
+} from '../commands/map.js';
+import { lockPathForTest } from '../single-flight.js';
 
 describe('shouldSkipAutoMap', () => {
   afterEach(() => { delete process.env.IX_AUTO_MAP_CLOUD; });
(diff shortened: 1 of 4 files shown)


KageBinary:
This is the right architectural call. Deleting the in-process ingest path — the one that built its own patches with `buildPatch` and could mint node ids that diverged from what `ix map` produced — removes a whole class of watch-vs-map drift rather than papering over it. The `childCliArgs` handling is careful too: copying loader flags but deliberately not `--inspect` is exactly right, and `spawn` with an argv array means there's no shell in the path.

I pushed one commit (`58dfc99`) for the retry loop.

**`runCanonicalMap` retried exit 75 in a `for (;;)` at a flat 1s with no ceiling.** Exit 75 normally clears in seconds, but a lock whose holder died is only stolen once it passes `DEFAULT_LOCK_MAX_MS` (20 min, `single-flight.ts`). Until then this spawned a Node process every second — about 1,200 of them, and silently, because the child runs `--silent`. Exponential backoff capped at 30s reaches the same 20 minutes in 45 attempts. The ceiling is set just past the stale-lock window on purpose: a genuinely long map is still waited out, and only a lock that outlives its own staleness window gives up. Exhaustion throws now, so the scheduler's error path reports it instead of the refresh quietly vanishing.

It also says why it's waiting once the wait stops looking momentary — retrying silently for minutes is indistinguishable from a dead watcher.

Your existing `[75, 0]` test still passes unchanged.

**One thing that is your call, not a defect.** Before this PR a changed file was parsed in-process and committed as one patch; now each debounced batch spawns a full `ix map` with `IX_MAP_FULL_INGEST=1`. The debounce and the coalescing scheduler soften it, but that's a real cost change on a large repo, and it arrives as a side effect of a correctness fix. Worth a deliberate yes from @ianhock rather than inheriting it.

Related: `mapModeForIngest` reads `IX_MAP_FULL_INGEST` from the ambient environment, so anyone who exports it for any reason turns *every* `ix map` into a full ingest. Same shape for `IX_MAP_COALESCE_EXIT_CODE` changing `ix map`'s exit code. Both are reasonable knobs; they should probably be documented rather than discovered.

PR #422, opened 2026-08-13
feat(mcp): enrich Ix MCP tools with structured outputs, semantic annotations, and stronger agent interoperability

Alot1z:
## Summary

Enriches Ix's MCP server (built on the official `@modelcontextprotocol/sdk`) with an agent-facing contract layer: every tool now advertises truthful semantic annotations, and the JSON-backed tools emit structured results instead of bare text, so agents receive machine-readable output they can act on without re-parsing.

This is an incremental enhancement of the existing MCP layer introduced in #397 and repaired in #400 / #401. It does not replace the current architecture: the stdio server, argv isolation (`--flag=value`, `--` separators), orphan bookkeeping, per-command scope-cache invalidation, single-flight lock ownership, output caps, and Pro-aware tool gating all remain in place.

## What this adds

**Tool annotations** (`readOnlyHint`, `destructiveHint`, `idempotentHint`, `openWorldHint`, `title`)

Every tool is classified against a type-checked lookup table keyed by `IX_MCP_TOOL_NAMES`, so TypeScript fails the build if a tool is ever added without a classification. Read-only graph queries carry `readOnlyHint: true` and `idempotentHint: true`; mutating tools (`ix_map`, `ix_ingest`, `ix_decide`) carry `destructiveHint: true` and `idempotentHint: false`; `ix_ingest` additionally reaches outside the host (GitHub), so it is marked `openWorldHint: true`. Every tool has a non-empty `title`. Annotations are hints — the hard security boundaries remain inside Ix's executor.

**Structured output** (`outputSchema` + `structuredContent`)

`ix_map`, `ix_ingest`, and `ix_smells` now declare an `outputSchema` and emit `structuredContent` alongside their text, so an agent can consume the JSON directly. The shapes are verified against each command's actual emitter (`map` and `ingest` already return JSON objects; `smells` returns `{ candidates, ... }`). Pro tools (`ix_briefing`, `ix_decisions`, `ix_decide`) live in the private `@ix/pro` package and are intentionally left unchanged here — their shapes cannot be verified from this repository, and advertising a schema we cannot prove would violate the SDK's runtime validation.

## Why annotations and structured content matter

- An agent deciding whether `ix_ingest` is safe to call can now read `destructiveHint` instead of guessing from the description.
- An agent that receives `ix_map`'s JSON can walk `entities`/`edges` directly instead of re-requesting and re-parsing.
- The SDK validates `structuredContent` against `outputSchema` at runtime, so a shape drift between Ix and its MCP contract becomes a loud failure rather than silent agent confusion.

## Tests

- `mcp-annotations.test.ts`: every catalog name resolves to a classified annotation set; read-only tools are idempotent; mutating tools are destructive and non-idempotent; every tool advertises a truthful non-empty title.
- `mcp-structured-output.test.ts`: the three structured tools declare a JSON `outputSchema`; calls against them return `structuredContent` matching the declared schema.
- Full suite: **865 tests pass**, typecheck clean, lint clean (0 errors).

## Historical context

Earlier MCP implementation and hardening work in the `Alot1z/Ix-remap` development fork:

- https://github.com/Alot1z/Ix-remap/commit/74b848c83a0d547069660615dddcf1ea0ad0749c
- https://github.com/Alot1z/Ix-remap/commit/0d99ae0f1866367c3fbc9bbcc16f0add2dd4dd57
- https://github.com/Alot1z/Ix-remap/commit/869b64df3357a8370c4c85c9a1fa2b553b899e24
- https://github.com/Alot1z/Ix-remap/commit/36c7c7eccd8068d48df4f61394b42a3ffa62483c
- https://github.com/Alot1z/Ix-remap/commit/1a5b0b93c9e8871610370c0f36212be8f6cf6980

For context on the existing upstream MCP implementation and follow-up hardening, see #397, #400, and #401.

## Reference / evidence

- Ix-findings knowledge site (evidence layer; not a runtime dependency):
  - MCP architecture: https://alot1z.github.io/Ix-findings/mcp
  - MCP implementation details: https://alot1z.github.io/Ix-findings/mcp/implementation
  - MCP security analysis: https://alot1z.github.io/Ix-findings/mcp/security
  - MCP test coverage: https://alot1z.github.io/Ix-findings/mcp/tests
  - This PR's evidence pages: https://alot1z.github.io/Ix-findings/prs/422/analysis
  - Upstream MCP lineage (#397): https://alot1z.github.io/Ix-findings/prs/397/conversation
  - Historical fork commits: https://alot1z.github.io/Ix-findings/entities/commit-74b848c83a0d547069660615dddcf1ea0ad0749c


KageBinary:
Good change overall — the totality of `TOOL_ANNOTATIONS` over `IX_MCP_TOOL_NAMES` is the best idea in here, and the note about record schemas being silently dropped from `tools/list` is the kind of thing that costs someone an afternoon to rediscover.

I pushed one commit (`b7079af`) on two annotations that claimed more than the code backs.

### `ix_read` was titled "Read graph-bounded source"

It isn't. `ix read` resolves an exact file path *before* it consults the graph, so the title describes only part of the command. Retitled to what it does.

Worth being explicit about why this one matters more than a wording nit: clients use `readOnlyHint` to decide whether a call needs the user's approval. `ix read` on `main` today will read **any absolute path on the host** — no workspace containment, no graph lookup — so marking it read-only quietly converts "an agent can ask to read that file" into "an agent reads that file".

That containment gap is pre-existing and **not something this PR introduced**; I've opened #424 to fix the primitive. This PR is fine to land alongside it. Flagging it here only because this is the change that raises the stakes on it.

### The Pro tools asserted hints this package can't check

`ix_briefing`, `ix_decisions` and `ix_decide` are implemented in `@ix/pro`. This PR already declines to give them an `outputSchema`, on the grounds that their shape "cannot be verified from this repository" — I think that reasoning applies with more force to the behavioral hints, because `readOnlyHint` is the one that removes a confirmation prompt rather than changing how output is formatted. The caution was being spent on the cosmetic annotation and withheld from the security-relevant one.

They now carry a title and nothing else, via an `UNVERIFIED(...)` helper. The table stays total, so a new tool still can't be added without a decision — the decision can now be "not ours to make". Updated the doc paragraph to say so, and added tests pinning both changes.

### Merge order with #423

These two conflict in `src/mcp/server.ts`, and resolving the text isn't enough: #423 adds `ix_context` to the catalog, so after a merge `tsc` gives

```
error TS2741: Property 'ix_context' is missing in type … but required in type Record<…>
```

which is precisely your guard doing its job. Suggest this one lands first and #423 rebases on top with an `ix_context` entry.

KageBinary:
Second pass found one more, fixed in `6313fbe`.

### `structuredContent` was attached to tools with no `outputSchema`

`withStructuredContent` sits inside `runJson`, and `runJson` serves **five** tools, not two — `ix_map`, `ix_ingest`, and the three Pro tools. So `ix_briefing` and `ix_decisions` were emitting `structuredContent` with nothing describing it. Verified by calling them and inspecting the results:

```
ix_map         declaresOutputSchema=true  emitsStructuredContent=true
ix_ingest      declaresOutputSchema=true  emitsStructuredContent=true
ix_briefing    declaresOutputSchema=false emitsStructuredContent=true   <-- 
ix_decisions   declaresOutputSchema=false emitsStructuredContent=true   <-- 
```

This is the same shape as the annotation issue. The Pro tools are left out of `TOOL_OUTPUT_SCHEMA` because their output shape can't be checked from this package — but the code then handed clients the parsed object regardless, which gives them exactly the unverified value the omission was meant to withhold, minus the contract that would let them validate it. The schema and the structured content have to travel together.

Gated on `TOOL_OUTPUT_SCHEMA` membership. Their answer is unchanged and still returned as text.

The doc also said "the remaining tools return their `llm`-format text" — not true of the Pro tools, which run in JSON. Corrected to describe what actually happens.

New test fails if the gate is removed. Full suite 870 pass, typecheck and lint clean.

PR #423, opened 2026-08-13
feat(agent): unify Ix context, provenance, and investigation workflows for agents

Alot1z:
## Summary

Adds a bounded, deterministic, provenance-aware agent context layer to Ix by composing the graph intelligence the project already has — the resolver, the fact collector, the `/v1/context` service, and the provenance API — into one coherent bundle. It stops agents from stitching together disconnected graph results by hand.

This is **not** a second context engine. It reuses the existing relevance, provenance, freshness, and claim/decision/conflict/intent primitives and adds only the bundling, budgeting, deterministic ranking, and investigation state on top.

## What this adds

**`ix context [target]`** — a deterministic context bundle for a symbol, file, or entity:

- `target`, `entities`, `relationships` from the same resolver/collector/context service as `ix explain`
- `claims`, `decisions`, `conflicts`, `intents` from `/v1/context`
- `provenance` (source URI, hash, extractor, source type, observed revision, history length) from the provenance API
- explicit `freshness` classification (`current` / `stale`) from collected facts — never presents revision-specific evidence as current
- budgets (`--max-entities`, `--max-relationships`, `--max-evidence`, `--max-chars`) with **explicit truncation metadata**, never silent truncation
- deterministic evidence ranking (stable tier: target → structural → claim → decision → conflict → intent → relationship → provenance, with an id tiebreaker), so identical input yields identical output apart from the declared `generatedAt`
- claims/decisions/conflicts/intents are ordered deterministically, independent of incidental service array order
- formats: `text`, `json`, `llm`, and `--out <path>` for portable JSON export
- `--as-of-rev <n>` for historical context

**Resumable investigation state** — `ix context <target> --save <id>`, `ix context --resume <id>`, `ix context --diff <id>`:

- saves a bundle as a named investigation under `~/.ix/investigations` (versioned schema `ix-investigation/1`, injective id encoding so distinct ids can never collide or escape the directory)
- resumes without a backend — recovers target, explored entities, evidence, claims, decisions, conflicts, freshness, and the full context generation
- diffs a saved investigation against a fresh build and reports added/removed entities, relationships, evidence, and claims plus the previous→current freshness classification; the saved revision and depth are preserved for the diff unless explicitly overridden
- refuses malformed or unknown-schema files rather than guessing
- both persistence paths validate the bundle against the versioned `ix-context-bundle/1` schema before writing and use an atomic temp+rename write, so malformed payloads are refused and a partial file is never visible

**`ix_context` MCP tool** — the bundle is exposed over MCP with bounded budget parameters (including `max_chars`) and optional historical `--as-of-rev`, wired through the same argv isolation and JSON formatting as the other JSON-backed tools. It returns the bundle as `structuredContent` against a stable `outputSchema` shared with the CLI, so the MCP surface and persisted state validate the same contract. OSS/Pro catalog gating is unchanged.

## Design notes

- The only new scoring is the deterministic tier ordering; the underlying relevance/confidence comes from the existing context service.
- Every reduction is counted and reported in `truncation` — the bundle never silently drops content.
- Investigation state is a file under `~/.ix`, not chat history.

## Tests

- `context.test.ts`: determinism for identical input (timestamp excluded), tier ordering with stable tiebreak, budget enforcement with explicit truncation counts, staleness classification, injective id encoding.
- `context-investigation.test.ts`: save/resume round-trip under a temp `IX_HOME`, refusal of missing/malformed state and malformed bundles, deterministic saved→fresh deltas, freshness changes surfacing in the diff, revision/depth preservation, hostile-id collision safety.
- `mcp.test.ts`: `ix_context` forwards target + budgets as the CLI contract, omits unset budgets, returns `structuredContent` for parseable output, and marks unparseable output as an MCP error.
- Full suite: **875 tests pass**, typecheck clean, lint clean (0 errors). CodeQL (javascript-typescript) passes on the PR head with no findings.

## Historical context

Earlier MCP implementation and hardening work in the `Alot1z/Ix-remap` development fork:

- https://github.com/Alot1z/Ix-remap/commit/74b848c83a0d547069660615dddcf1ea0ad0749c
- https://github.com/Alot1z/Ix-remap/commit/0d99ae0f1866367c3fbc9bbcc16f0add2dd4dd57
- https://github.com/Alot1z/Ix-remap/commit/869b64df3357a8370c4c85c9a1fa2b553b899e24
- https://github.com/Alot1z/Ix-remap/commit/36c7c7eccd8068d48df4f61394b42a3ffa62483c
- https://github.com/Alot1z/Ix-remap/commit/1a5b0b93c9e8871610370c0f36212be8f6cf6980

For context on the existing upstream MCP implementation and follow-up hardening, see #397, #400, and #401.

## Reference / evidence

- Ix-findings knowledge site (evidence layer; not a runtime dependency):
  - MCP implementation details: https://alot1z.github.io/Ix-findings/mcp/implementation
  - MCP security analysis: https://alot1z.github.io/Ix-findings/mcp/security
  - MCP test coverage: https://alot1z.github.io/Ix-findings/mcp/tests
  - This PR's evidence pages: https://alot1z.github.io/Ix-findings/prs/423/analysis
  - Upstream MCP lineage (#397): https://alot1z.github.io/Ix-findings/prs/397/conversation
  - Historical fork commits: https://alot1z.github.io/Ix-findings/entities/commit-74b848c83a0d547069660615dddcf1ea0ad0749c

### DIFF
--- a/ix-cli/src/cli/__tests__/context-investigation.test.ts
+++ b/ix-cli/src/cli/__tests__/context-investigation.test.ts
@@ -25,8 +25,14 @@ let home: string;
 beforeEach(() => {
   home = mkdtempSync(join(tmpdir(), "ix-investigation-test-"));
   process.env.IX_HOME = home;
-  mkdirSync(join(home, ".ix", "investigations"), { recursive: true });
+  // IX_HOME is the Ix home directory; investigations live in a subdirectory of
+  // it, which saveInvestigation creates. Nothing is pre-created here — the tests
+  // below assert on where the code actually writes, not on a directory the test
+  // made itself.
 });
+
+/** Where saved investigations are expected to live, given IX_HOME. */
+const investigationsDir = () => join(home, "investigations");
 
 afterEach(() => {
   delete process.env.IX_HOME;
@@ -126,9 +132,27 @@ describe("ix context investigation state", () => {
     expect(loaded?.bundle.evidence).toEqual(bundle.evidence);
   });
 
+  it("writes into an investigations subdirectory of IX_HOME, not its root", () => {
+    // IX_HOME is the Ix home itself — it holds config.yaml, bin/, cli/ and
+    // dotfiles like .version-check.json. Saved state belongs in a subdirectory
+    // of it, not loose among them.
+    writeFileSync(join(home, "config.yaml"), "endpoint: http://localhost:8090\n");
+    writeFileSync(join(home, ".version-check.json"), JSON.stringify({ latest: "0.9.3" }));
+
+    saveInvestigation("widget-check", bundleWith([makeClaim("renders to DOM", 0.9)]));
+
+    expect(readdirSync(investigationsDir())).toEqual(["widget-check.json"]);
+    expect(readdirSync(home).sort()).toEqual([".version-check.json", "config.yaml", "investigations"]);
+  });
+
   it("refuses to resume a missing or malformed investigation", () => {
     expect(loadInvestigation("does-not-exist")).toBeUndefined();
-    writeFileSync(join(home, ".ix", "investigations", "broken.json"), "not json", "utf8");
+    // The fixture has to sit where the loader actually looks, or this passes
+    // because the file is absent rather than because it is malformed — and
+    // would keep passing with the JSON guard deleted.
+    mkdirSync(investigationsDir(), { recursive: true });
+    writeFileSync(join(investigationsDir(), "broken.json"), "not json", "utf8");
+    expect(existsSync(join(investigationsDir(), "broken.json"))).toBe(true);
     expect(loadInvestigation("broken")).toBeUndefined();
   });
 
@@ -136,7 +160,7 @@ describe("ix context investigation state", () => {
     const malformed = { ...bundleWith([]), entities: "not-an-array" } as unknown as ReturnType<typeof buildBundle>;
     saveInvestigation("bad-shape", malformed);
     expect(loadInvestigation("bad-shape")).toBeUndefined();
-    expect(existsSync(join(home, ".ix", "investigations", "bad-shape.json"))).toBe(false);
+    expect(existsSync(join(investigationsDir(), "bad-shape.json"))).toBe(false);
   });
 
   it("computes a deterministic delta between saved and fresh state", () => {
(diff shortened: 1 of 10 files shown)


KageBinary:
Thanks for this — the injective `sanitizeId` encoding, the atomic temp+rename writes, and the exact-serialized-size evidence budget are all genuinely well done, and the CodeQL rule IDs in the commit messages are a nice touch.

I pushed one commit (`49ff98b`) fixing a directory bug and the tests around it.

### `investigationDir()` put the subdirectory in the wrong branch

```js
process.env.IX_HOME || join(homedir(), ".ix", "investigations")
```

`IX_HOME` is the Ix home **directory** — `backend-status.ts:22`, `docker.ts:8` and `upgrade.ts:18` all read it as `IX_HOME || ~/.ix` and then join their own subdirectory onto it. With the subdirectory only in the fallback, the two branches disagree: set `IX_HOME` and saved investigations land loose in the Ix home beside `config.yaml`, `bin/` and `cli/`, and `investigations/` is never created at all.

Verified by running `saveInvestigation` with `IX_HOME` pointed at a populated home:

```
[".version-check.json", "bin", "cli", "config.yaml", "widget.json"]
```

That also lets a chosen `--save` id collide with real Ix state. `sanitizeId` passes `[A-Za-z0-9._-]` through untouched, including leading dots, so `--save .version-check` overwrote `$IX_HOME/.version-check.json` — the upgrade version cache from `upgrade.ts:19`. Reading it back afterwards gave `{"schema":"ix-investigation/1", …}`.

Fixing the directory removes the collision on its own, but `sanitizeId` now also encodes a leading dot so no id can produce a dotfile. Encoding it only in first position keeps the mapping injective.

### The tests encoded the bug rather than catching it

`beforeEach` created `$IX_HOME/.ix/investigations/` and the assertions then read `readdirSync(home)` — two facts only consistent with writing to the root.

The malformed-JSON case was worse than merely wrong. Its fixture went to `.ix/investigations/broken.json`, a path the loader never reads, so it passed because the file was **absent**, not because it was malformed — and would have kept passing with the JSON guard deleted entirely.

Fixtures now go where the code looks. I checked the corrected assertions actually bite: reverting `investigationDir` to the original line fails 2 of the 8 cases.

### Three `ix_context` MCP tests were green on failed calls

`ix_context` declares `contextBundleSchema` as its `outputSchema`, and the SDK validates `structuredContent` against it (`validateToolOutput`, `server/mcp.js`). Stubbing `{"schema": "ix-context-bundle/1"}` produces:

```
isError: true
"Output validation error: … Required at generatedAt, Required at target, …"
```

Those tests asserted only on the argv array, never on `isError`, so they went green on a tool call that had failed. They now stub a whole bundle via a shared `contextBundle()` helper and assert the call succeeded, plus one new case pinning that a short bundle really is surfaced as an error. Real bundles carry all fifteen fields and validate fine — this was a test-quality problem, not a runtime break.

### Merge order with #422

Both PRs rewrite `registerTool` in `src/mcp/server.ts`, and they conflict. Resolving the text isn't enough — `#422` makes `TOOL_ANNOTATIONS` total over `IX_MCP_TOOL_NAMES` specifically so a new tool can't be added unclassified, and this PR adds `ix_context` to that catalog:

```
error TS2741: Property 'ix_context' is missing in type … but required in type Record<…>
```

That's #422's guard working as designed. Suggest landing #422 first, then rebasing this and adding an `ix_context` entry — read-only, non-destructive, so it's an uncontroversial one.

KageBinary:
Second pass, this time on the bundle logic rather than the persistence paths. Found one correctness bug, fixed in `06471f6`.

### Every entity got the target's staleness

```js
const stale = facts.stale;          // the TARGET's staleness
...
entities.push({ id: node.id, ..., stale });   // stamped onto all of them
```

That's correct for the bundle-level `freshness`, which *is* about the target. It's wrong for `entities[].stale`, which promises a per-entity answer. An untouched dependency was reported stale whenever the target was, and a genuinely stale one reported current whenever the target was not.

This one is worth more than its size: `stale` is the field an agent reads to decide whether to trust the rest of the bundle. A confidently wrong value there is worse than no value, because it defeats the check rather than skipping it. Your existing staleness test passes because it only asserts on `freshness`, which was right all along.

Each non-target entity is now asked about its own source file. The target keeps the answer `collectFacts` already produced, so it isn't re-probed.

### The obvious fix would have been a performance trap

`isFileStale` calls `loadIngestBaseline`, which re-reads and re-parses the **entire** ingest mtime cache — a file holding every ingested path — on every call. Probing per entity before the budget slice would have turned one JSON parse into up to 500.

So the probe runs *after* budgeting (cost bounded by `--max-entities`), and I factored `createStaleProbe()` out of `isFileStale` so the baseline loads once per bundle. `isFileStale` now delegates to it and is otherwise unchanged, so `ix explain` / `ix locate` / `ix read` behave exactly as before.

The probe is injectable, so `buildBundle` stays pure under test — your determinism test still holds.

Two new tests, both fail with the fix reverted: one that a current dependency of a stale target reports current, and one pinning that probe count never exceeds the budget.

---

Running totals on this branch: 881 tests pass, `tsc --noEmit` clean, eslint 0 errors, knip no dead exports. Merge order note from my earlier comment still stands — #422 first, then rebase this and add the `ix_context` annotation entry.

PR #424, opened 2026-08-14
fix(read): contain ix read to the workspace roots

KageBinary:
## What

`ix read` would read **any absolute path on the host**. No workspace containment, no graph lookup.

```console
$ ix read /tmp/outside-any-workspace/secret.txt --format json
{
  "targetType": "file",
  "path": "/tmp/outside-any-workspace/secret.txt",
  "content": "SENTINEL-OUTSIDE-WORKSPACE-12345\n"
}
```

Three of the four resolution steps could land there:

| step | how it escapes |
|---|---|
| 2 — exact file path | `path.isAbsolute(rawTarget) ? rawTarget : …` honoured the caller's absolute path as-is (`read.ts:201`) |
| 3 — filename match | `matchPath` came back from the graph and was opened unchecked |
| 4 — symbol match | `absoluteFromSourceUri` passes an already-absolute `source_uri` straight through (`config.ts:182`) |

Steps 3 and 4 mean a backend row is enough — the caller does not have to ask for the path at all.

## Why it matters

This is the read primitive behind the `ix_read` MCP tool, so the caller is usually an agent and the target is a string it chose. A prompt-injected agent asking for `~/.ssh/id_rsa` got it.

Worth being precise about one thing: the read scope this file already imports does **not** cover this. `activeReadScope` bounds *graph queries* by workspace/system id; it never touches the filesystem.

## The fix

`isReadablePath` in `config.ts`, applied at all three filesystem paths.

Readable roots are the workspace the invocation resolves to **plus every workspace registered with `ix init`** — reads legitimately span those (a stitched system, `ix view --all`), and each is a path the user chose. The rest of the disk is not readable.

Symlinks are resolved on both sides. A link planted inside a workspace cannot hand back a file outside it; a workspace whose root is itself a symlink (macOS `/tmp`, a checkout reached through one) still reads normally. Resolving only the candidate would have broken the second case.

The refusal names the allowed roots — "denied" with no boundary is indistinguishable from a broken install, and an agent that cannot see the boundary just retries the same path.

```console
$ ix read /tmp/outside-any-workspace/secret.txt
Refusing to read file outside the workspace: /tmp/outside-any-workspace/secret.txt
  allowed root: /home/ianhock/ix-work
  allowed root: /home/ianhock/ix/IX-Memory
  Use --root to read from a different workspace, or ix init to register one.
```

`selectWorkspaceForCwd` now shares the containment helper it used to inline.

## Tests

`read-containment.test.ts`, 9 cases on the boundary itself rather than the command, so they hold regardless of which resolution step reaches the filesystem. They cover the two containment bugs that are easy to write by accident: a sibling sharing a string prefix with the root (`/home/ws-evil` vs `/home/ws`), and a symlink that is lexically inside but resolves outside.

## Verification

- 867 tests pass (62 files), `tsc --noEmit` clean, `eslint` 0 errors, `knip` no dead exports
- Confirmed by running the command before and after — output above is real, not illustrative

## Context

Found while reviewing the four open external-fork PRs. **Not caused by any of them** — this is pre-existing on `main`. It surfaced because #422 annotates `ix_read` as `readOnlyHint: true`, and MCP clients use that hint to skip the approval prompt. That PR is fine to land once this is in; without this, it removes the last human checkpoint in front of the behaviour above.

### DIFF
--- a/ix-cli/src/cli/__tests__/read-containment.test.ts
+++ b/ix-cli/src/cli/__tests__/read-containment.test.ts
@@ -1,0 +1,119 @@
+import { mkdirSync, mkdtempSync, rmSync, symlinkSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join, resolve } from "node:path";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+
+import { isPathInside, isReadablePath, readableRoots } from "../config.js";
+
+// `ix read` used to hand back any absolute path that existed on disk: no
+// workspace containment, no graph lookup. These cover the boundary itself
+// rather than the command, so they hold regardless of which of the four
+// resolution steps reaches the filesystem.
+
+let home: string;
+let workspace: string;
+let outside: string;
+let savedHome: string | undefined;
+let savedProfile: string | undefined;
+
+beforeEach(() => {
+  home = mkdtempSync(join(tmpdir(), "ix-read-home-"));
+  workspace = mkdtempSync(join(tmpdir(), "ix-read-ws-"));
+  outside = mkdtempSync(join(tmpdir(), "ix-read-outside-"));
+  savedHome = process.env.HOME;
+  savedProfile = process.env.USERPROFILE;
+  process.env.HOME = home;
+  process.env.USERPROFILE = home;
+
+  mkdirSync(join(home, ".ix"), { recursive: true });
+  writeFileSync(
+    join(home, ".ix", "config.yaml"),
+    ["endpoint: http://localhost:8090", "format: text", ""].join("\n"),
+  );
+  mkdirSync(join(workspace, "src"), { recursive: true });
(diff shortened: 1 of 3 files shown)


KageBinary:
Re-verified this and found a hole in my own tests, now fixed in `d0ea5d1`.

The original 9 cases pinned `isPathInside` and `isReadablePath`. I deleted all three `guardReadable` call sites from `read.ts` and **every one of them still passed** — they proved the boundary was correct without proving anything stood behind it. A helper nobody calls is still a correct helper.

Added three command-level cases that drive the registered command and assert the refusal reaches stderr and no content reaches stdout, for a plain outside path and for a symlink that is lexically inside the workspace. Both refusal cases fail with the guards removed.

That is the same failure mode I flagged on #423, so it was worth catching here before asking anyone else to fix theirs.

870 tests pass, typecheck and lint clean, CI green on `d0ea5d1`.

PR #426, opened 2026-08-14
fix(mcp): fail truncated in-process output

### LINKED ISSUE #425: [bug] MCP reports truncated in-process output as success
## Description

The in-process MCP runner reports a command as successful after its output exceeds the configured byte limit. The runner records that the output was truncated, but the final success calculation does not include that state. JSON-backed MCP tools can consequently return an empty structured object as a successful result.

## Steps to reproduce
1. Create an in-process runner with `maxOutputBytes` set to 100.
2. Register a `map` action that prints a JSON payload larger than 100 bytes.
3. Invoke `ix_map` through the MCP server.

## Expected behavior

The runner returns `ok: false`. The MCP result is marked as an error, preserves the truncation reason, and does not include `structuredContent`.

## Actual behavior

The runner returns `ok: true`. The MCP handler discards the warning and returns `{}` as successful `structuredContent`.

## CLI output

```text
[ix mcp] output exceeded 100 bytes and was truncated
```

## Environment
- OS: macOS 26.6.1
- Node: v26.7.0
- Ix version: main at fef671c

Hiro-Chiba:
Fixes #425

## Summary

Treat truncated in-process command output as an MCP execution failure instead of returning an incomplete structured result as successful.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Include the runner's truncation state in its success calculation.
- Preserve the existing truncation reason in the MCP error result.
- Cover the runner boundary and the MCP structured-output boundary with regression tests.

## Validation

- `npx vitest run src/cli/__tests__/mcp-runner.test.ts src/cli/__tests__/mcp-structured-output.test.ts src/cli/__tests__/mcp.test.ts` (48 passed)
- `npm test` (929 passed, 1 skipped; parser smoke passed)
- `npm run typecheck`
- `npm run build`
- `npm run knip`
- `git diff --check`

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/mcp-runner.test.ts
+++ b/ix-cli/src/cli/__tests__/mcp-runner.test.ts
@@ -403,7 +403,7 @@ describe("in-process ix runner", () => {
     });
   });
 
-  it("truncates output instead of growing the server heap without bound", async () => {
+  it("fails the run when output is truncated to protect the server heap", async () => {
     const run = createInProcessRunner({
       createProgram: createTestProgram,
       maxOutputBytes: 500,
@@ -411,6 +411,7 @@ describe("in-process ix runner", () => {
 
     const result = await run(["flood"]);
 
+    expect(result.ok).toBe(false);
     expect(result.stdout.length).toBeLessThanOrEqual(500);
     expect(result.stderr).toContain("exceeded 500 bytes and was truncated");
   });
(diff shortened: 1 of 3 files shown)


KageBinary:
Superseded by #436.

The `fix/mcp-error-output-cap` branch already contains this commit (`5e2bdcf`) verbatim, so merging #436 lands this change unchanged — nothing here is lost. Closing so the queue carries one PR per change rather than a stack that resolves to the same code.

The truncation fix itself is good; it just ships via #436. Thanks!

PR #430, opened 2026-08-14
fix(context): validate pick option

### LINKED ISSUE #429: [bug] ix context crashes on a non-numeric --pick value
## Description

`ix context` does not validate `--pick` as a complete integer before resolving an ambiguous target. A non-numeric value becomes `NaN` and reaches the shared picker, while a partially numeric value such as `1nope` is silently accepted as `1`.

## Steps to reproduce

1. Use a graph where the context target resolves to multiple candidates.
2. Run `ix context <target> --pick nope`.
3. Run `ix context <target> --pick 1nope`.

## Expected behavior

Both invalid values are rejected as option errors before target resolution, and the command exits with a non-zero status.

## Actual behavior

`nope` reaches the picker as `NaN` and produces a raw `TypeError` when the selected candidate is read. `1nope` is accepted as candidate `1` because `parseInt` only parses the numeric prefix.

## CLI output

```text
TypeError: Cannot read properties of undefined (reading 'id')
```

## Environment

- OS: macOS 26.6.1
- Node: v26.7.0

Hiro-Chiba:
Fixes #429

## Summary

Validate the `ix context --pick` value at the command boundary so malformed values cannot reach target resolution.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes

- Parse `--pick` with a strict whole-value integer validator.
- Reject malformed values before the context action and resolver run.
- Keep existing handling for valid integers, including the shared out-of-range behavior.
- Add command-level coverage for invalid and valid values.

## Validation

- `npx vitest run src/cli/__tests__/context-pick-validation.test.ts src/cli/__tests__/context.test.ts src/cli/__tests__/context-investigation.test.ts` (29 passed)
- `npm test` (67 files, 932 passed, 1 skipped, parser smoke passed)
- `npm run typecheck`
- `npm run build`
- `npm run lint` (0 errors, 41 existing warnings)
- `npm run knip`
- Built CLI smoke test confirmed `nope` exits with status 1 and a formatted option error
- `git diff --check`

## Checklist

- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/context-pick-validation.test.ts
+++ b/ix-cli/src/cli/__tests__/context-pick-validation.test.ts
@@ -1,0 +1,65 @@
+import { Command, type CommanderError } from "commander";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const resolveFileOrEntity = vi.hoisted(() => vi.fn());
+
+vi.mock("../resolve.js", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("../resolve.js")>()),
+  resolveFileOrEntity,
+}));
+
+import { registerContextCommand } from "../commands/context.js";
+
+async function runContext(args: string[]): Promise<{ error?: CommanderError; stderr: string }> {
+  const stderr: string[] = [];
+  const program = new Command()
+    .name("ix")
+    .exitOverride()
+    .configureOutput({ writeErr: (chunk) => stderr.push(chunk) });
+  registerContextCommand(program);
+
+  try {
+    await program.parseAsync(["context", "Widget", ...args], { from: "user" });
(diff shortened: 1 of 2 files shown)


KageBinary:
Superseded by #438.

The `fix/shared-pick-validation` branch already contains this commit (`2265009`), and then goes further: it removes the `parsePickOption` added here from `context.ts` in favour of the shared one in `options.ts` and applies it across all 16 commands that expose `--pick`. So #438 delivers this fix plus the consistent version — nothing here is lost.

Closing in favour of #438. Thanks!

PR #432, opened 2026-08-14
fix(mcp): reject conflicting ingest sources

### LINKED ISSUE #431: [bug] ix_ingest silently ignores path when github is also set
## Description

The MCP `ix_ingest` tool accepts both a local `path` and a GitHub source in the same request. The handler forwards both arguments, but the CLI prioritizes GitHub ingestion and silently ignores the positional local path.

## Steps to reproduce

1. Call `ix_ingest` with both `path: "src"` and `github: "owner/repo"`.
2. Inspect the command arguments or the resulting ingestion source.

## Expected behavior

The MCP tool rejects the conflicting source inputs before invoking the CLI and reports that only one source may be provided.

## Actual behavior

The tool reports success and invokes the CLI with both inputs. The GitHub branch is selected and the local path is ignored without a warning.

## CLI output

```text
ix ingest --github=owner/repo --format=json -- src
```

## Environment

- OS: macOS 26.6.1
- Node: v26.7.0

Hiro-Chiba:
Fixes #431

## Summary

Reject MCP `ix_ingest` requests that specify both a local path and a GitHub source instead of silently discarding the local path.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes

- Return an MCP error before the runner is called when both source inputs are present.
- Omit structured output for the rejected request.
- Preserve the existing local-only, GitHub-only, and no-source behavior.
- Add regression coverage for conflicting and valid source modes.

## Validation

- `npx vitest run src/cli/__tests__/mcp.test.ts` (25 passed)
- `npm test` (67 files, 929 passed, 1 skipped, parser smoke passed)
- `npm run typecheck`
- `npm run build`
- `npm run lint` (0 errors, 41 existing warnings)
- `npm run knip`
- `git diff --check`

## Checklist

- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/mcp.test.ts
+++ b/ix-cli/src/cli/__tests__/mcp.test.ts
@@ -280,9 +280,40 @@ describe("ix mcp", () => {
       return { ok: true, stdout: "{}", stderr: "" };
     });
 
-    await client.callTool({ name: "ix_ingest", arguments: { github: "owner/repo", limit: 25 } });
+    await client.callTool({
+      name: "ix_ingest",
+      arguments: { github: "owner/repo", since: "2026-01-01", limit: 25 },
+    });
 
-    expect(calls).toEqual([["ingest", "--github=owner/repo", "--limit=25", "--format=json"]]);
+    expect(calls).toEqual([
+      ["ingest", "--github=owner/repo", "--since=2026-01-01", "--limit=25", "--format=json"],
+    ]);
+  });
+
+  it("rejects simultaneous ix_ingest path and github sources before running the CLI", async () => {
+    const calls: string[][] = [];
+    const client = await connect(async (args) => {
+      calls.push(args);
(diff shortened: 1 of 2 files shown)


KageBinary (approved):
Reviewed. The guard sits at the layer issue #431 specifies, and both findings raised against it were refuted on a closer read:

- "It doesn't fix the CLI-side silent discard" — true that `ix ingest src --github=owner/repo` still drops the path, but `ingest.ts` is untouched by this PR (0-line diff) and #431's stated expectation is explicitly the MCP layer. Worth a separate issue, not a change here.
- "`since` on a path ingest" — pre-existing, reachable identically from the shell, and path ingest is already incremental via the mtime baseline, so the "asked for incremental, got full" harm doesn't exist.

Mutation-tested: remove the guard and the added regression test goes red at the production call site.

PR #434, opened 2026-08-14
fix(ingest): resolve renamed type imports

### LINKED ISSUE #433: [bug] renamed TypeScript imports drop inheritance and type-reference edges
## Description

Renamed TypeScript imports can lose inheritance and type-reference relationships even when the imported binding and provider symbol are both known.

## Steps to reproduce

1. Export `User` and `Base` from a provider file.
2. Import them with `import { type User as ExternalUser, Base as LocalBase }`.
3. Use `ExternalUser` in a type annotation and extend `LocalBase`.
4. Ingest the files and inspect `REFERENCES` and `EXTENDS` relationships.

## Expected behavior

The renamed local bindings resolve to the original provider symbols, producing the same relationships as non-renamed imports.

## Actual behavior

The import binding is parsed, but `REFERENCES` and `EXTENDS` relationships are omitted.

## CLI output

The resulting graph contains the file-level `IMPORTS` relationship but no `REFERENCES` or `EXTENDS` relationship for the renamed bindings.

## Environment

- OS: macOS 26.6.1
- Node: v26.7.0

Hiro-Chiba:
Fixes #433

## Summary

Resolve renamed consumer-side TypeScript bindings for inheritance and type references.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Prefer an exported public-name mapping when the provider defines one.
- Fall back to the direct provider symbol for ordinary renamed imports.
- Cover same-batch and cross-batch `EXTENDS`, `REFERENCES`, and `CALLS` resolution.

## Validation

- Focused core ingestion tests: 68 passed, 1 skipped.
- Ix CLI tests: 928 passed, 1 skipped, including parser smoke tests.

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/core-ingestion/src/__tests__/resolveEdges.test.ts
+++ b/core-ingestion/src/__tests__/resolveEdges.test.ts
@@ -583,6 +583,39 @@ describe('resolveEdges', () => {
     });
   });
 
+  it('resolves renamed imports used in inheritance and type references', () => {
+    const provider = parseFile(
+      '/repo/types.ts',
+      'export interface User { id: string; }\nexport class Base {}\n',
+    )!;
+    const consumer = parseFile(
+      '/repo/consumer.ts',
+      'import { type User as ExternalUser, Base as LocalBase } from "./types";\n'
+        + 'export class Child extends LocalBase {}\n'
+        + 'export function use(value: ExternalUser) { return value.id; }\n',
+    )!;
+
+    const resolved = resolveEdges([consumer, provider]);
+    expect(resolved).toContainEqual({
+      srcFilePath: '/repo/consumer.ts',
+      srcName: 'Child',
+      dstFilePath: '/repo/types.ts',
+      dstName: 'LocalBase',
+      dstQualifiedKey: 'Base',
+      predicate: 'EXTENDS',
+      confidence: 0.9,
+    });
+    expect(resolved).toContainEqual({
+      srcFilePath: '/repo/consumer.ts',
+      srcName: 'use',
+      dstFilePath: '/repo/types.ts',
+      dstName: 'ExternalUser',
+      dstQualifiedKey: 'User',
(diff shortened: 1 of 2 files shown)


KageBinary:
Reviewed. The renamed-import fix is correct and the cross-batch path holds up — `filePublicNames`/`fileHasSymbol` are single merged maps, so same-batch and cross-batch are the same branch.

One defect to fix before merge: the new fallback matches `binding.imported` against every parsed symbol, and for a default import that value is the literal sentinel `'default'`, not an export name. Any provider member *called* `default` — a method, getter, or static — then satisfies it and produces a 0.9-confidence edge to the wrong member. Existing tests all use a named default export (`export default actualName`), which resolves earlier, which is why CI is green.

Fix + regression test (mutation-tested): **Hiro-Chiba/Ix#1** → merge it into `fix/renamed-type-imports` and this PR updates in place.

KageBinary:
Update on how this lands: **#443** now carries your commits plus the review fixes, targeting `main` directly. It's up for review by @josephismikhail.

The reasoning for each fix is in this thread above and in #443's description.

Alot1z:
Superseded by #443 which carries this commit (819d67e) plus a default-import guard fix. No action needed here â€” see #443 for the complete review.

Alot1z:
Superseded by #443 which carries this commit (819d67e) plus a default-import guard fix. No action needed here â€” see #443 for the complete review.

KageBinary:
Landed on `main` — your commit `819d67e` is an ancestor of main via #443, which carried it forward with an added guard (a default import must not bind to a provider member that happens to be named `default`). GitHub closed this one without the merged marker because the superset PR was the thing merged, but your commit and authorship are in the history. Thanks!

PR #436, opened 2026-08-14
fix(mcp): cap thrown error output

### LINKED ISSUE #435: [bug] thrown command errors bypass the MCP output cap
## Description

The in-process MCP output limit applies to captured console and stream output, but a thrown command error is appended after that limit is enforced. A large `Error.message` can therefore bypass `maxOutputBytes` and be returned in full through MCP.

## Steps to reproduce

1. Configure the in-process runner with `maxOutputBytes` set to 128.
2. Run an action that throws `new Error("E".repeat(4096))`.
3. Inspect the runner result or the corresponding MCP error response.

## Expected behavior

The returned error text is bounded by the configured output limit and includes the normal truncation marker.

## Actual behavior

The result contains the full 4096-byte error message without a truncation marker.

## CLI output

The runner returns `ok: false`, but `stderr` still contains all 4096 bytes. The MCP response forwards the same oversized error text.

## Environment

- OS: macOS 26.6.1
- Node: v26.7.0

Hiro-Chiba:
Fixes #435

Depends on #426.

## Summary

Apply the in-process MCP output cap to thrown command errors as well as captured output.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Route thrown error text through the existing bounded-output collector.
- Preserve the existing error result while adding the standard truncation marker when needed.
- Cover both the runner boundary and the MCP error response.

## Validation

- Focused runner and MCP tests: 31 passed.
- Ix CLI tests: 931 passed, 1 skipped, including parser smoke tests.

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/mcp-runner.test.ts
+++ b/ix-cli/src/cli/__tests__/mcp-runner.test.ts
@@ -403,7 +403,7 @@ describe("in-process ix runner", () => {
     });
   });
 
-  it("truncates output instead of growing the server heap without bound", async () => {
+  it("fails the run when output is truncated to protect the server heap", async () => {
     const run = createInProcessRunner({
       createProgram: createTestProgram,
       maxOutputBytes: 500,
@@ -411,6 +411,7 @@ describe("in-process ix runner", () => {
 
     const result = await run(["flood"]);
 
+    expect(result.ok).toBe(false);
     expect(result.stdout.length).toBeLessThanOrEqual(500);
     expect(result.stderr).toContain("exceeded 500 bytes and was truncated");
   });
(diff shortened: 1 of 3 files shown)


KageBinary:
One more, non-blocking but worth folding in: routing thrown text through `appendChunk` means it is dropped outright once the cap has fired, because `appendChunk` starts with `if (run.truncated) return;`. A command that floods and *then* throws reaches the client as only `output exceeded N bytes and was truncated` — no reason. main appended it unconditionally, so it's a regression in the case where the diagnostic matters most.

Fix + test: **Hiro-Chiba/Ix#4**. Your existing cap test is unaffected and passes unchanged.

KageBinary:
Update on how this lands: **#444** now carries your commits plus the review fixes, targeting `main` directly. It's up for review by @josephismikhail.

The reasoning for each fix is in this thread above and in #444's description.

Alot1z:
Superseded by #444 which carries both commits (5e2bdcf, 033181f) plus a fix for the failure-reason regression. No action needed here â€” see #444 for the complete review.

PR #438, opened 2026-08-14
fix(cli): validate pick options consistently

### LINKED ISSUE #437: [bug] malformed --pick values are accepted by commands other than context
## Description

Most commands that accept `--pick` do not validate the full value consistently. Inputs such as `1nope`, zero, negative numbers, or unsafe integers can be partially parsed or reach command execution instead of failing at the CLI boundary.

## Steps to reproduce

1. Run a command other than `context` that accepts `--pick`, such as `ix trace Widget --pick 1nope`.
2. Repeat with `--pick 0`, `--pick -1`, or an integer larger than `Number.MAX_SAFE_INTEGER`.
3. Observe whether the command action starts.

## Expected behavior

All commands accept only a positive safe integer and reject the complete invalid value before running resolution or backend work.

## Actual behavior

Validation differs by command. Partial numeric strings can be accepted, and other invalid values reach later resolution paths.

## CLI output

The affected commands do not consistently report an invalid `--pick` value at argument parsing time.

## Environment

- OS: macOS 26.6.1
- Node: v26.7.0

Hiro-Chiba:
Fixes #437

Depends on #430.

## Summary

Use one strict `--pick` parser across all commands that support candidate selection.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Add a shared parser that accepts only positive safe integers.
- Apply it to all 16 commands that expose `--pick`.
- Reject partial strings, non-positive values, and unsafe integers before command actions run.

## Validation

- Related command tests: 128 passed.
- Ix CLI tests: 954 passed, 1 skipped, including parser smoke tests.

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/context-pick-validation.test.ts
+++ b/ix-cli/src/cli/__tests__/context-pick-validation.test.ts
@@ -1,0 +1,65 @@
+import { Command, type CommanderError } from "commander";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const resolveFileOrEntity = vi.hoisted(() => vi.fn());
+
+vi.mock("../resolve.js", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("../resolve.js")>()),
+  resolveFileOrEntity,
+}));
+
+import { registerContextCommand } from "../commands/context.js";
+
+async function runContext(args: string[]): Promise<{ error?: CommanderError; stderr: string }> {
+  const stderr: string[] = [];
+  const program = new Command()
+    .name("ix")
+    .exitOverride()
+    .configureOutput({ writeErr: (chunk) => stderr.push(chunk) });
+  registerContextCommand(program);
+
+  try {
+    await program.parseAsync(["context", "Widget", ...args], { from: "user" });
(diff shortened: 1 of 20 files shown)


KageBinary (approved):
Reviewed. Consolidates four divergent `--pick` validators into one strict parser across all 16 commands, and carries #430's commit (now closed as superseded — this branch contains it and then replaces its local helper with the shared one).

Checked the thing most likely to bite: the stricter validator rejecting input that used to work. It doesn't, in any way that matters. `--pick 0` and `--pick -3` passed main's `opts.pick ? parseInt(...) : undefined` through to `applyPick`, which printed `out of range (1-N)` and exited 0; now they're rejected at the option boundary with exit 1. Strictly better, and `parseInt("2abc") === 2` silently accepting garbage is gone too.

One coverage note, not a blocker: no behavioural test asserts a *valid* `--pick` reaches the resolver for any command except `context` — the other 15 are guarded by source-text greps. That's the same shape of coverage main already had, and `tsc` catches the type half.

PR #440, opened 2026-08-14
fix(ingest): resolve tsconfig module mappings

### LINKED ISSUE #439: [bug] tsconfig paths and baseUrl are ignored during edge resolution
## Description

TypeScript and JavaScript imports configured through `tsconfig.json` or `jsconfig.json` `paths` and `baseUrl` are not used during graph edge resolution. Ix falls back to a repository-wide filename stem, which can select an unrelated file or omit the edge when names are ambiguous.

## Steps to reproduce

1. Configure `paths` with `"@core": ["src/adapters/worker"]`.
2. Import and call `execute` from `@core`.
3. Add an unrelated `legacy/core.ts` that also exports `execute`.
4. Ingest the repository and inspect `IMPORTS` and `CALLS`.

## Expected behavior

The configured module target resolves to `src/adapters/worker.ts`. A missing configured target remains unresolved instead of falling back to an unrelated stem.

## Actual behavior

Ix can connect both relationships to `legacy/core.ts`, or omit them when multiple stem candidates exist. Equivalent `baseUrl` imports are also not resolved from their configured root.

## CLI output

The wrong target can receive `IMPORTS` and `CALLS` with confidence `0.9`, while the configured target receives no edge.

## Environment

- OS: macOS 26.6.1
- Node: v26.7.0

Hiro-Chiba:
Fixes #439

## Summary

Resolve local TypeScript and JavaScript imports through `tsconfig.json` and `jsconfig.json` module mappings before filename-stem fallback.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Read nearest standard config files with JSONC comments, trailing commas, local relative `extends`, `paths`, and `baseUrl`.
- Select runtime or type targets according to the relationship and caller language.
- Treat a missing configured target as authoritative to prevent unrelated stem matches.
- Preserve existing fallback behavior when a configured provider exists but does not define the requested symbol.
- Match tracked paths on case-insensitive filesystems while returning their real casing.
- Keep package `exports` mappings outside this change.

## Validation

- Focused core ingestion tests: 69 passed, 1 skipped.
- Resolver and ingestion tests: 14 passed.
- Ix CLI tests: 936 passed, 1 skipped, including parser smoke tests.

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/core-ingestion/src/__tests__/resolveEdges.test.ts
+++ b/core-ingestion/src/__tests__/resolveEdges.test.ts
@@ -287,6 +287,115 @@ describe('resolveEdges', () => {
     });
   });
 
+  it('prefers configured JS/TS module targets over unrelated stem matches', () => {
+    const consumer = parseFile(
+      '/repo/src/consumer.ts',
+      'import { execute } from "@core";\nexport function run() { return execute(); }\n',
+    )!;
+    const intended = parseFile(
+      '/repo/src/adapters/worker.ts',
+      'export function execute() { return "worker"; }\n',
+    )!;
+    const unrelated = parseFile(
+      '/repo/legacy/core.ts',
+      'export function execute() { return "legacy"; }\n',
+    )!;
+
+    const resolved = resolveEdges([consumer, intended, unrelated], undefined, undefined, {
+      resolveModuleSpecifier: (_source, specifier) =>
+        specifier === '@core' ? ['/repo/src/adapters/worker.ts'] : undefined,
+    });
+
+    expect(resolved).toContainEqual(expect.objectContaining({
+      srcFilePath: '/repo/src/consumer.ts',
+      dstFilePath: '/repo/src/adapters/worker.ts',
+      predicate: 'IMPORTS',
+      confidence: 0.9,
+    }));
+    expect(resolved).toContainEqual(expect.objectContaining({
+      srcFilePath: '/repo/src/consumer.ts',
+      dstFilePath: '/repo/src/adapters/worker.ts',
+      predicate: 'CALLS',
+      confidence: 0.9,
+    }));
+    expect(resolved.some(edge => edge.dstFilePath === '/repo/legacy/core.ts')).toBe(false);
+  });
+
+  it('uses mapped runtime files for calls and declaration files for references', () => {
+    const consumer = parseFile(
+      '/repo/src/consumer.ts',
+      'import { run, type Foo } from "@pkg";\n'
+        + 'export function call() { return run(); }\n'
+        + 'export function use(value: Foo) { return value; }\n',
+    )!;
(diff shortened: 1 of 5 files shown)


KageBinary:
Reviewed. The `paths` resolution and the `$`-token fix in `bf7a4b4` are both right, and the CodeQL alert on this PR is **stale** — its multi-star rationale is unreachable because `matchPathPattern` rejects a second `*` before substitution. Worth re-running the scan and dismissing.

Blocking issue: `return config?.baseUrl ? [] : undefined` makes a bare `baseUrl` authoritative. `[]` means *definitively not local*, and `baseUrl` matches every non-relative specifier — so on any repo that sets it (Next.js, CRA, Nest, most monorepo templates) every third-party import including `react` becomes an authoritative miss and its edges are dropped. Ix's own tsconfigs set neither `baseUrl` nor `paths`, so dogfooding can't see it.

Also flagging `resolveExtends`'s `existsSync`-then-read — CodeQL `js/file-system-race`, which this repo has a settled pattern for.

Both fixed, with 5 tests: **Hiro-Chiba/Ix#2**. Note it replaces your `treats an unresolved baseUrl module as authoritative` test — that behaviour is the defect. Your explicit-`paths` authority test is untouched and still passes.

KageBinary:
Added a second commit to **Hiro-Chiba/Ix#2**: memoize the resolver.

Resolution runs per bound relationship and `index.ts` issues two full lookups of the same triple for each, so a repo with `paths`/`baseUrl` pays the walk repeatedly. Measured on a synthetic 2000-file workspace: **24.87 µs/call → 3.38 µs/call, ~7.4x**.

Worth flagging the trap, since the obvious key is wrong: `resolveCandidate` reads the source *extension* as well as the directory (a `.js` caller prefers `.js` over `.ts`). Keying on `dirname` alone serves one caller's answer to the other in the same folder and silently flips the preference order — there's a test that runs both orders plus a repeat, and it reds if the extension is dropped from the key.

KageBinary:
Update on how this lands: **#445** now carries your commits plus the review fixes, targeting `main` directly. It's up for review by @josephismikhail.

The reasoning for each fix is in this thread above and in #445's description.

Alot1z:
Superseded by #445 which carries both commits (7d83423, bf7a4b4) plus baseUrl authority, security, and memoization fixes. No action needed here â€” see #445 for the complete review.

KageBinary:
Landed on `main` — your commits `7d83423` and `bf7a4b4` are ancestors of main via #445, which carried them forward with bounded/confined tsconfig reads and a fix so `baseUrl` no longer claims authority over every bare specifier (which would otherwise suppress the stem fallback). GitHub closed this one without the merged marker because the superset PR was the thing merged, but your commits and authorship are in the history. Thanks!

PR #442, opened 2026-08-14
fix(php): resolve imports by namespace

### LINKED ISSUE #441: [bug] PHP imports can resolve to classes from the wrong namespace
## Description

PHP imports are flattened to their final class name during graph resolution. If a repository contains another class with the same short name in a different namespace, Ix can create high-confidence `IMPORTS`, `REFERENCES`, and `CALLS` relationships to the wrong class.

## Steps to reproduce

1. Add `/repo/App/User.php` with `namespace App; class User`.
2. In another file, write `use Vendor\\Package\\User;` and use `User` as a parameter, constructor, and member-call receiver.
3. Do not add a `Vendor\\Package\\User` provider.
4. Ingest the files and inspect the resolved relationships.

## Expected behavior

No relationship targets `App\\User`. When `Vendor\\Package\\User` exists, imports and uses resolve to that exact class, including PHP's case-insensitive class naming.

## Actual behavior

Ix resolves the explicit import and its uses to `App\\User` with confidence `0.9` because the short class name is unique.

## CLI output

The graph contains four incorrect relationships to `/repo/App/User.php`: one `IMPORTS`, one `REFERENCES`, and two `CALLS` relationships.

## Environment

- OS: macOS 26.6.1
- Node: v26.7.0

Hiro-Chiba:
Fixes #441

## Summary

Resolve explicit PHP type imports by their declared namespace instead of a repository-wide short-name match.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Preserve semicolon, braced, and global PHP namespace scope on parsed type entities.
- Build case-insensitive fully qualified indexes for PHP classes, interfaces, traits, and enums across batches.
- Resolve explicit type imports through their exact fully qualified name without changing function and constant import behavior.
- Distinguish constructor, function, member, and static calls through edge resolution and patch construction.
- Leave aliased and grouped PHP import parsing outside this change.

## Validation

- Focused PHP parser and resolution tests: 103 passed, 1 skipped.
- Ix CLI tests: 928 passed, 1 skipped, including parser smoke tests.

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/core-ingestion/src/__tests__/patchBuilder.resolution.test.ts
+++ b/core-ingestion/src/__tests__/patchBuilder.resolution.test.ts
@@ -109,6 +109,32 @@ describe('buildPatchWithResolution', () => {
     }));
   });
 
+  it('keeps same-named PHP constructor and function calls distinct', () => {
+    const consumer = parseFile(
+      '/repo/Consumer.php',
+      `<?php
+use Vendor\\Package\\Helper;
+function Helper(): void {}
+function run(): void { new Helper(); Helper(); }
+      `,
+    )!;
+    const provider = parseFile(
+      '/repo/Vendor/Package/Helper.php',
+      '<?php namespace Vendor\\Package; class Helper {}',
+    )!;
+    const resolved = resolveEdges([consumer, provider]);
+    const patch = buildPatchWithResolution(consumer, 'hash', '', resolved);
+    const callEdges = patch.ops.filter(op => op.type === 'UpsertEdge' && op.predicate === 'CALLS');
+
+    expect(callEdges).toHaveLength(2);
+    expect(callEdges).toContainEqual(expect.objectContaining({
+      dst: nodeId('/repo/Vendor/Package/Helper.php', 'Helper'),
+    }));
+    expect(callEdges).toContainEqual(expect.objectContaining({
+      dst: nodeId('/repo/Consumer.php', 'Helper'),
+    }));
+  });
+
   it('materialises external stub node for unresolved :: package calls', () => {
     const file = '/repo/model.R';
     const externalNodeId = nodeId('external://dplyr', 'dplyr::filter');
--- a/core-ingestion/src/__tests__/queries.php.test.ts
+++ b/core-ingestion/src/__tests__/queries.php.test.ts
@@ -22,6 +22,83 @@ use Vendor\\Contracts\\DomainService;
     });
   });
 
+  it('records PHP namespace scope for semicolon and braced declarations', () => {
+    const result = parseFile(
+      '/repo/Types.php',
+      `<?php
+namespace First\\Space;
+class FirstType {}
+namespace Second\\Space;
+class SecondType {}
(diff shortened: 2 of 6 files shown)


KageBinary:
Reviewed. The FQCN indexing is a real improvement and the `trimPhpNamespace` rewrite in `83079ff` correctly kills the flagged ReDoS — that CodeQL alert is **stale** and can be dismissed after a re-scan.

Blocking issue: widening the capture to `[(qualified_name) (name)]` also captures the alias, because in `namespace_use_clause` the alias is a direct `(name)` sibling. `use Vendor\\Package\\User as Admin;` therefore emits an `IMPORTS` edge to an unrelated `Admin` class the file never imports, and promotes the wrong short-name matches from 0.5 to **0.9** — the opposite of what this PR is for.

Two related clause-scoping bugs: `importAliased` is computed from the whole statement, so one alias in a comma-separated `use` disables FQCN resolution for every clause; and that same unanchored `/\\s+as\\s+/` is a second instance of the ReDoS class CodeQL missed (~3.7s on a `use` padded with 100k spaces, which is valid PHP).

All three fixed by one root-cause change — scope to the clause: **Hiro-Chiba/Ix#3**.

KageBinary:
Added a second commit to **Hiro-Chiba/Ix#3** covering two further regressions I measured against main rather than inferred:

**1. A `use` leaks across braced namespace blocks.** `phpFqcnImportsByLocal` is keyed by file, but a `use` is scoped to its namespace block:

```php
namespace A { use Vendor\\Package\\Thing; function run(Thing $t) {...} }
namespace B { class Thing {} function go(Thing $t) {...} }
```

`go()` resolved to `Vendor/Package/Thing.php` at 0.9 although `B` declares its own `Thing` two lines up. main emits no edge for `go` at all. One namespace per file is the PSR-12 norm, so the index now skips a multi-block file rather than mis-resolve it — restoring main's behaviour exactly for that shape.

**2. Duplicate CALLS edges reach the graph.** `phpCallKind` splits the per-file dedup key, so `log()` and `Logger::log()` survive as two relationships. When both resolve to the same member, patch-builder salts the edge id by call kind and emits two `UpsertEdge` ops with identical `src`/`dst`/`predicate` but different ids — and the dedup downstream keys on id, so both commit. Measured: main 1 op, this branch 2. The salt still applies when the two kinds resolve to *different* nodes, which is what it's for.

KageBinary:
Update on how this lands: **#446** now carries your commits plus the review fixes, targeting `main` directly. It's up for review by @josephismikhail.

The reasoning for each fix is in this thread above and in #446's description.

Alot1z:
Superseded by #446 which carries both commits (ea79b11, 83079ff) plus clause-scoping, namespace block, DoS, and dedup fixes. No action needed here â€” see #446 for the complete review.

PR #443, opened 2026-08-14
fix(ingest): resolve renamed type imports, with a default-import guard

KageBinary:
Carries **@Hiro-Chiba's #434** plus one review fix. Landing it here closes #434 as merged.

> ⚠️ **Merge with a merge commit, not squash.** This branch contains Hiro-Chiba's commit alongside mine; a squash would collapse an external contributor's work into one commit authored by me.

## Their change (`819d67e`)

Resolves renamed consumer-side TS bindings for inheritance and type references — `import { Base as LocalBase }` now resolves `EXTENDS`/`REFERENCES`/`CALLS` to the provider's real symbol.

## My fix (`eab1075`)

The fallback matched `binding.imported` against every parsed symbol in the provider file. For `import run from "./m"`, `binding.imported` is the literal sentinel `'default'` (set at `index.ts:514`), not an export name — and `fileHasSymbol` holds every entity including non-exported methods. So any provider member *called* `default` satisfied it:

```
/repo/m.ts  export class M { default() { return 1; } }
            export default new M();
/repo/n.ts  import run from "./m"; export function go() { return run(); }
```

main emits nothing; #434 emits `CALLS go -> m.ts` with `dstQualifiedKey: "M.default"` at **0.9** — confident, and wrong. Guarded with the same `!== 'default'` exclusion `index.ts:3199` already uses.

CI was green because every existing default-import test uses a *named* default export, which resolves before the fallback is reached.

## Verification

Mutation-tested: delete the guard and the added test alone goes red. `tsc --noEmit` clean; core-ingestion shows the same 6 pre-existing fixture-missing failures as the untouched base.

Reviewed as part of a full pass over all 9 open PRs; the per-PR findings and a security review of the combined tree are in the thread on #434.

### DIFF
--- a/core-ingestion/src/__tests__/resolveEdges.test.ts
+++ b/core-ingestion/src/__tests__/resolveEdges.test.ts
@@ -616,6 +616,25 @@ describe('resolveEdges', () => {
     });
   });
 
+  it('does not bind a default import to a provider member named "default"', () => {
+    // `imported` is the sentinel 'default' for `import run from "./m"`, not an
+    // export name. Without the guard, the renamed-import fallback matches the
+    // *method* `M.default` and emits a confident edge to the wrong member.
+    const provider = parseFile(
+      '/repo/m.ts',
+      'export class M { default() { return 1; } }\nexport default new M();\n',
+    )!;
+    const consumer = parseFile(
+      '/repo/n.ts',
+      'import run from "./m";\nexport function go() { return run(); }\n',
+    )!;
+
+    const resolved = resolveEdges([consumer, provider]);
+    expect(
+      resolved.filter((e) => e.srcName === 'go' && e.dstQualifiedKey === 'M.default'),
+    ).toEqual([]);
+  });
+
   it('uses caller-supplied parse results instead of re-parsing', () => {
     // The CLI parses prescan sources on its worker pool and hands the results
     // in, so the index costs no main-thread parsing. Proven by supplying a
--- a/core-ingestion/src/__tests__/resolveEdges.test.ts
+++ b/core-ingestion/src/__tests__/resolveEdges.test.ts
@@ -583,6 +583,39 @@ describe('resolveEdges', () => {
     });
   });
 
+  it('resolves renamed imports used in inheritance and type references', () => {
+    const provider = parseFile(
+      '/repo/types.ts',
+      'export interface User { id: string; }\nexport class Base {}\n',
+    )!;
+    const consumer = parseFile(
+      '/repo/consumer.ts',
+      'import { type User as ExternalUser, Base as LocalBase } from "./types";\n'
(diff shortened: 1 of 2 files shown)


Alot1z:
## Review â€” PR #443: Renamed TypeScript imports + default-import guard

**Verdict: APPROVE**

### What this PR does
Carries Hiro-Chiba's #434 (resolve renamed TS imports for EXTENDS/REFERENCES/CALLS) plus a default-import guard fix by KageBinary.

### Commit analysis
- `819d67e` (Hiro-Chiba): Resolves renamed consumer-side TS bindings â€” `import { Base as LocalBase }` now resolves EXTENDS/REFERENCES/CALLS to the provider's real symbol.
- `eab1075` (KageBinary): Guards the renamed-import fallback against the `'default'` sentinel. For `import run from ./m`, `binding.imported` is the literal sentinel `'default'`, not an export name. Without the guard, any provider member *called* `default` satisfies the lookup and produces a 0.9-confidence edge to the wrong member.

### Code review
The fix at `index.ts:3992` adds `binding.imported !== 'default'` to the fallback condition, matching the exclusion already used at `index.ts:3199` when building rename bindings. Named renamed imports (the case this PR exists to fix) are unaffected: they short-circuit through `filePublicNames`.

### Test coverage
- `resolves renamed imports used in inheritance and type references` â€” verifies EXTENDS and REFERENCES edges
- `does not bind a default import to a provider member named default` â€” regression test
- Mutation-tested: deleting the guard makes the new test fail
- `tsc --noEmit` clean

### Security
No security concerns. No OSS/Pro boundary issues.

### Notes
CI was green before the guard because every existing default-import test uses a named default export (`export default actualName`), which resolves before the fallback.

Alot1z:
Found and independently verified an additional ingest correctness regression in the current #443 implementation: a renamed import (`import { Base as LocalBase } from "./m"`) falls back to any symbol in the provider file — including a class method — so `extends LocalBase` emits a confident wrong EXTENDS edge to `M.Base` (@0.9), where the PR base never emitted one. The same happens for `import { type User as ExternalUser }`, which emits REFERENCES to `M.User`.

The issue is still reproducible against the current upstream PR state (head `eab1075`), and I did not find an existing fix or duplicate PR addressing it.

Follow-up commit `cba11a3` applies the minimal fix — require a plain module-level key for the renamed-import fallback binding (a member can never be a module export) — and adds regression coverage for the renamed EXTENDS and renamed REFERENCES cases.

Verification: resolveEdges 70/70 on the fix tree, `tsc` clean; the wrong edges reproduce against `eab1075` and are eliminated after.

Follow-up commit:
https://github.com/Alot1z/Ix-remap/commit/cba11a3363b1fbe57819381f05490d2f395674cd

PR #445, opened 2026-08-14
fix(ingest): resolve tsconfig module mappings, with authority + security fixes

KageBinary:
Carries **@Hiro-Chiba's #440** plus three review fixes, two of them security. Landing it here closes #440 as merged.

> ⚠️ **Merge with a merge commit, not squash** — contains Hiro-Chiba's two commits alongside mine.

## Their change (`7d83423`, `bf7a4b4`)

Resolves local TS/JS imports through `tsconfig.json`/`jsconfig.json` `paths`/`baseUrl` before the filename-stem fallback. New 360-line module. The follow-up commit fixes `$`-token expansion in wildcard substitution — **the CodeQL alert on #440 is stale** and can be dismissed after a re-scan (its multi-star rationale is unreachable: `matchPathPattern` rejects a second `*` before substitution).

## My fixes

**1. `795ec2c` — `baseUrl` claimed authority over every bare specifier.** `return config?.baseUrl ? [] : undefined` — but `[]` means *definitively not local*, and `baseUrl` matches every non-relative specifier. On any repo that sets it (the default in Next.js, CRA, Nest, most monorepo templates) every third-party import including `react` became an authoritative miss and its IMPORTS/CALLS edges were silently dropped. Authority now requires an *explicit* `paths` pattern; a bare `baseUrl` and a catch-all `"*"` rule return "no opinion". This **replaces their `treats an unresolved baseUrl module as authoritative` test** — that behaviour is the defect. Their explicit-`paths` authority test is untouched and still passes.

**2. `9e58df1` — three security holes on the config-reading path.** All reachable from repo content, and all *before* ingestion's size gate (the resolver is built at `ingest.ts:1049`, 51 lines ahead of the `MAX_FILE_BYTES` check):
- **No size cap.** A 64 MB JSONC file — which git packs to a ~95 KB repo — gives `FatalProcessOutOfMemory`, signal 6, uncatchable. main skips it as `tooLarge`, so this was the first pre-gate whole-file read and it bypassed an existing control.
- **Device nodes and FIFOs.** `openSync` follows symlinks, so a link committed in the repo pointing at `/dev/zero` read forever (**exit 134, core dumped**); a FIFO blocked the event loop. Both report size 0, so a size check alone doesn't help. Under MCP the in-process runner means `withTimeout` can never fire and the server wedges uncancellably.
- **No containment.** `resolveExtends` only checked `startsWith(".")`, which isn't containment — `resolve` clamps at `/`, so enough `..` segments name any absolute path.

Fixed by opening once and fstat-ing the same handle (no stat-then-read window — CodeQL `js/file-system-race`, the pattern `ingest.ts` already uses), rejecting non-regular files and anything over 1 MB, and filtering `extends` candidates to inside the workspace. No content ever escaped — `resolveCandidate` only consults an in-memory Map built from the caller's file list.

**3. `c43303e` — memoization.** 24.87 µs → 3.38 µs per call, ~7.4x. Note the trap: the obvious `dirname` key is **wrong**, because `resolveCandidate` also reads the source *extension* (a `.js` caller prefers `.js`); keying on the directory alone serves one caller's answer to another. There's a test that runs both orders.

## Verification

Every fix mutation-tested individually. ix-cli 944 passed, 1 skipped; `typecheck`, `lint` (0 errors), `knip` clean.

### DIFF
--- a/core-ingestion/src/__tests__/resolveEdges.test.ts
+++ b/core-ingestion/src/__tests__/resolveEdges.test.ts
@@ -287,6 +287,115 @@ describe('resolveEdges', () => {
     });
   });
 
+  it('prefers configured JS/TS module targets over unrelated stem matches', () => {
+    const consumer = parseFile(
+      '/repo/src/consumer.ts',
+      'import { execute } from "@core";\nexport function run() { return execute(); }\n',
+    )!;
+    const intended = parseFile(
+      '/repo/src/adapters/worker.ts',
+      'export function execute() { return "worker"; }\n',
+    )!;
+    const unrelated = parseFile(
+      '/repo/legacy/core.ts',
+      'export function execute() { return "legacy"; }\n',
+    )!;
+
+    const resolved = resolveEdges([consumer, intended, unrelated], undefined, undefined, {
+      resolveModuleSpecifier: (_source, specifier) =>
+        specifier === '@core' ? ['/repo/src/adapters/worker.ts'] : undefined,
+    });
+
+    expect(resolved).toContainEqual(expect.objectContaining({
+      srcFilePath: '/repo/src/consumer.ts',
+      dstFilePath: '/repo/src/adapters/worker.ts',
+      predicate: 'IMPORTS',
+      confidence: 0.9,
+    }));
+    expect(resolved).toContainEqual(expect.objectContaining({
+      srcFilePath: '/repo/src/consumer.ts',
+      dstFilePath: '/repo/src/adapters/worker.ts',
+      predicate: 'CALLS',
+      confidence: 0.9,
+    }));
+    expect(resolved.some(edge => edge.dstFilePath === '/repo/legacy/core.ts')).toBe(false);
+  });
+
+  it('uses mapped runtime files for calls and declaration files for references', () => {
+    const consumer = parseFile(
+      '/repo/src/consumer.ts',
+      'import { run, type Foo } from "@pkg";\n'
+        + 'export function call() { return run(); }\n'
+        + 'export function use(value: Foo) { return value; }\n',
+    )!;
+    const runtime = parseFile('/repo/lib/index.js', 'export function run() { return 1; }\n')!;
+    const declarations = parseFile('/repo/lib/index.d.ts', 'export interface Foo { id: string; }\n')!;
+
+    const resolved = resolveEdges([consumer, runtime, declarations], undefined, undefined, {
+      resolveModuleSpecifier: (_source, _specifier, kind) =>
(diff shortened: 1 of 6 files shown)


KageBinary:
Added one more commit (`adc97c1`) closing the symlink gap I flagged as non-blocking earlier.

Containment was pure path arithmetic, which cannot see through a symlinked directory: every segment of `./linked/tsconfig.json` is inside the workspace even when `linked/` points elsewhere, so an ingested repo could still name a file outside the tree by shipping one link. Wrote the failing test first to confirm the escape was real rather than theoretical.

The lexical check stays — it is what stops `../` traversal from the config text — and the resolved path is now checked too. The resolve happens while the handle is open and is confirmed to name the same inode, so swapping the link between the open and the resolve cannot smuggle an outside file in under an inside name.

The part I would most like a second pair of eyes on: **the workspace root is resolved too.** Comparing a resolved file against an unresolved root rejects every config whenever the root is itself reached through a link — macOS `/var` -> `/private/var`, a home on a network mount, a `~/code` symlink. That would have switched tsconfig resolution off entirely for those users, with no error and nothing in CI to catch it, since the matrix here is ubuntu + windows. The added test uses a symlinked root specifically to hold that down.

Both halves mutation-tested: dropping the containment call fails only the escape test, and comparing against the unresolved root fails only the symlinked-root test. Full `test:unit` is 949 passing.

Context for the review: this is the same module where my earlier "fix" for FIFOs shipped a guard that could not fire (the non-regular-file check sat *after* an open that blocks forever on a FIFO), with a comment asserting it worked and a passing `/dev/zero` test that exercised a different path. That is why I am flagging the root-resolution decision explicitly rather than letting it ride on my own reading.

Alot1z:
Found and independently verified an additional ingest correctness regression in the current #445 implementation: a configured module mapping names the module authoritatively, but the renamed-import fallback still binds `import { Base as LocalBase } from "@core"` to a class method `W.Base` — a member can never be a module export — emitting a confident wrong EXTENDS edge (@0.9) where the base never had one. The same happens for `import { type User as ExternalUser }`, which emits REFERENCES to `W.User`.

The issue is still reproducible against the current upstream PR state (head `adc97c1`), and I did not find an existing fix or duplicate PR addressing it.

Follow-up commit `f9274cc` applies the minimal fix — the same plain-key requirement on the configured-binding path — and adds regression coverage for the configured renamed EXTENDS and REFERENCES cases.

Verification: resolveEdges 72/72 on the fix tree, `tsc` clean; the wrong edges reproduce against `adc97c1` and are eliminated after.

Follow-up commit:
https://github.com/Alot1z/Ix-remap/commit/f9274cc0870c65a496a5b983c814d9fed6a65c0f

PR #446, opened 2026-08-14
fix(php): resolve imports by namespace, with clause-scoping and DoS fixes

KageBinary:
Carries **@Hiro-Chiba's #442** plus three review fixes, one a DoS. Landing it here closes #442 as merged.

> ⚠️ **Merge with a merge commit, not squash** — contains Hiro-Chiba's two commits alongside mine.

## Their change (`ea79b11`, `83079ff`)

Resolves explicit PHP type imports by declared namespace instead of a repo-wide short-name match. The follow-up replaces the flagged ReDoS regex with a char-code trim — **the CodeQL alert on #442 is stale** and can be dismissed after a re-scan.

## My fixes

**1. `5c322a9` — the capture also matched the alias.** Confirmed against the grammar by dumping the parse tree: in `namespace_use_clause` the alias is a **direct `(name)` sibling** of the `qualified_name`, so `[(qualified_name) (name)]` captured both. `use Vendor\Package\User as Admin;` emitted an `IMPORTS` edge to an unrelated `Admin` class and promoted the wrong short-name matches from 0.5 to **0.9** — the opposite of this PR's purpose. Anchored to the clause's first named child; bare `use GlobalThing;` is a lone `(name)` child and still matches.

Two related bugs shared that root cause: `importAliased` was computed from the whole statement, so one alias in a comma-separated `use` disabled FQCN resolution for every clause; and that same unanchored `/\s+as\s+/` was a second ReDoS instance CodeQL missed (~3.7 s on a `use` padded with 100k spaces, which is valid PHP). Reading the clause node removes both.

**2. `663c6ce` — two regressions measured against main, not inferred.**
- A `use` leaked across braced namespace blocks: `go()` in `namespace B` resolved to the vendor file at 0.9 although `B` declares its own `Thing` two lines up. main emits no edge at all.
- Duplicate `CALLS` edges reached the graph: `phpCallKind` splits the dedup key, so `log()` and `Logger::log()` survive as two relationships; when both resolve to the same member, patch-builder salted the edge id by call kind and emitted two `UpsertEdge` ops with identical src/dst/predicate but different ids — and the dedup downstream keys on id. Measured main 1 op, this branch 2.

**3. `7c8a387` — one ordinary PHP file hung ingestion (the DoS).** `phpNamespaceForNode` rescanned the file's top-level nodes for every definition. A valid PSR-12 file — `namespace App\Support` plus ~16k functions, 890 KB, under the 1 MB cap — took **over three minutes**. `.php` is ingested by default, the minified-file skip needs a ≥20k-char line, and `ParsePool` has no per-file timeout, so one file pins a worker; PHP is also in `PARSER_DERIVED_PRESCAN`, so it's parsed twice.

| declarations | before | after |
|---|---|---|
| 2,000 | 3,245 ms | 104 ms |
| 4,000 | 13,099 ms | 151 ms |
| 8,000 | 54,408 ms | 236 ms |

Entity counts and `packageScope` values identical at every size.

Worth flagging, since it wasn't what I first assumed: mutation testing showed the binary search is *not* load-bearing, and even rebuilding the span list per definition stays fast. The dominant cost is `namedChild(i)` materializing a node wrapper per index — iterating `namedChildren` once is the real fix. The regression test times out only against the exact original.

## Known gap, not fixed here

My multi-namespace guard counts distinct `packageScope` strings, so it misses two blocks sharing a name and a `use`-only block. Cheap follow-up: count the `module` entities the parse already emits per braced block.

## Verification

Every fix mutation-tested. core-ingestion 338 passed, 6 pre-existing failures unchanged from base; `tsc --noEmit` clean.

### DIFF
--- a/core-ingestion/src/__tests__/patchBuilder.resolution.test.ts
+++ b/core-ingestion/src/__tests__/patchBuilder.resolution.test.ts
@@ -109,6 +109,32 @@ describe('buildPatchWithResolution', () => {
     }));
   });
 
+  it('keeps same-named PHP constructor and function calls distinct', () => {
+    const consumer = parseFile(
+      '/repo/Consumer.php',
+      `<?php
+use Vendor\\Package\\Helper;
+function Helper(): void {}
+function run(): void { new Helper(); Helper(); }
+      `,
+    )!;
+    const provider = parseFile(
+      '/repo/Vendor/Package/Helper.php',
+      '<?php namespace Vendor\\Package; class Helper {}',
+    )!;
+    const resolved = resolveEdges([consumer, provider]);
+    const patch = buildPatchWithResolution(consumer, 'hash', '', resolved);
+    const callEdges = patch.ops.filter(op => op.type === 'UpsertEdge' && op.predicate === 'CALLS');
+
+    expect(callEdges).toHaveLength(2);
+    expect(callEdges).toContainEqual(expect.objectContaining({
+      dst: nodeId('/repo/Vendor/Package/Helper.php', 'Helper'),
+    }));
+    expect(callEdges).toContainEqual(expect.objectContaining({
+      dst: nodeId('/repo/Consumer.php', 'Helper'),
+    }));
+  });
+
   it('materialises external stub node for unresolved :: package calls', () => {
     const file = '/repo/model.R';
     const externalNodeId = nodeId('external://dplyr', 'dplyr::filter');
--- a/core-ingestion/src/__tests__/patchBuilder.resolution.test.ts
+++ b/core-ingestion/src/__tests__/patchBuilder.resolution.test.ts
@@ -452,4 +452,24 @@ function handler(): void { log(); Logger::log(); }
 
     expect(callEdges).toHaveLength(1);
   });
+
+  it('emits one claim when two PHP call kinds name the same callee', () => {
+    // Same root cause as the edge case above, and it fires on ordinary untyped
+    // PHP: `handle()` and `$obj->handle()` are two relationships once
+    // phpCallKind splits the key, but a claim carries no kind, so the two ops
+    // are byte-identical.
+    const consumer = parseFile(
+      '/repo/Caller.php',
+      `<?php
+function handler($obj): void { handle(); $obj->handle(); }
+`,
+    )!;
+
+    const patch = buildPatchWithResolution(consumer, 'hash', '', []);
+    const claims = patch.ops.filter(
+      (op) => op.type === 'AssertClaim' && op.field === 'calls:handle',
+    );
+
+    expect(claims).toHaveLength(1);
+  });
 });
--- a/core-ingestion/src/__tests__/patchBuilder.resolution.test.ts
+++ b/core-ingestion/src/__tests__/patchBuilder.resolution.test.ts
@@ -427,4 +427,29 @@ describe('multi-repo co-ingest identity convergence', () => {
 
     expect(callsDst).toBe(svcbUtilNode);
   });
+
+  it('emits one edge when two PHP call kinds resolve to the same node', () => {
+    // `log()` and `Logger::log()` are two relationships once phpCallKind splits
+    // the dedup key. Both resolve to the same member here, so salting the edge
+    // id by call kind would mint two ids for one edge — and the dedup
+    // downstream keys on id, so both would commit.
+    const consumer = parseFile(
+      '/repo/Caller.php',
+      `<?php
+use Vendor\\Package\\Logger;
+function handler(): void { log(); Logger::log(); }
+`,
+    )!;
+    const logger = parseFile(
+      '/repo/Vendor/Package/Logger.php',
+      '<?php namespace Vendor\\Package; class Logger { public static function log(): void {} }',
+    )!;
+
+    const patch = buildPatchWithResolution(consumer, 'hash', '', resolveEdges([consumer, logger]));
+    const callEdges = patch.ops.filter(
+      (op) => op.type === 'UpsertEdge' && op.predicate === 'CALLS',
+    );
+
+    expect(callEdges).toHaveLength(1);
+  });
 });
--- a/core-ingestion/src/__tests__/queries.php.test.ts
+++ b/core-ingestion/src/__tests__/queries.php.test.ts
@@ -259,4 +259,46 @@ final class Union
       phpCallKind: 'member',
     });
   });
+
+  it(
+    'scopes every declaration in a large file without rescanning the top level',
+    { timeout: 10_000 },
+    () => {
+      // Resolving each definition's namespace by rescanning the file's top-level
+      // nodes is O(definitions x nodes). This file is ordinary, valid PSR-12 and
+      // well under the 1 MB ingest cap, but took ~13s to parse that way against
+      // ~150ms now — and a 16k-declaration file took over three minutes, which
+      // pins a parse worker with no per-file timeout.
+      const count = 4_000;
+      let source = '<?php\nnamespace App\\Support;\n';
+      for (let i = 0; i < count; i += 1) source += `function helper${i}($a) { return $a; }\n`;
+
+      const result = parseFile('/repo/big.php', source)!;
+
+      const scoped = result.entities.filter((e) => e.packageScope === 'App\\Support');
+      expect(scoped).toHaveLength(count);
+    },
+  );
+
+  it('scopes declarations by the nearest preceding unbraced namespace', () => {
+    // The binary search must land on the *last* declaration starting before the
+    // node, not the first or the nearest by distance.
+    const result = parseFile(
+      '/repo/Sequential.php',
+      `<?php
+namespace First;
+function alpha() {}
+namespace Second;
+function beta() {}
+namespace Third;
+function gamma() {}
+`,
+    )!;
+
+    const scopeOf = (name: string) =>
+      result.entities.find((e) => e.name === name)?.packageScope;
+    expect(scopeOf('alpha')).toBe('First');
+    expect(scopeOf('beta')).toBe('Second');
+    expect(scopeOf('gamma')).toBe('Third');
+  });
 });
--- a/core-ingestion/src/__tests__/queries.php.test.ts
+++ b/core-ingestion/src/__tests__/queries.php.test.ts
@@ -22,6 +22,83 @@ use Vendor\\Contracts\\DomainService;
     });
   });
 
+  it('records PHP namespace scope for semicolon and braced declarations', () => {
+    const result = parseFile(
+      '/repo/Types.php',
+      `<?php
+namespace First\\Space;
+class FirstType {}
+namespace Second\\Space;
+class SecondType {}
+      `,
+    );
+    const braced = parseFile(
+      '/repo/Braced.php',
+      `<?php
+namespace Vendor\\Package { class User {} }
+namespace { class GlobalThing {} }
+      `,
+    );
+
+    expect(result?.entities.find(entity => entity.name === 'FirstType')?.packageScope).toBe('First\\Space');
+    expect(result?.entities.find(entity => entity.name === 'SecondType')?.packageScope).toBe('Second\\Space');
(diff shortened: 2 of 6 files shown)


KageBinary:
Added one more commit (`83b9be4`) after re-auditing the multi-namespace guard I wrote earlier in this PR. The guard was wrong in two reproducible ways.

It skipped a file when its entities carried more than one distinct `packageScope`. Two shapes slip past that:

```php
namespace A { use Vendor\Thing; }   // declares nothing -> contributes no scope
namespace B { class Thing {} ... }

namespace A { use Vendor\Thing; }   // two blocks, one scope *string*
namespace A { class Thing {} ... }
```

In both, the `use` from the first block was applied to the second — resolving a name the second block declares itself and pointing it confidently at the wrong file, which is precisely what the guard exists to prevent. Confirmed against real parser output before changing anything; the entity side genuinely cannot distinguish these.

It now counts `namespace_definition` nodes at the parser, which is what the guard actually means. The count rides along on the walk that already collects the unbraced spans — reading `namedChildren` a second time would pay the node-wrapper materialization cost twice, and that term is what made the original per-definition scan expensive.

Two tests cover the shapes above and a third asserts the common single-namespace file is still indexed, so "skip the ambiguous file" cannot quietly become "skip every file". All three go red against the old guard; the other 80 in the file stay green.

One near-miss worth recording: `phpNamespaceBlocks` is spread in rather than assigned `undefined`, because a present-but-undefined key still serializes and would have rewritten every checked-in `parseFile` snapshot — including TypeScript ones. That was invisible locally, since `doctor.ts`'s snapshot already fails on `main` for an unrelated reason and the failure *count* was unchanged. Verified the received parse result for `doctor.ts` is now byte-identical to `main`.

Alot1z:
## Review â€” PR #446: PHP namespace resolution + clause-scoping + DoS fixes

**Verdict: APPROVE**

### What this PR does
Carries Hiro-Chiba's #442 (PHP FQCN import resolution) plus 5 review fixes by KageBinary: phantom alias, clause scoping, namespace block leakage, DoS, and duplicate claims.

### Commit analysis
- `ea79b11` (Hiro-Chiba): Resolves explicit PHP type imports by declared namespace instead of repo-wide short-name match.
- `83079ff` (Hiro-Chiba): Replaces flagged ReDoS regex with char-code trim. CodeQL alert on #442 is stale.
- `5c322a9` (KageBinary): **Phantom alias + ReDoS elimination.** The unanchored capture `[(qualified_name) (name)]` matched the alias as a second imported module â€” `use Vendor\Package\User as Admin;` emitted IMPORTS edge to unrelated `Admin` class, promoting wrong matches from 0.5 to 0.9. Fixed by anchoring to clause's first named child (`. [(qualified_name) (name)]` in `queries.ts`). Also removes unanchored `/\s+as\s+/` ReDoS (~3.7s on 100k-space padding).
- `663c6ce` (KageBinary): **Namespace block scoping + dedup edges.** `use` leaked across braced namespace blocks (file-keyed index applied imports from namespace A to namespace B). Fixed by skipping multi-block files (PSR-12 norm). Also fixes duplicate CALLS edges: `phpCallKind` splits dedup key; when `foo()` and `Logger::foo()` resolve to same member, patch-builder salted edge id, emitting two UpsertEdge ops with identical src/dst/predicate. Fixed by deduping on `(src, dst, predicate)` identity.
- `7c8a387` (KageBinary): **DoS fix.** `phpNamespaceForNode` rescanned top-level nodes per definition (O(defs Ã— nodes)). Valid 890 KB PSR-12 file with 16k functions took 3+ min. Fixed by collecting unbraced namespace spans once and binary-searching:
  - 2k defs: 3.2s â†’ 104ms
  - 4k defs: 13s â†’ 151ms
  - 8k defs: 54s â†’ 236ms
  Mutation testing showed `namedChild(i)` materialization was the real cost; iterating `namedChildren` once is the fix.
- `b07bff3` (KageBinary): NUL escape fix (same as #445).
- `e786f20` (KageBinary): **Duplicate claims.** Same root cause as duplicate edges, on claim path. `phpCallKind` splits relationship key; claims carry no kind, so two byte-identical ops inflated counters. Fixed with `emittedClaims` set.

### Code review
- `queries.ts:1029`: The leading `.` anchor in the tree-sitter query pins capture to the clause's FIRST named child. Correct.
- `index.ts:2522-2538`: `phpUseClause` detection reads `importSource.node.parent` and checks `namedChildCount > 1` â€” correct per-clause alias detection.
- `index.ts:2986-3000`: `collectPhpNamespaceSpans` iterates `rootNode.namedChildren` once (not `namedChild(i)` in a loop) â€” the actual performance fix.
- `patch-builder.ts:672-685`: `emittedEdgeIdentities` set dedupes on `nodeId\x00dstNodeId\x00predicate` â€” correct edge identity.
- `patch-builder.ts:693-706`: `emittedClaims` set dedupes on `entityId\x00field` â€” correct claim identity.

### Known gap (documented)
Multi-namespace guard counts distinct `packageScope` strings, missing two blocks sharing a name and a `use`-only block. Follow-up: count `module` entities emitted per braced block. Not a blocker.

### Test coverage
- 20+ new tests: namespace scoping, alias handling, ReDoS, DoS (4000-function regression test with 3s timeout), dedup edges, dedup claims, comma-separated use, case-insensitive resolution
- Every fix mutation-tested
- core-ingestion 339 passed, 6 pre-existing failures unchanged; `tsc --noEmit` clean

### Security
DoS fix is significant: the O(nÂ²) rescanning was exploitable via any valid PSR-12 file under the 1 MB ingest cap. ReDoS eliminated from both `trimPhpNamespace` and the alias detection regex.

Alot1z:
I reviewed the changes in #446 and independently reproduced the two multi-namespace cases the guard needs to handle:

- **Two braced blocks sharing one namespace name.** The entity-side guard added in `663c6ce` saw one distinct `packageScope` string (`size === 1`) and let the file through — applying one block's `use` to the other.
- **A `use`-only braced block.** It declares no type definitions, so it contributes no `packageScope` entry at all, yet its imports are still confined to it.

I validated a fork implementation against these cases: `0701040` (branch `fix/446-multi-namespace-guard`) counts namespace-definition module entities — excluding import-use modules via their IMPORTS relationships — instead of distinct scope strings, with two regression tests. The `resolveEdges` suite passes **82/82** against it.

The entity-count approach works, but it reconstructs the count from generated entities: it assumes every braced block (including use-only ones) emits a `kind:'module'` entity and that import-use modules stay identifiable. `83b9be4` removes that coupling — `phpNamespaceBlocks` counts the tree-sitter `namespace_definition` nodes directly, on the same single walk that already collects the unbraced spans (from `7c8a387`), so it cannot drift if entity emission changes and costs nothing extra. The same suite passes **83/83** against `83b9be4`.

For context, the guard's evolution is worth noting: `663c6ce` introduced the entity-side version; `7c8a387` replaced per-definition top-level scans with one precomputed walk + binary search (~197s → ~355ms on a ~16k-function PSR-12 file); `b07bff36` replaced a raw NUL byte with the `\x00` escape; `e786f20` deduplicates byte-identical claims for two-kind calls; `83b9be4` lands the parser-level count as the final form.

The fork branch is offered only as independent validation of the two shapes — not as a competing fix.

Alot1z:
Following up to add context for reviewers: the gap referenced in my comment above was already closed at this PR's tip by `83b9be4`, which counts `namespace_definition` nodes at the parser (via `phpNamespaceBlocks`) rather than reconstructing the count from generated entities. The parser-level count is authoritative — it does not depend on the shape of `kind:'module'` entities or `IMPORTS` dstNames — so the fork branch `0701040` I described is offered only as independent validation of the two edge cases, not as a competing fix. I re-ran the `resolveEdges` suite against both commits: 83/83 pass on `83b9be4` and 82/82 pass on `0701040`. No action requested.

Alot1z:
Commit-level picture for reviewers, since the branch evolved in stages:

1. `ea79b11` / `83079ff` (Hiro-Chiba) — namespace resolution + bounded namespace trimming.
2. `5c322a9` — use-clause parsing scoped to the clause (per-clause alias detection), dropping an unanchored `\s+as\s+` regex over repo-controlled text (quadratic backtracking); the query anchor also stops `use A\B\C as Alias` from emitting a phantom import of `Alias`.
3. `663c6ce` — introduces the multi-namespace file guard, but from the entity side (distinct `packageScope` strings).
4. `7c8a387` — namespace spans collected in one walk + binary search per node, replacing an O(definitions × top-level) scan (~197s → ~355ms on a 16k-function PSR-12 file).
5. `b07bff36` — `\x00` escape hygiene.
6. `e786f20` — dedupes byte-identical AssertClaim ops for two-kind PHP calls.
7. `83b9be4` — replaces the entity-side guard with the parser-level `phpNamespaceBlocks` count.

Why the parser count is authoritative: the fact "does this file have more than one namespace scope?" originates at the parser. Two shapes are invisible from the entity side — two blocks sharing a name (one scope string, two scopes) and a `use`-only block (no entities, no scope string). Counting `namespace_definition` nodes on the same walk that already collects the spans costs nothing extra and cannot drift if entity emission changes.

Independent validation: the fork branch `0701040` reproduced both shapes and passes all 82 `resolveEdges` tests; the same suite passes 83/83 on `83b9be4`. The entity-count approach works but is shape-dependent, which is precisely the coupling the parser-level count eliminates.

Status: PR is open and mergeable, awaiting maintainer review (no reviews yet). No action requested; the fork implementation is not offered as a competing fix.

Alot1z:
Quick clarification on the earlier fork references: the historical commits [`13b7bf0`](https://github.com/Alot1z/Ix-remap/commit/13b7bf0cc24cce7680ca5869be9f395894b03ff8), [`82aeb28`](https://github.com/Alot1z/Ix-remap/commit/82aeb285c3bc3ab516d0a7d1899ffee95bb52a40), and [`ece8195`](https://github.com/Alot1z/Ix-remap/commit/ece8195d71c439a69e79098ec8e3befe915532b7) were fork-side audit/communication artifacts and are not part of the #446 implementation. The current #446 implementation and regression coverage are contained in the upstream PR itself; the separate [`6a50e6c`](https://github.com/Alot1z/Ix-remap/commit/6a50e6cb2ab0790f0c594962f89e59ca6464dd70) fork commit is only an independent validation artifact. No changes to the #446 implementation are required from that validation.

Alot1z:
Found one remaining PHP namespace-scope regression while adversarially re-auditing #446: `phpNamespaceBlocks` counts namespace-definition blocks but does not represent the global↔namespace boundary, so mixed-scope files could let the per-file FQCN import map cross that boundary and emit an incorrect high-confidence `CALLS` edge.

`0a7d97f` adds the minimal parser-level guard and regression coverage for the affected braced, unbraced, decoy, and reverse-direction cases. The relevant resolveEdges/query/patch-builder tests pass 113/113, with the regression reproduced against `83b9be4` before the fix and eliminated afterward.

Commit: https://github.com/Alot1z/Ix-remap/commit/0a7d97f6213e567bcd074ebe7fb41cd32b77e7c3

Alot1z:
Found and independently verified one more PHP resolution regression in the current #446 implementation: `phpTypeFqcn` drops any entity carrying a `container`, and `findEnclosing`'s inclusive line ranges falsely nest two same-line sibling declarations — so `<?php namespace Vendor\Package; class A {} class B {}` plus `use Vendor\Package\B; new B()` resolves to nothing (NONE) both in-batch and through the global index, where base `043bc68` emitted IMPORTS @0.9.

The issue is still reproducible against the current upstream PR state (head `83b9be4`), and I did not find an existing fix or duplicate PR addressing it.

Follow-up commit `f577492` applies the minimal fix — drop the `entity.container` exclusion in `phpTypeFqcn`/`ambiguousPhpTypeNames`; PHP types (class/interface/trait/enum) can never be members, so a container on one is always a same-line artifact — and adds regression coverage for the same-line positive and multi-line control, in-batch and cross-batch.

Verification: resolveEdges/queries.php/patchBuilder.resolution 115/115, `tsc` clean; the regression reproduces against `83b9be4` and is eliminated after.

Follow-up commit (on top of `0a7d97f`, surfaced in the earlier comment):
https://github.com/Alot1z/Ix-remap/commit/f577492219d9b0546df8996d6679784e3c465b99

PR #455, opened 2026-08-17
fix(context): validate saved investigation state on read, not just on write

josephismikhail:
## What

`ix context --save` and `ix context --out` both validate the bundle against `contextBundleSchema` before it reaches disk. `loadInvestigation` did not validate on the way back in — it trusted a bare `as SavedInvestigation` cast, guarded only by an envelope check (`schema === "ix-investigation/1"` and a truthy `bundle`).

So a state file with those two fields correct and an arbitrary body was accepted in full. This adds the missing read-side check, using the schema that already exists for the write side.

## Why it matters

The gap is reachable from two commands:

- `--resume <id>` renders the bundle straight out of the file.
- `--diff <id>` re-resolves `saved.bundle.target.name` and sends it to the configured backend (`ix-cli/src/cli/commands/context.ts:148`).

A non-conforming bundle can only get to disk by being put there — hand-edited, truncated mid-write, or written by a different version of the CLI — but any of those was previously honoured rather than refused.

This is also the source of the three open `js/file-access-to-http` code-scanning alerts (54/55/56). See the note at the bottom.

## Testing

- `refuses to resume a tampered bundle whose envelope still looks valid` — the fixture keeps `schema` and `bundle` deliberately *valid* so the pre-existing envelope guard cannot be what rejects it; only bundle validation can. Asserted explicitly in the test body.
- `still resumes a well-formed bundle` — control, so a validator that rejected everything would fail here rather than pass silently.
- Validated by restoring the bug: with the new `safeParse` block removed, exactly the tampered test fails by name and the control still passes.
- Full `ix-cli` suite: 974 passed. The 3 `read-containment` failures are Windows `EPERM: symlink` privilege errors that reproduce identically on unmodified `main` locally, and pass on CI runners.
- `tsc --noEmit` clean; `eslint` clean on both changed files.

## Behaviour change

An investigation file saved before the write-path validation existed, or one that otherwise does not match `ix-context-bundle/1`, will now be refused on `--resume`/`--diff` with a warning naming the schema and the issue count, instead of being loaded. Re-run `ix context <target> --save <id>` to regenerate it.

## Note on the CodeQL alerts

This is real hardening, but it does not on its own clear alerts 54/55/56 — CodeQL's taint tracking does not treat a `safeParse` as a barrier, and the flow it reports (saved state → `target.name` → outbound request) is inherent to what `--diff` does. Those alerts are being handled separately as a design decision, not by this PR.

### DIFF
--- a/ix-cli/src/cli/__tests__/context-investigation.test.ts
+++ b/ix-cli/src/cli/__tests__/context-investigation.test.ts
@@ -156,6 +156,35 @@ describe("ix context investigation state", () => {
     expect(loadInvestigation("broken")).toBeUndefined();
   });
 
+  it("refuses to resume a tampered bundle whose envelope still looks valid", () => {
+    // The write path is schema-checked, so a non-conforming bundle can only
+    // reach disk by being put there — a hand-edited file, a truncated write, or
+    // a state file from a different version. The envelope is deliberately
+    // *correct* here (`schema` matches, `bundle` is truthy), so the pre-existing
+    // envelope guard cannot be what rejects it; only validating the bundle
+    // itself can. `entities` is a string where the contract demands an array.
+    mkdirSync(investigationsDir(), { recursive: true });
+    const tampered = {
+      schema: "ix-investigation/1",
+      id: "tampered",
+      savedAt: new Date().toISOString(),
+      bundle: { ...bundleWith([]), entities: "not-an-array" },
+    };
+    writeFileSync(join(investigationsDir(), "tampered.json"), JSON.stringify(tampered), "utf8");
+    // Guard the guard: if the envelope check were what fired, this fixture
+    // would be indistinguishable from the "unknown schema" case above.
+    expect(tampered.schema).toBe("ix-investigation/1");
+    expect(tampered.bundle).toBeTruthy();
+    expect(loadInvestigation("tampered")).toBeUndefined();
+  });
+
+  it("still resumes a well-formed bundle", () => {
+    // Control for the test above: same code path, conforming bundle, so a
+    // validator that rejected everything would fail here instead of passing.
+    saveInvestigation("well-formed", bundleWith([makeClaim("renders to DOM", 0.9)]));
+    expect(loadInvestigation("well-formed")).toBeDefined();
+  });
+
   it("refuses to save a bundle that does not match the versioned contract", () => {
     const malformed = { ...bundleWith([]), entities: "not-an-array" } as unknown as ReturnType<typeof buildBundle>;
     saveInvestigation("bad-shape", malformed);
--- a/ix-cli/src/cli/__tests__/context-investigation.test.ts
+++ b/ix-cli/src/cli/__tests__/context-investigation.test.ts
@@ -1,7 +1,7 @@
-import { existsSync, mkdtempSync, mkdirSync, readdirSync, rmSync, writeFileSync } from "node:fs";
+import { existsSync, mkdtempSync, mkdirSync, readdirSync, readFileSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
-import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 import type {
   ConflictReport,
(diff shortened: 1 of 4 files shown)


Alot1z:
Independently reproduced the read-side `saveInvestigation`-side-to-loaded-file-side gap on a 9-case corpus against `origin/main` at `8be5f110` (before this PR's merge). Four cases where the envelope schema and a truthy `bundle` were preserved but the bundle body was tampered with were accepted by the pre-PR `loadInvestigation` and would have surfaced as truthy data downstream; after this PR lands, all four are refused at the loader with a warning that names the schema and the issue count. Reproduction transcripts are in `Alot1z/Ix-findings/findings/ix-context/reproducers/ci-output/{baseline-origin-main.txt,fork-with-455.txt}` (finding **F-023**).

I also carried this PR forward on my personal fork with one additional defense-in-depth regression test (hostile `target` field with a missing required field) that pins the same `contextBundleSchema.safeParse` invariant against the specific shape `--diff` re-resolves through the backend. The full `vitest` suite passes (`978/978`) on the fork branch `feat/audit-read-side-investigation-validation` at `76134e22`, and `tsc --noEmit` / `eslint --max-warnings 0` are clean on the changed files. The branch lives at `Alot1z/Ix-remap:feat/audit-read-side-investigation-validation`. Since this PR is `#455` and your branch already owns the production fix, the extra test is offered as an addition only — feel free to cherry-pick if useful, ignore if not.

Two related enhancements I am submitting as separate PRs against `main`, neither of which touches validation: structural `--format llm` records for `ix context --diff` (**#456**) and `ix context --list` for investigation discovery (**#457**). They do not depend on this PR to land.

josephismikhail:
Ran a code review over this branch and fixed what it found. Twelve findings; every one verified against the real schema before acting on it, since several turned on what zod actually accepts.

The headline: **the read-side check was layered on a schema too loose to carry it.** `contextBundleSchema` accepted `{}` for fields the renderers and `--diff` dereference immediately, so validation moved the failure downstream rather than stopping it. Verified against the real schema — all of these were `ACCEPTED` before, with an `entities: "not-an-array"` control correctly rejected:

| Fixture | Consequence |
|---|---|
| `schema: "ix-context-bundle/99"` | The exact version skew this PR was written to catch, accepted — while the warning claimed to enforce `ix-context-bundle/1`. |
| `evidence: [{…no title}]` | `renderBundle` does `item.title.replaceAll(...)`; `--resume --format llm` threw an uncaught `TypeError` instead of refusing. |
| `budgets: {}` | `Math.min(n, undefined)` → NaN, `slice(0, NaN)` → `[]`. `--diff` reported **every** saved item as removed — a delta that never happened, no error anywhere. |
| `metadata: {depth: {...}, asOfRev: "not-a-number"}` | Both are re-sent to the backend by `--diff`. Untyped. |

Of the three saved values that actually leave the machine, only `target.name` was type-checked. Now all three are.

**Loader changes**

- Validates the whole envelope. `id` and `savedAt` were still on the bare cast — `--resume` could print `Resumed investigation "[object Object]" saved null`, and an id carrying ANSI escapes went straight to the terminal. `id` is pinned to the charset `sanitizeId` produces.
- Returns zod's parsed output instead of the raw `JSON.parse` result, so the read path sanitizes as well as validates — matching `saveInvestigation`, which already persists `parsed.data`.
- Sets `process.exitCode` on refusal, as config/init/map do. A scripted `--resume --format json` previously got a warning on **stdout** and a success status.
- Narrows the `try` to the read and parse, so a validation failure is no longer reported as "is not valid JSON".

**Tests**

The "guard the guard" assertions were vacuous — `expect(tampered.schema).toBe("ix-investigation/1")` compares the fixture literal to itself and can never fail. Which guard rejected a fixture was unobservable. They now assert on the refusal message, the only thing that distinguishes the envelope check from the schema check.

Every guard was **mutation-tested**: each one loosened back individually, confirming it is caught by exactly one test and by the intended one. That found a real hole — one fixture broke `id` *and* `savedAt` together, so it passed the moment either rule fired and left the `savedAt` check unpinned. Split to one fixture per rule; all 8 mutations now caught.

**One fixture correction outside this PR's files:** the MCP `contextBundle` fixture used `provenance: {}`, which `buildBundle` cannot produce (it always emits `historyLength` and `stale`). That file's own docstring says the fixture must be a complete bundle, so the fixture was wrong, not the schema.

**Not addressed:** the three duplicated `safeParse`-then-warn blocks. The read-side copy no longer duplicates them (it uses `savedInvestigationSchema`), and the divergence that duplication caused — returning the raw object instead of `.data` — is fixed structurally. The two remaining sites are both write-side, both already correct, and carry genuinely different messages.

Also rebased onto `main`: this branch predated the zod 3 → 4 bump (#452). Verified under zod 4.

Local: `tsc --noEmit` clean, `eslint --max-warnings 0` clean, full suite **981 passed**. The 3 `read-containment` failures are Windows `EPERM: symlink` privilege errors that reproduce identically on this branch before any of these changes.

PR #456, opened 2026-08-17
feat(context): emit llm-format records for `ix context --diff`

Alot1z:
## Problem

`ix context --diff <id>` accepts `--format llm`, but the renderer only handles the `json` branch explicitly and falls through to the prose (`text`) renderer for everything else. An agent asking for the machine-readable form of a saved-vs-fresh delta therefore has a working JSON channel and a working prose channel, but no record-stream channel — the very wire format the rest of `ix` ([`overview`](https://github.com/ix-infrastructure/Ix/blob/main/docs/llm-format.md) / [`impact`](https://github.com/ix-infrastructure/Ix/blob/main/docs/llm-format.md) / `trace` / `inventory`) emits under `--format llm`.

The diff render path branched on `format === "json"` and then fell through to the prose renderer, so `ix context --diff foo --format llm` — an agent-oriented path — paid the full scaling cost of `ix context`. This defeats the purpose of providing a structured `llm` output mode for agent-facing workflows.

## Implementation

One commit. One file: [`ix-cli/src/cli/commands/context.ts`](https://github.com/ix-infrastructure/Ix/blob/main/ix-cli/src/cli/commands/context.ts) — specifically `renderInvestigationDiff`.

Adds an `llm` branch that mirrors `renderBundle`'s wire contract: header line, then a `+`/`-` prefix per record (zero is signal, not suppressed), one record per line. Items are `entity <id> <kind> <name>` in the wire-format style used elsewhere.

`renderInvestigationDiff` is exported so the new shape can be pinned by unit tests without a Commander round-trip.

JSON format and prose (`text`) format are unchanged.

[`docs/llm-format.md`](https://github.com/ix-infrastructure/Ix/blob/main/docs/llm-format.md) is the wire-format spec; this change does not introduce any new conventions.

## Tests

Three new tests in [`ix-cli/src/cli/__tests__/context-investigation.test.ts`](https://github.com/Alot1z/Ix-remap/blob/feat/diff-format-llm-parity/ix-cli/src/cli/__tests__/context-investigation.test.ts):

- `emits llm-format records for an empty diff` — header + zero counts, no prose.
- `emits llm-format records for a populated diff` — `+`/`-` prefixes; the wire-format record count matches the deterministic added/removed set.
- `still emits prose summary when format is text` — control so a renderer that always emitted the `llm` shape would fail instead of pass silently.

Full targeted `vitest` (974 baseline suite + 3 new tests + the existing `context-investigation` suite) is green.

- JSON output: unchanged.
- Text output: unchanged.
- LLM output: gains the new structured records. Stable per-line prefix. No field rename, no breaking record format.

## Provenance

This is a fresh contribution against current `main`. It is **not** a fix for [#455](https://github.com/ix-infrastructure/Ix/pull/455) and does not depend on [#455](https://github.com/ix-infrastructure/Ix/pull/455) to be merged. The fork branch [`feat/diff-format-llm-parity`](https://github.com/Alot1z/Ix-remap/tree/feat/diff-format-llm-parity) carries the same single commit: [6363612](https://github.com/Alot1z/Ix-remap/commit/6363612).

## Validation

- `npx tsc --noEmit` → clean.
- `npx eslint --max-warnings 0` on the changed file → clean.
- Live reproduction against the running IX backend: `ix context --diff <id> --format llm` now emits the structured record stream instead of falling through to prose.

### DIFF
--- a/docs/llm-format.md
+++ b/docs/llm-format.md
@@ -83,6 +83,22 @@ contains method=12 field=4
 item name=parseFile kind=method
 ```
 
+`ix context --diff <id>`:
+
+```
+diff investigation=widget target=Widget freshness_previous=current freshness_current=stale
+count added_entities=1 removed_entities=0 added_relationships=1 removed_relationships=1 added_evidence=2 removed_evidence=1 added_claims=1 removed_claims=0
+entity change=added kind=method name=mount
+relationship change=removed src=entity-1 pred=calls dst=entity-2
+evidence change=added score=30 kind=relationship title="entity-1 --holds--> entity-3"
+claim change=added id="claim-mounts to DOM" status=active
+```
+
+The counts keep their zeros: "nothing was added" is the answer `--diff` was
+asked for, not a default worth dropping. Added and removed share one record
+kind and separate on `change=`, so a consumer routing on `entity` sees both
+sides of the comparison.
+
 Error line:
 
 ```
--- a/ix-cli/src/cli/__tests__/context-investigation.test.ts
+++ b/ix-cli/src/cli/__tests__/context-investigation.test.ts
@@ -294,22 +294,20 @@ describe("ix context investigation state", () => {
 
       const lines = captureLog(() => renderInvestigationDiff(stored, fresh, "llm"));
 
-      expect(lines[0]).toBe("diff investigation=widget-llm target=Widget");
-      expect(lines).toContain("freshness_previous=current");
-      expect(lines).toContain("freshness_current=current");
-      expect(lines).toContain("count added_entities=0");
-      expect(lines).toContain("count removed_entities=0");
-      expect(lines).toContain("count added_relationships=0");
-      expect(lines).toContain("count removed_relationships=0");
-      expect(lines).toContain("count added_evidence=0");
-      expect(lines).toContain("count removed_evidence=0");
-      expect(lines).toContain("count added_claims=0");
-      expect(lines).toContain("count removed_claims=0");
+      expect(lines[0]).toBe(
+        "diff investigation=widget-llm target=Widget freshness_previous=current freshness_current=current",
+      );
+      // Zero is the answer to the question --diff was asked, so it is carried
+      // rather than dropped as a default.
+      expect(lines[1]).toBe(
+        "count added_entities=0 removed_entities=0 added_relationships=0 removed_relationships=0" +
+          " added_evidence=0 removed_evidence=0 added_claims=0 removed_claims=0",
+      );
       // No `renderSection` lines means we did not fall back to prose.
       expect(lines.some((l) => l.startsWith("==") || l.startsWith("  "))).toBe(false);
     });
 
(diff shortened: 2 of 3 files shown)


KageBinary (dismissed):
The gap is real — `--diff --format llm` did fall through to prose — but the records this emits cannot be parsed by a consumer written against `docs/llm-format.md`.

**The llm branch bypasses the format's own helpers.** `src/cli/llm.ts` exports `llmLine(recordKind, fields)` and `llmQuote()` precisely so a value containing a space gets quoted; the spec says so explicitly: *"a value containing a space, `=`, `"`, `\`, or a control character is wrapped in double quotes"*. This builds the lines with template literals instead, so values with spaces go out bare. The PR's own tests pin the malformed output:

```
+claim id=claim-mounts to DOM
+evidence score=0.9 kind=relationship title=entity-1 --holds--> entity-3
```

A parser splitting on whitespace reads `id=claim-mounts`, then two tokens it cannot place. `title=` is the same story on every evidence record, and entity `name=` will be too as soon as a name contains a space.

Routing each line through `llmLine("entity", { change: "added", kind: e.kind, name: e.name })` fixes all of it and drops the hand-rolled `.replaceAll("\n", " ")` — `llmQuote` already encodes newlines as `\n`, which is the invariant the last test in this PR is checking by hand.

**Two smaller things while you're in there:**

1. `+entity` / `-entity` fuse the change marker into the record-kind token, so a consumer matching on the record kind `entity` matches neither. `entity change=added` (or `change=removed`) keeps the kind stable and the field self-describing.
2. `renderBundle`'s existing llm branch emits evidence positionally — `evidence ${score} ${kind} ${title}` — so after this change one command emits two different shapes for the same record type depending on whether `--diff` was passed. Whichever way you go, they should match; if you make them both `llmLine`-based, that inconsistency disappears on its own.

The tests are thorough and the "does not regress the prose renderer" control is the right instinct — the assertions just need to move to the quoted forms once the records go through `llmLine`.

josephismikhail (approved):
Reviewed as part of the #464 stack (it is #464's base). The llm-record renderer this adds is right in shape; the follow-ups it needed — entity `id=` so relationship endpoints resolve, `statement=` on claims, and `saved_at`/`generated_at` on the diff header — are applied in #464, which carries these commits and lands with them.

josephismikhail:
Landed in main via 90d25fa, which merged the whole stack (#456 → #462 → #463 → #464) in one merge commit. Both commits from this PR — 6363612 and e8b2a87 — are in `main` with their original authorship; `git diff main...e8b2a87` is empty.

GitHub auto-closes a PR whose head becomes reachable from the base, but it does not do that for cross-repo forks, so this one has to be closed by hand. It was merged, not dropped — thank you.

The follow-ups it needed are in the same merge: `entity` records carry `id=` so the `relationship` records that name their endpoints by entity id resolve to something the reader has seen, `claim` records carry `statement=` (the id is the backend's `c-8f31a2`, which said a claim changed but not what it says), and the `diff` header carries `saved_at`/`generated_at` so `freshness_previous=current` is not the only thing said about a baseline's age.

PR #457, opened 2026-08-17
feat(context): add `ix context --list` for discovery

Alot1z:
## Problem

Neither humans nor agents had any way to enumerate what `ix context --save <id>` had stored — the only path was reading every JSON file under `IX_HOME/investigations` by hand. An `--list` surface is the obvious counterpart to `--resume`/`--diff`/`--save`, is useful for autonomous workflows that need to recover saved investigations across a session boundary, and matches the discovery story the rest of `ix` already exposes (`ix map` lists files; `ix search` lists candidates).

## Implementation

One commit. Two files:

- [`ix-cli/src/cli/commands/context.ts`](https://github.com/Alot1z/Ix-remap/blob/feat/context-list-investigations/ix-cli/src/cli/commands/context.ts) — adds `--list`, `listInvestigations()`, `renderInvestigationList()`.
- [`ix-cli/src/cli/__tests__/context-investigation.test.ts`](https://github.com/Alot1z/Ix-remap/blob/feat/context-list-investigations/ix-cli/src/cli/__tests__/context-investigation.test.ts) — adds new tests.

Behavior:

- Reads `IX_HOME/investigations/*.json`.
- Mirrors the read-side validation `loadInvestigation` performs on a single file: envelope check (`schema === "ix-investigation/1"`) plus `contextBundleSchema.safeParse(parsed.bundle)`. A corrupt or version-skewed file is **skipped** with a single warning rather than poisoning the listing.
- Sort: newest-first by `savedAt`, falling back to the sanitized `id` when two files share the same millisecond.
- Wire formats match the rest of the CLI:
  - `text`: tabular view (`id`, target, `savedAt`, freshness, counts, truncated).
  - `llm`: total header followed by one metadata record per saved investigation.
  - `json`: full saved envelopes so the output round-trips through `--resume <id>` without backend round-trips.

The CLI action handler also rejects `--list` if `--save` or `--diff` was passed (concerns would conflict).

## Tests

Three new tests in `ix-cli/src/cli/__tests__/context-investigation.test.ts`:

- **happy path** — two saves come back with both ids in deterministic order.
- **corrupt / version-skewed handling** — a directory with one valid file, one bad-envelope file, one truncated-JSON file, and one `bundle.entities: "not-an-array"` file returns only the valid entry in the listing. This is the same hardening surface [#455](https://github.com/ix-infrastructure/Ix/pull/455) closed for `--resume`/`--diff`.
- **empty directory** — a workspace pre-created with no investigations returns `[]`.

- New flag `--list`; nothing removed.
- Three new exports. No existing public API changed.
- The validation contract parallels `loadInvestigation`, so warnings observed when an investigation directory contains a bad file use the same shape users already see from `--resume` on a bad id.

## Provenance

This is a fresh contribution against current `main`. It is **not** a fix for [#455](https://github.com/ix-infrastructure/Ix/pull/455) and does not depend on [#455](https://github.com/ix-infrastructure/Ix/pull/455). The fork branch [`feat/context-list-investigations`](https://github.com/Alot1z/Ix-remap/tree/feat/context-list-investigations) carries the same single commit: [ea26d93](https://github.com/Alot1z/Ix-remap/commit/ea26d93).

## Validation

- `npx tsc --noEmit` → clean.
- `npx eslint --max-warnings 0` on the changed files → clean.
- Live reproduction: `ix context --list` returns the sorted list of saved investigations in `text`/`json`/`llm` formats; corrupt files are skipped with a single warning.

### DIFF
--- a/ix-cli/src/cli/__tests__/context-investigation.test.ts
+++ b/ix-cli/src/cli/__tests__/context-investigation.test.ts
@@ -15,6 +15,7 @@ import type { EntityFacts } from "../explain/facts.js";
 import {
   buildBundle,
   diffInvestigations,
+  listInvestigations,
   loadInvestigation,
   mergeDiffOptions,
   saveInvestigation,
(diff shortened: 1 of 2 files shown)


KageBinary (changes requested):
`--list` is a real gap and the validation mirroring `loadInvestigation` is the right instinct. Two things to fix, plus a rebase.

**1. `listInvestigations()` writes to stdout, which corrupts the machine formats.**

`renderWarning` is `console.log` with chalk (`src/cli/ui.ts:69`), not stderr. So the warning about skipped files is emitted from inside the enumerator, *before* the renderer runs, and lands in the middle of the payload:

```
$ ix context --list --format json | jq .
  Warning  1 saved investigation file(s) did not match the contract; skipped...
parse error: Invalid numeric literal at line 1
```

`--format llm` has the same problem: a decorated prose line in a record stream.

The enumerator shouldn't print at all. Return the skipped count alongside the results and let `renderInvestigationList` decide — a field in the JSON, an `error code=... message="..."` record for llm (the spec defines that shape), and the current warning for text. That also fixes the ordering: right now the warning appears before the `Saved investigations (N)` header rather than with the listing it describes.

**2. Same quoting problem as #456.** The llm branch builds `key=value` with template literals, so `target=${s.bundle.target.name}` goes out unquoted. `src/cli/llm.ts` exports `llmLine`/`llmQuote` for exactly this; `docs/llm-format.md` requires quoting any value containing a space.

**3. The branch is CONFLICTING with `main`** — #455 landed in the same file. Needs a rebase before it can go anywhere.

One design note, not a blocker: `--list` combined with `--save`/`--diff` warns and returns with exit code 0. #462 is turning exactly that class of combination into a hard error with a non-zero exit; whichever lands second should follow the other.

KageBinary:
Superseded by #463, which carries your commit `ea26d93` plus a fix on a branch in this repo — I could not push to your fork from here, and this branch was conflicting with `main` besides.

The substance: `listInvestigations` reported skipped files through `renderWarning`, which is `console.log`, so one corrupt file put a prose line inside the JSON payload and `ix context --list --format json | jq` failed on it. The enumerator returns the count now and the renderer decides how to say it. The records also go through `llmLine`, and validation moved to `savedInvestigationSchema` since #455 landed while this was open.

Leaving this open rather than closing it — your call on which branch you would rather carry forward.

KageBinary:
Closing in favour of #463, which contains this branch's commit `ea26d93` unchanged — verified as an ancestor, so nothing here is lost. The work is yours; #463 adds a fix commit on top and merges `main` (which had moved under this branch since #455 landed).

Thanks for the feature — the gap was real and `--list` is the right shape for it.

Alot1z:
Thanks for the carry — closing this in favour of #463 is right, and the fix on top (the enumerator returning the count and leaving the prose decision to the renderer) is the correct shape: a corrupt file should not be able to put a prose line inside a JSON payload. Nothing here needs preserving beyond `ea26d93`.

PR #458, opened 2026-08-17
feat(context): surface effective/requested/saved budgets on `ix context --diff`

Alot1z:
## Problem

`ix context --diff <id>` accepts `--max-*` CLI flags, but those values currently do not affect the fresh-side budgets; the saved investigation's budgets remain effective. The action handler in [`ix-cli/src/cli/commands/context.ts`](https://github.com/ix-infrastructure/Ix/blob/main/ix-cli/src/cli/commands/context.ts) calls:

```ts
const fresh = await buildFreshBundle(
  target ?? saved.bundle.target.name,
  { ...opts, ...mergeDiffOptions(saved, opts) },
  saved.bundle.budgets,
);
```

`buildFreshBundle` accepts only `{kind, path, pick, depth, asOfRev}` from `opts`. Even though `opts` is spread into the second argument, `opts.max*` is filtered out by the parameter type, and the third argument (the budgets that actually bound the fresh bundle) is unconditionally `saved.bundle.budgets`.

The resulting transparency gap is that the diff output does not expose this precedence, so a user or agent cannot determine the effective comparison budget without inspecting the implementation. The change surfaces the silent precedence without altering it.

## Evidence

Reproduction against the running IX backend (workspace `b4-fixture`, rev 122, `ghcr.io/ix-infrastructure/ix-memory-layer:latest`):

```text
# Save with a small evidence budget
ix context centroid --kind function --save b4-small \
  --max-entities 5 --max-evidence 2 --max-relationships 1

# Try to reflow the diff using --max-*
ix context --diff b4-small --kind function \
  --max-entities 50 --max-evidence 25 --max-relationships 100 \
  --format json
# → "added.evidence=[] removed.evidence=[]"
# (fresh side kept max-evidence=2; CLI flag ignored)
```

Converse (saved large, CLI wants small):

```text
ix context --diff b4-large --kind function \
  --max-entities 1 --max-evidence 1 --max-relationships 0 --format json
# → "added.evidence=[] removed.evidence=[]"
# (fresh side kept max-evidence=25; CLI flag ignored)
```

Full reproducer and adversarial review live in [`Alot1z/Ix-findings/findings/ix-context/reproducers/b4-budget-precedence.md`](https://github.com/Alot1z/Ix-findings/blob/main/findings/ix-context/reproducers/b4-budget-precedence.md); see finding **F-024** in [`Alot1z/Ix-findings/planning/findings/registry.json`](https://github.com/Alot1z/Ix-findings/blob/main/planning/findings/registry.json).

## Implementation

One commit. Two files: [`ix-cli/src/cli/commands/context.ts`](https://github.com/Alot1z/Ix-remap/blob/feat/context-diff-budget-transparency/ix-cli/src/cli/commands/context.ts) and [`ix-cli/src/cli/__tests__/context-investigation.test.ts`](https://github.com/Alot1z/Ix-remap/blob/feat/context-diff-budget-transparency/ix-cli/src/cli/__tests__/context-investigation.test.ts).

`InvestigationDiff` gains a `budgets: { saved, requested?, effective, note }` field:

- `saved` is `saved.bundle.budgets` (always present).
- `requested` is the subset of `--max-*` flags the user actually passed on the command line (`undefined` when none supplied). Missing keys render as `not-given` in human form so partial subsets do not silently zero out.
- `effective` is what `buildFreshBundle` was actually invoked with — today this is always equal to `saved`, centralised in one line so any future change that lets CLI overrides win has a single observable truth.
- `note` is a one-sentence explanation of the precedence rule.

`parseRequestedBudgets` is exported for direct testing. It parses raw commander string values, returns `undefined` when no flag was supplied, and treats whitespace / non-numeric tokens as absent.

This requires dropping the `"50"` / `"100"` / `"25"` / `"12000"` commander defaults on `--max-*`; the same numeric defaults are reapplied at build time via `clampInt(opts.max*, 1, 500, 50)` etc., so direct `ix context` runs without `--diff` are byte-for-byte unchanged.

Text and LLM render paths add a four-line `budgets:` block. JSON keeps the full structure.

## Tests

Three new tests in `ix-cli/src/cli/__tests__/context-investigation.test.ts`:

- `exposes effective budgets on every --diff, sourced from the saved investigation` — without a CLI override, `requested` is absent and `effective` mirrors `saved`; with CLI overrides parsed from raw strings, `requested` captures them and the note makes the precedence explicit; the `effective === saved` invariant is pinned.
- `treats whitespace, empty, or non-numeric --max-* strings as not provided` — empty objects, blanks, and `undefined` all return `undefined`; numeric integer `0` is honoured as a real override.
- `renders the budget block in --diff text and llm output` — captures `console.log` via `vi.spyOn`; asserts the four-line `budgets:` block appears in text, in `llm`, and in the JSON single-line output. A control case prevents a renderer that always emits one shape from passing silently.

- JSON output adds a new top-level `budgets` field on the existing `ix-investigation-diff/1` schema. Additive. Consumers generally ignore unknown fields. No zod schema in the codebase validates this output.
- Text and LLM outputs gain a four-line `budgets:` block.
- **Runtime precedence is unchanged**: `effective` always equals `saved`. This is pinned by tests.
- `--diff` runtime semantics for `--max-*` flags on the CLI are unchanged. No breaking change. No new flag. No existing flag has a new effect.
- Direct `ix context <target> --max-* N` (without `--diff`) is byte-for-byte unchanged because the dropped commander defaults are still applied via `clampInt` at build time.

## Provenance

This is a fresh contribution against current `main`. It is **not** a fix for [#455](https://github.com/ix-infrastructure/Ix/pull/455) and does not depend on [#455](https://github.com/ix-infrastructure/Ix/pull/455) to be merged. The fork branch [`feat/context-diff-budget-transparency`](https://github.com/Alot1z/Ix-remap/tree/feat/context-diff-budget-transparency) carries the same single commit: [58aea40](https://github.com/Alot1z/Ix-remap/commit/58aea40).

It is independent of B-1 ([`feat/diff-format-llm-parity`](https://github.com/Alot1z/Ix-remap/tree/feat/diff-format-llm-parity), [6363612](https://github.com/Alot1z/Ix-remap/commit/6363612)) and B-2 ([`feat/context-list-investigations`](https://github.com/Alot1z/Ix-remap/tree/feat/context-list-investigations), [ea26d93](https://github.com/Alot1z/Ix-remap/commit/ea26d93)), which are proposed as separate PRs.

The audit and forensic reproduction underlying the silent-ignore finding are documented as **F-024** in [`Alot1z/Ix-findings/planning/findings/registry.json`](https://github.com/Alot1z/Ix-findings/blob/main/planning/findings/registry.json).

## Validation

- `npx tsc --noEmit` → clean.
- `npx eslint --max-warnings 0` on the changed files → clean.
- Live reproduction: `ix context --diff b4-large --kind function --max-evidence 1 --format llm` now prints:

```text
budgets
  saved     : entities=50 relationships=100 evidence=25 chars=12000
  requested : entities=1 relationships=0 evidence=1 chars=not-given (CLI override; not applied to --diff fresh side)
  effective : entities=50 relationships=100 evidence=25 chars=12000
  note      : Saved investigation budgets govern --diff; --max-* flags on the CLI are recorded above for transparency but not applied to the fresh side.
```

### DIFF
--- a/ix-cli/src/cli/__tests__/context-investigation.test.ts
+++ b/ix-cli/src/cli/__tests__/context-investigation.test.ts
@@ -1,7 +1,7 @@
 import { existsSync, mkdtempSync, mkdirSync, readdirSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
-import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 import type {
   ConflictReport,
@@ -17,6 +17,8 @@ import {
   diffInvestigations,
   loadInvestigation,
   mergeDiffOptions,
+  parseRequestedBudgets,
+  renderInvestigationDiff,
   saveInvestigation,
 } from "../commands/context.js";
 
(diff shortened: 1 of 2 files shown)


KageBinary (changes requested):
Surfacing the saved-budgets precedence is worth doing — it is genuinely invisible today, and "requested but not applied" is the kind of thing an agent cannot infer. Two problems with the rendering, one with a comment.

**1. The `llm` branch emits prose.**

```js
if (format === "llm") {
  console.log(`  budgets`);
  console.log(`    saved     : ${formatBudgets(diff.budgets.saved)}`);
```

That is the text renderer with the colon moved: two-space indents, aligned padding, a bare header line. `docs/llm-format.md` is explicit — one record per line, `key=value` pairs, *"no decorative whitespace, separators, or headers"*. It should be one record built with `llmLine` from `src/cli/llm.ts`, e.g.

```
budgets scope=saved entities=5 relationships=1 evidence=2 chars=12000
budgets scope=requested evidence=25
budgets scope=effective entities=5 relationships=1 evidence=2 chars=12000
```

with the omitted fields simply absent (`llmField` drops nullish values, so "not-given" needs no sentinel) and the note dropped or moved to a single `note="..."` field.

**2. `parseRequestedBudgets` does not do what its own doc says.** The comment reads *"Apply the same clamping rules the action handler uses for direct --save runs"* — but there is no clamping in the function. The handler uses `clampInt(opts.maxEntities, 1, 500, 50)` and friends; this accepts `0`, negatives, and anything up to `Number.MAX_SAFE_INTEGER`. Since the values are reported and not applied it is not dangerous today, but the comment describes behaviour the code does not have, and the moment someone makes `requested` win on the fresh side it becomes a real bug. Either clamp or correct the comment.

**3.** `JSON.parse(json.lines[0].replace(/^undefined$/, ""))` in the test is working around something in the capture rather than fixing it — worth understanding what emits `undefined` there.

Worth noting for whoever lands this and #456 together: both change `renderInvestigationDiff`'s llm path, and they will conflict.

KageBinary:
Superseded by #464, which carries your commit `58aea40` plus a fix — I could not push to your fork from here.

The substance: the `--format llm` branch was the prose block with the colons moved, and it could not have been otherwise, because on this branch the whole diff renderer is prose and there was no record stream to emit a budget record into. So #464 is stacked on #456 and emits three `budgets scope=…` records, with `applied=false` carrying the precedence rule as a field rather than a sentence.

One note worth keeping: `parseRequestedBudgets` not clamping is *correct* — these values are reported and never applied, so clamping would misreport what the caller asked for. It was the comment that was wrong, and it now says that.

Leaving this open rather than closing it.

KageBinary:
Closing in favour of #464, which contains this branch's commit `58aea40` unchanged — verified as an ancestor. #464 adds a fix commit and stacks on #456, which is what the llm half of this needed to exist at all.

Thanks — the saved-budgets precedence being invisible was a genuine find, and it is data in the output now rather than something you have to read the source for.

Alot1z:
Acknowledged and agreed — #464 is the better home for `58aea40` now that the llm record stream exists to carry the budget record, and stacking on #456 is what the `--format llm` half needed. Noted on the `parseRequestedBudgets` point: the code was right and the comment was wrong — values that are reported and never applied should not be clamped, and clamping would have misreported what the caller asked. Thanks for fixing the comment rather than the code.

PR #460, opened 2026-08-18
fix(view): stop serving a cached Compass over a new install

KageBinary:
## The report

> it reads that release but it still looks the same

rc.10 installed, status bar showing `0.10.0-rc.10+release.6bce261`, screen
showing the build before it.

## What was actually stale

Not the install. I pulled the published **Windows** zip and checked the bundle
inside it: stamp `0.10.0-rc.10+release.6bce261`, `index.html` naming
`assets/index-BbVrNAev.js`, and that file containing every marker from the
change (`data-file-dot`, `What joins these files`, `Type references`). The right
code was on disk.

The browser was serving the old one. The generated `ix view` server sends
`Content-Type` and nothing else — **no `Cache-Control`, no `ETag`, no
`Last-Modified`**. That is not "do not cache"; with no freshness information at
all a browser applies its own heuristic and may reuse a response without
revalidating. `ix view` serves `127.0.0.1` on a port it reuses, so the cache key
survives an upgrade: the old `index.html` comes back, names the old
content-hashed bundles, those are cached too, and the entire previous Compass
runs.

## The build stamp cannot catch this — and reads as if it disproves it

`useBuildVersion` fetches `/.version`: a **separate request for a separate
file**. So it reports the version on disk while the JS executing came from
cache. Its own comment says it shows

> the bundle actually being served — not what some build-time constant claimed

which against a stale cache is precisely backwards. The stamp exists to tell
"same issue" apart from "the fix has not reached me", and in this failure mode
it says the opposite of the truth. That is what turned a caching problem into a
bug report against the fix.

## The change

| path | header | why |
|---|---|---|
| `/`, `index.html`, `/.version` | `no-store` | no fingerprint in the name, so a stale copy is indistinguishable from a fresh one |
| `/assets/*` | `public, max-age=31536000, immutable` | Vite names these by content — the name changes when the file does |
| SPA fallback | `no-store` | a stale `index.html` requests a bundle that no longer exists; caching the fallback under the missing asset's URL is how one bad load becomes permanent |

The point is not to disable caching. The fingerprinted assets are the part worth
caching and they get a year; only the two unfingerprinted entry points are
forced to revalidate.

## Tests

Four new, in the existing spawn-a-real-server harness. Three go red without the
change with `expected null to be 'no-store'` — the header was simply absent.
Full `ix-cli` suite green: **990 passed**, 69 files.

## One note for review

No backticks in the new comments. The whole server is emitted from a template
literal in `view.ts`, so a backtick ends the literal and a backslash escape is
eaten before it reaches the generated file. The first draft of this comment
contained `` `ix view` `` and took the server down — caught by the tests, worth
knowing before editing that block.

## Not fixed here

The stamp's own claim. Making it genuinely report the running bundle needs a
build-time constant baked into the JS and compared against the fetched file, so
a mismatch could say "you are running a cached build" outright. Worth doing, but
it is a Compass change and a bigger one; with caching fixed the stamp is honest
again in practice.

### DIFF
--- a/ix-cli/src/cli/__tests__/view-running-port.test.ts
+++ b/ix-cli/src/cli/__tests__/view-running-port.test.ts
@@ -267,15 +267,18 @@ describe("view running port state", () => {
 describe("runningInstanceLines", () => {
   it("never builds a URL from the requested port", async () => {
     const { runningInstanceLines } = await import("../commands/view.js");
-    const lines = runningInstanceLines(19123, 19124, true);
+    // The token is supplied so the assertion stays exact; it is the cache bust
+    // that makes an already-running visualizer reachable after `ix upgrade`
+    // replaced the dist underneath it.
+    const lines = runningInstanceLines(19123, 19124, true, "tok");
 
-    expect(lines[0]).toBe("  http://localhost:19123");
+    expect(lines[0]).toBe("  http://localhost:19123/?ix=tok");
     expect(lines.some(l => l.includes("http://localhost:19124"))).toBe(false);
   });
 
   it("does not warn when the running port is the one that was asked for", async () => {
     const { runningInstanceLines } = await import("../commands/view.js");
-    expect(runningInstanceLines(8080, 8080, true)).toEqual(["  http://localhost:8080"]);
+    expect(runningInstanceLines(8080, 8080, true, "tok")).toEqual(["  http://localhost:8080/?ix=tok"]);
   });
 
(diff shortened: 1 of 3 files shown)


josephismikhail (approved):
Reviewed by two independent review agents over two rounds; every finding from both is addressed in the two follow-up commits, and each fix is mutation-validated (the bug restored, the new test confirmed red, then reverted). Full ix-cli suite green apart from the three read-containment symlink tests that always EPERM on Windows locally.

PR #461, opened 2026-08-18
fix(ci): the Windows install test never ran on a release

KageBinary:
## What went wrong

#459 shipped the Windows installer smoke test with a `release: published`
trigger and this claim in the workflow header:

> A published release is the moment the installer's inputs change, and the only
> moment a real user is about to run it against them.

**It never fired.** rc.11 went out today and no run started.

The release is created by `github-actions[bot]` using `secrets.GITHUB_TOKEN`,
and GitHub does not start new workflow runs from events created with that token.
So the trigger could not have fired for any release this project makes — the two
paths I exercised by hand (the PR, and a manual dispatch against rc.10) are the
only two that ever worked, which is exactly why it looked verified.

My error, and worth being precise about the shape of it: the test itself is
sound and proved its assertions bite. What was untested was the *trigger*, and
the two triggers I did exercise were the ones that could not reveal the problem.

## The fix

`release.yml` calls the workflow directly once the assets are up, as a
`workflow_call`. Same run, no event required, and a failure attaches to the run
that produced the artifacts rather than to nothing.

It is deliberately **not a gate** — by the time it runs the release exists — but
that is the difference between finding out now and finding out from a user.
Making it block would mean moving it before publish, which means installing a
release that does not yet exist.

The `release:` trigger and the dead `github.event.release.tag_name` expression
are removed; `workflow_dispatch` stays for checking a specific version by hand.

## Cleanup that falls out of it

The zizmor `cache-poisoning` exception added in #459 goes too. That finding was
raised *only* because of the `release:` trigger; with the trigger gone there is
nothing to accept. An exception nobody needs is one nobody rechecks.

## Verification

This one is harder to prove before merge than #459 was — the PR cannot trigger a
release. What I can say:

- both workflows parse, and the call passes `needs.release.outputs.version`,
  which is newly exposed from the release job's existing `version` step
- the called workflow is unchanged apart from its trigger, and its assertions
  were proved to fail on a planted regression in #459
- the first real proof will be the next tag; if it does not appear there, that
  is the same class of bug again and worth treating as such


josephismikhail (approved):
Reviewed by two independent review agents over two rounds. The first round's fix put actionlint in actions-lint.yml, where the `main` ruleset — which requires exactly one check, `CI Passed` — meant it could report a broken reusable-workflow call and gate nothing; it is now a step of `Lint & Typecheck`, which `CI Passed` needs, and confirmed running and green on this PR. The version assertion closes the other half: an empty `version` input degrades to "latest stable", so a wiring break would have installed an unrelated build and passed every check against it.

PR #462, opened 2026-08-19
feat(cli): reject contradictory mode/output flag combinations

Alot1z:
## Problem

Both `ix context` and `ix diff` accept combinations of mutually exclusive flags whose execution paths silently drop the typed flag instead of surfacing the conflict. Documented combinations only sometimes produce a warning; undocumented combinations are pure no-ops.

### `ix context`

The action handler `return`s early on `--resume` (line 141) and `--diff` (line 146) before the `--save` branch (line 204) and `--out` block (line 211) ever run. The result is six distinct silently-dropped combinations:

| Form                                                           | Dropped flag | User intent                              |
| -------------------------------------------------------------- | ------------ | ---------------------------------------- |
| `ix context --resume <id> --diff <id>`                         | depends on `${RESUME_OR_DIFF}` winning  | upside-down compare vs saved             |
| `ix context --resume <id> --save <id>`                         | `--save`     | persist the saved investigation as new  |
| `ix context --resume <id> --out <file>`                        | `--out`      | persist the rendered output              |
| `ix context --diff <id> --save <id>`                           | `--save`     | persist the fresh side as a new investigation |
| `ix context --diff <id> --out <file>`                          | `--out`      | persist the comparison result to a file  |
| `ix context --save <id> --out <file>`                          | `--out`      | two distinct write destinations          |

### `ix diff`

`--summary` `return`s at line 472 before the `--content` branch at line 510 is reached. `--full --limit <n>` runs `opts.full ? undefined : opts.limit` (line 514), silently dropping `--limit`. A user asking for the full textual diff with `--summary --content` got a summary record. A user asking for "no limit, but at least 20" got an unbounded response.

## Implementation

Two exported pure functions and one line of glue per action handler:

- `detectContextModeConflict(opts)` — returns a message naming the conflict, or `undefined` if the combination is legal.
- `detectDiffModeConflict(opts)` — same shape, scoped to `ix diff` flag pairs.
- Each action handler calls the detector before any network call and `console.error("Error:", message); process.exitCode = 1; return;` when the detector reports a conflict. This mirrors the explicit-conflict style in `ix subsystems` (where `--detailed` without `--list`, `--explain` with `--list`, etc. behave the same way) and is strictly stronger than the existing `--out`/`--format` cross-mode `renderWarning(...)` fallthrough, where one of the flags is silently coerced and the other wins.

The detector functions are exported with structural option types so they can be unit-tested without driving Commander.

## Tests

30 new tests in `ix-cli/src/cli/__tests__/context-mode-conflict.test.ts`:

- **detector-only**: every legal combination (no flags, single mode flag, save+format text, out+format json) returns `undefined`; every documented conflict returns the expected reason string.
- **integration through the real `registerX` functions**: 6 `ix context` combos and 4 `ix diff` combos drive a `new Command().exitOverride()` program; stderr contains `Error:` followed by the conflict-specific reason; `process.exitCode === 1`; stdout is empty. No mock backend is required because the detector short-circuits before any network call.
- **smoke**: help text for both commands still lists every flag the user can type.

- **Hard errors** on six new mutually-exclusive `ix context` combinations and four new `ix diff` combinations. Each was previously a silent-ignore.
- **No flag was renamed, no default was removed, no new flag was added.**
- **No existing happy-path call changes.** The detector is the first statement in the action handler; legal combinations take exactly the same code path as before.
- **Runtime semantics on `--list` (B-2, currently in flight as #457), `--save`, `--resume`, `--diff`, `--out` are otherwise unchanged.** None of those branches adjust; the detector only inspects combos among them.

## Provenance

A bounded pass over the `ix-cli` source on the audit baseline identified `ix context`/`ix diff` mode combinations as the largest remaining cluster of silent-ignore flag gaps beyond the `--diff --max-*` transparency work tracked in [F-024 / #458](https://github.com/ix-infrastructure/Ix/pull/458). This contribution is independent of the existing open PRs:

- [#455](https://github.com/ix-infrastructure/Ix/pull/455) — F-023 read-side `loadInvestigation` validation. Different code paths, no overlap with this conflict-detection layer.
- [#456](https://github.com/ix-infrastructure/Ix/pull/456) ([`6363612`](https://github.com/Alot1z/Ix-remap/commit/6363612)) — B-1 `--format llm` records on `ix context --diff`. This PR can land independently without affecting the new error path, since the detector runs first.
- [#457](https://github.com/ix-infrastructure/Ix/pull/457) ([`ea26d93`](https://github.com/Alot1z/Ix-remap/commit/ea26d93)) — B-2 `ix context --list`. The `--list` validation this PR carries (`--list + --save/--diff` → warning, return) is unchanged; the new conflict detector would add `--list + --resume/--out` to the warning-equivalent surface area, but those cases are already unreachable because `--resume` short-circuits the action handler — adding detector coverage there is a follow-up if maintainers prefer stronger consistency.
- [#458](https://github.com/ix-infrastructure/Ix/pull/458) ([`58aea40`](https://github.com/Alot1z/Ix-remap/commit/58aea40)) — F-024 budget transparency on `--diff`. Different code paths; no overlap.

This branch is `Alot1z/Ix-remap:feat/context-mode-conflict-validation` at [`5431147`](https://github.com/Alot1z/Ix-remap/commit/5431147).

The branch is rebased on the audit baseline `8be5f110` ahead of the recent `chore(deps)` and Windows-CI changes (#452, #450, #451, #453, #454, #459). The diff in this PR omits those non-conflicting bumps; merging the bumps separately is unaffected. PR base is `main` (snapshot `8be5f110`), the same base the prior four PRs (#455–#458) were opened against for the same reason.

## Validation

- `vitest run`: 1005 passed | 12 pre-existing skips | 70 test files.
- `npx tsc --noEmit`: clean.
- `npm run build`: clean (`@ix/cli@0.9.3 build` produces `dist/`).
- `npx eslint --max-warnings 0 src/cli/commands/context.ts src/cli/__tests__/context-mode-conflict.test.ts`: clean.
- `npx eslint --max-warnings 0 src/cli/commands/diff.ts`: 2 pre-existing on `origin/main` (dead-code on `sliceHunksToSpan` / `getLineSpan` at file lines 266, 397 — neither exported nor called anywhere in the project). These warnings exist at `8be5f110` independently of this commit and are out of scope here; flagging them so reviewers don't assume this commit introduced them.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -267,7 +267,9 @@ ix inventory --kind function --format json
 - Use `ix inventory` instead of `ix search ""` for listing entities by kind
 - Use `ix diff --summary` for broad revision comparisons (server-side, fast)
 - Use `--full` only when you need every individual change
-- Always use `--limit` to cap result sets
+- Always use `--limit` to cap result sets — except on `ix diff`, where
+  `--summary` and `--full` each already fix the volume and passing `--limit`
+  alongside either is refused rather than silently ignored
 - Use `--format llm` when you are reading the output; `--format json` only when chaining results between commands or extracting a specific field
 - Use `--path` or `--language` to restrict text searches
 - Use exact entity IDs from previous JSON results
--- a/ix-cli/src/cli/__tests__/context-mode-conflict.test.ts
+++ b/ix-cli/src/cli/__tests__/context-mode-conflict.test.ts
@@ -1,0 +1,246 @@
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { Command } from "commander";
+
+import {
+  detectContextModeConflict,
+  registerContextCommand,
+} from "../commands/context.js";
+import {
+  detectDiffModeConflict,
+  registerDiffCommand,
+} from "../commands/diff.js";
+
+/**
+ * C-1..C-4 silent-ignore flag gaps in `ix context` and C-5 in `ix diff` are now
+ * surfaced as hard errors at the top of the action handler. These tests pin
+ * both the pure functions and the integration through the real Commander
+ * registration. The detectors run before any network call, so the action
+ * returns before touching the real Ix backend; no mocks are needed for the
+ * conflict paths.
+ */
+
+let home: string;
+let origExitCode: number | string | undefined;
+
+beforeEach(() => {
+  home = mkdtempSync(join(tmpdir(), "ix-mode-conflict-test-"));
+  process.env.IX_HOME = home;
+  origExitCode = process.exitCode;
+  process.exitCode = undefined;
+});
+
+afterEach(() => {
+  delete process.env.IX_HOME;
+  rmSync(home, { recursive: true, force: true });
+  process.exitCode = origExitCode;
+});
+
+describe("detectContextModeConflict", () => {
+  it("returns undefined when no flags are set", () => {
+    expect(detectContextModeConflict({})).toBeUndefined();
+  });
+
+  it("returns undefined for a single mode flag with neutral extras", () => {
+    expect(detectContextModeConflict({ resume: "x" })).toBeUndefined();
+    expect(detectContextModeConflict({ diff: "x" })).toBeUndefined();
+    expect(detectContextModeConflict({ save: "y" })).toBeUndefined();
+    expect(detectContextModeConflict({ out: "/tmp/x.json" })).toBeUndefined();
+  });
+
+  it("returns undefined for the legal save+resolve target run", () => {
+    // Fresh build with --save is fine; this is the contract path.
+    expect(detectContextModeConflict({ save: "y", format: "text" })).toBeUndefined();
+    // Fresh build with --out is fine in json mode (the existing renderWarning
+    // path still applies; the conflict detector does not flag it).
+    expect(detectContextModeConflict({ out: "/tmp/x.json", format: "json" })).toBeUndefined();
+  });
+
+  it("flags --resume + --diff", () => {
+    const msg = detectContextModeConflict({ resume: "x", diff: "y" });
+    expect(msg).toMatch(/--resume and --diff cannot be combined/);
+  });
+
+  it("flags --resume + --save", () => {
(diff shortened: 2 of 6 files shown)


KageBinary (dismissed):
The premise checks out — I read the handlers rather than taking the description on trust. `opts.save` returns before the `--out` block, and `--resume`/`--diff` return before both, so every combination in your table really is a silently dropped flag. Turning those into errors is right, and the detectors being pure functions with a Commander integration test on top is the right shape.

**One required change: the error bypasses the `--format llm` contract.**

`reportContextModeConflict` writes chalk-coloured prose to stderr. Every other command in the CLI routes errors through `llmError` when the caller asked for records:

```js
if (opts.format === "llm") console.log(llmError("unresolved_target", `...`));
```

(`imports.ts:9`, `trace.ts:475`, `smells.ts:71`, `locate.ts:187`, `callers.ts:14` — and `docs/llm-format.md` specifies the shape: `error code=<slug> message="..."`, in-stream, process still exits non-zero.) An agent that passes `--format llm` unconditionally — which the docs tell it to do — gets a non-zero exit and nothing parseable explaining why. `detectContextModeConflict` already returns a message string, so this is a branch in the reporter plus a slug per conflict.

**One question for a maintainer rather than for you:** `--save` + `--out` is declared mutually exclusive, which turns today's early-return into a documented rule. Unlike the others it is not obviously a conflict — persisting an investigation and writing the rendered bundle to a path are different operations, and a user could reasonably want both. The other five are genuinely contradictory; this one is a limitation being frozen into an error message. @ianjakobhock / @josephismikhail — worth a call before it ships.

Also worth a line in the PR body: this is a behaviour change for anyone scripting these combinations today. Pre-1.0 and preferable to a silent no-op, but it belongs in the release notes rather than being discovered.

The tests are good — particularly asserting through the real Commander registration rather than only the detector.

josephismikhail (approved):
Reviewed by two independent review agents over two rounds. Findings addressed on this branch and in the stack: rule ordering so the flag actually in charge is the one named, one shared conflict reporter instead of two byte-identical copies, `ContextModeOptions` derived from `ContextOptions` rather than hand-copied, an actionable `--resume --out` message, and the CLAUDE.md / skill guidance that told agents to always pass `--limit` — which this PR turns into a hard error on `ix diff --summary`.

PR #463, opened 2026-08-19
feat(context): add `ix context --list` for discovery

KageBinary:
Carries **@Alot1z's** commit from #457 (`ea26d93`) plus a fix commit, on a branch in this repo. Pushing to their fork is not something I can do from here — #457 could not be updated in place, and it was conflicting with `main` besides. **Supersedes #457**; the original is still open for @Alot1z to look at.

## The feature

`ix context --list` enumerates saved investigations. Nothing else showed you what you had stored — the only path was reading the JSON files directly.

## What the fix commit changes

**The skip report was in the payload.** `listInvestigations` called `renderWarning`, and every renderer in `ui.ts` writes to **stdout**. One corrupt file therefore prepended a chalk-coloured prose line to the JSON:

```
$ ix context --list --format json | jq .
  Warning  1 saved investigation file(s) did not match the contract; skipped...
parse error: Invalid numeric literal at line 1
```

`--format llm` got the same line in the middle of a record stream. The enumerator counts skips and returns the count now; the renderer reports it in each format's own terms — stderr for the human warning, a `skipped=` field on the `investigations` record for llm, and nothing at all when nothing was skipped. A test pins that the enumerator prints nothing whatsoever, and another pins that JSON still parses with three files skipped.

**Records go through `llmLine`.** They were built with template literals, so a target name containing a space would split the record. Same fix as #456.

**Validation is `savedInvestigationSchema` now.** #455 landed in `main` while this sat, so the hand-rolled envelope check (`schema !== "ix-investigation/1"` then a separate bundle parse) is gone — one contract, the same one the write side and `loadInvestigation` enforce, with the same `parsed.data` assertion.

**`--list` with `--save`/`--diff`/`--resume` exits non-zero.** It warned and exited 0, which tells a script that the listing it never got was fine. `refuseInvestigation`, two functions down in the same file, already sets the status this way.

## Verification

1000 tests green, lint, typecheck and knip clean, against `main` as merged.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -267,7 +267,9 @@ ix inventory --kind function --format json
 - Use `ix inventory` instead of `ix search ""` for listing entities by kind
 - Use `ix diff --summary` for broad revision comparisons (server-side, fast)
 - Use `--full` only when you need every individual change
-- Always use `--limit` to cap result sets
+- Always use `--limit` to cap result sets — except on `ix diff`, where
+  `--summary` and `--full` each already fix the volume and passing `--limit`
+  alongside either is refused rather than silently ignored
 - Use `--format llm` when you are reading the output; `--format json` only when chaining results between commands or extracting a specific field
 - Use `--path` or `--language` to restrict text searches
 - Use exact entity IDs from previous JSON results
--- a/docs/llm-format.md
+++ b/docs/llm-format.md
@@ -83,6 +83,21 @@ contains method=12 field=4
 item name=parseFile kind=method
 ```
 
+`ix context --list`:
+
+```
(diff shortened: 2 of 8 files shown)


josephismikhail (approved):
Reviewed by two independent review agents over two rounds. All findings addressed: the `--list` guard moved into the detector (its `--resume` arm was unreachable below `if (opts.resume)`), refusals routed off stdout and given a non-zero status, ids decoded once at the read boundary so every format agrees and the listed id loads, `--format json` returning a summary plus `skipped` rather than every full bundle, and the newest-first test rewritten — it passed against an inverted comparator.

PR #464, opened 2026-08-19
feat(context): surface effective/requested/saved budgets on `ix context --diff`

KageBinary:
Carries **@Alot1z's** commit from #458 (`58aea40`) plus a fix, on a branch in this repo — pushing to their fork is not something I can do from here. **Supersedes #458**, which stays open for @Alot1z.

**Stacked on #456**: merge that first. The reason is the defect itself.

## What was wrong

The `format === "llm"` branch printed

```js
console.log(`  budgets`);
console.log(`    saved     : ${formatBudgets(diff.budgets.saved)}`);
```

— two-space indents, aligned colons, a bare header line, in the format `docs/llm-format.md` defines as *"no decorative whitespace, separators, or headers"*. It was the text renderer with the colon moved, and it could not have been anything else: on this branch the whole diff renderer is prose, and `--format llm` never had a record stream to put a budget record into. That is why this is stacked on #456 rather than standalone.

## What it emits now

```
budgets scope=saved entities=5 relationships=1 evidence=2 chars=12000
budgets scope=requested evidence=25 applied=false
budgets scope=effective entities=5 relationships=1 evidence=2 chars=12000
```

`applied=false` rather than the explanatory sentence: the precedence rule is that saved budgets govern `--diff`, and a field an agent can test beats a note it has to read. Absent overrides are simply absent — `llmField` drops nullish, so the record form needs no `not-given` sentinel the prose form uses.

## One comment that was load-bearing and wrong

`parseRequestedBudgets` said it applied *"the same clamping rules the action handler uses"*. It applies none — the handler uses `clampInt(opts.maxEntities, 1, 500, 50)`. Not dangerous while the values are reported and never applied, which is exactly the point: clamping them **would misreport what the caller asked for**. So the code is right and the comment was wrong; the comment now says so, and says where the ranges live if these are ever made to win on the fresh side.

Also drops `JSON.parse(json.lines[0].replace(/^undefined$/, ""))` from the test — the capture works, the guard was hiding nothing.

## Verification

1001 tests green, lint, typecheck and knip clean, with `main` merged in.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -267,7 +267,9 @@ ix inventory --kind function --format json
 - Use `ix inventory` instead of `ix search ""` for listing entities by kind
 - Use `ix diff --summary` for broad revision comparisons (server-side, fast)
 - Use `--full` only when you need every individual change
-- Always use `--limit` to cap result sets
+- Always use `--limit` to cap result sets — except on `ix diff`, where
+  `--summary` and `--full` each already fix the volume and passing `--limit`
+  alongside either is refused rather than silently ignored
 - Use `--format llm` when you are reading the output; `--format json` only when chaining results between commands or extracting a specific field
 - Use `--path` or `--language` to restrict text searches
 - Use exact entity IDs from previous JSON results
--- a/docs/llm-format.md
+++ b/docs/llm-format.md
@@ -135,16 +135,37 @@ is not enough: the id is the backend's, and the statement is what changed.
 
 ```
 investigations total=2 skipped=1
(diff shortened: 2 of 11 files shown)


josephismikhail (approved):
Reviewed by two independent review agents over two rounds; every finding from both is addressed, and each fix is mutation-validated (bug restored, new test confirmed red, reverted). Second round caught real defects in the first round's fixes — `requestedApplied` derived by value comparison reported `applied=true` for a flag that changed nothing, and the id decoder assumed a two-hex escape that `sanitizeId` does not always emit. Full ix-cli suite green apart from the three read-containment symlink tests that always EPERM on Windows locally.

PR #465, opened 2026-08-19
fix(ingest): bound the manifest reads a scanned repo controls

KageBinary:
Found while sweeping for 0.10.0. Not filed as an issue; it is the last unguarded read in the ingest path.

## The problem

`system.ts` reads `package.json`, `Cargo.toml`, `pom.xml`, `go.mod` and the rest **by name, out of whatever directory is being scanned**, with a bare `fs.readFileSync(path, "utf8")` in three places.

`readFileSync` does not fail on `package.json -> /dev/zero`. It does not return either — it consumes memory until the process dies. Measured: the call had to be killed after 15s. A plain 2 GB `Cargo.toml` does the same without needing a symlink, and a FIFO blocks whatever opens it.

This is reachable from `ix map`, `ix ingest` and `ix view` — anywhere a repository the user did not write gets scanned.

## The fix

One `readManifest`, `statSync` before the read, in that order:

- it **follows** the symlink, so the check applies to what would actually be read rather than to the link;
- `isFile()` refuses a device, a FIFO and a directory. **Stat rather than open** — opening a FIFO blocks until a writer appears, so a guard that opened would be the thing that hung;
- `size` refuses the honest-but-enormous file, at the same 1 MB the ingest walker already gives a source file.

That shape is `loadConfig`'s in `ts-module-resolution.ts`, whose comment makes the point this file missed: manifests *"sit outside ingestion's own size gate, so they need their own"*. `system.ts` was the third such reader and never got one.

## Tests

- an ordinary manifest still reads
- an oversized one is refused — **and the test goes red with the cap removed**, so it is the cap doing the work, not the JSON parse failing
- a manifest that is a directory is refused
- POSIX only: a manifest symlinked to `/dev/zero`, and a manifest that is a FIFO, both return promptly rather than never. Skipped on Windows, which has neither.

The device and FIFO guards are the two I could not prove by removal without hanging or OOM-ing the runner; they are covered by `isFile()` and by the direct measurement above.

1000 tests green, lint, typecheck and knip clean.

### DIFF
--- a/ix-cli/src/cli/__tests__/bounded-read.test.ts
+++ b/ix-cli/src/cli/__tests__/bounded-read.test.ts
@@ -1,0 +1,116 @@
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import * as fs from "node:fs";
+import * as os from "node:os";
+import * as nodePath from "node:path";
+
+import { readRepoFile } from "../bounded-read.js";
+
+/**
+ * `readRepoFile` is the one guard standing between ingestion and a file whose
+ * path AND contents the scanned repository chose. Every case below is a way a
+ * repo can name something that is not the small text file the caller expects.
+ */
+describe("readRepoFile", () => {
+  let root: string;
+
+  beforeAll(() => {
+    root = fs.mkdtempSync(nodePath.join(os.tmpdir(), "ix-bounded-"));
+  });
+
+  afterAll(() => {
+    fs.rmSync(root, { recursive: true, force: true });
+  });
+
+  it("reads an ordinary file", () => {
+    const p = nodePath.join(root, "plain.json");
+    fs.writeFileSync(p, '{ "name": "reads-fine" }');
+    expect(readRepoFile(root, p)).toBe('{ "name": "reads-fine" }');
+  });
+
+  it("refuses a file whose reported size is over the cap", () => {
+    const p = nodePath.join(root, "big.json");
+    fs.writeFileSync(p, "x".repeat(4096));
+    expect(readRepoFile(root, p, 1024)).toBeNull();
+    // ...and the same file is fine under a cap that admits it, so it is the cap
+    // doing the work rather than anything else about the file.
+    expect(readRepoFile(root, p, 8192)).not.toBeNull();
+  });
+
+  it("refuses a directory", () => {
+    const p = nodePath.join(root, "adir");
+    fs.mkdirSync(p, { recursive: true });
+    expect(readRepoFile(root, p)).toBeNull();
+  });
+
+  it("refuses a file that does not exist", () => {
+    expect(readRepoFile(root, nodePath.join(root, "absent.json"))).toBeNull();
+  });
+
+  /**
+   * The case the outer size check cannot catch, and the reason the read is
+   * capped separately.
+   *
+   * A `/proc` entry is a REGULAR file — `isFile()` is true — that reports size
+   * 0 while holding kilobytes. `fs.readFileSync(handle)` re-stats, sees 0, and
+   * falls back to reading 8 KB chunks until EOF with no limit, so a guard built
+   * only from `fstat().size` admits the whole file. Remove the cap inside
+   * `readCapped` and this goes green while nothing else in the suite moves.
+   */
+  it.skipIf(process.platform !== "linux")(
+    "refuses a size-0 regular file whose contents exceed the cap",
+    () => {
+      // Present on every Linux, stable, and comfortably over 100 bytes.
+      expect(fs.statSync("/proc/meminfo").isFile()).toBe(true);
+      expect(fs.statSync("/proc/meminfo").size).toBe(0);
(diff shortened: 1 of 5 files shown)


KageBinary:
Force-pushed `e0aea41`: **CodeQL was right and my first cut was wrong.**

It did `statSync(path)` and then `readFileSync(path)` — a check-then-use window, flagged as `js/file-system-race` at system.ts:80. I had reasoned myself into stat-then-read on the grounds that opening a FIFO blocks until a writer appears, so the guard must not be the thing that hangs.

`ts-module-resolution.ts` had already solved that, and better: open with `O_RDONLY | O_NONBLOCK` — which returns on a FIFO instead of waiting — then `fstat` **the handle**. Same open file, so there is no window at all, and the FIFO problem never arises. I had read that function and still missed the `O_NONBLOCK`.

So `readManifest` is now that shape exactly. The guards are unchanged in effect: `isFile()` refuses a device, FIFO or directory (a size check cannot — `/dev/zero` and a FIFO both report size 0), and `size` refuses the enormous file.

Both are load-bearing, and I checked by removing them rather than asserting it:

- without the size cap, the oversized-manifest test fails
- without `isFile()`, **the suite hangs on the `/dev/zero` case** until it is killed

1000 tests green, lint and typecheck clean.

KageBinary:
@josephismikhail — review please when you have a moment. Both this and #466 are the last two open before 0.10.0 GA, and both are green against the new main (verified merged, 1100 tests).

Three things I would point your eyes at, since the last round showed you read for semantics rather than shape:

1. **The guard is `isFile()`, not the size cap.** `/dev/zero` and a FIFO both report `size 0`, so a size check alone lets them through — the cap is only for the honest-but-enormous file. If `isFile()` is ever loosened this stops working silently.

2. **`O_NONBLOCK` is load-bearing and platform-conditional.** It is what makes opening a FIFO return instead of waiting for a writer. It is undefined on Windows, where it degrades to a plain read; Windows has no FIFO, so that is fine, but it does mean the Windows path is a different path.

3. **First cut was stat-then-read and CodeQL failed it** (`js/file-system-race`). I had reasoned myself into stat-then-read *because* opening a FIFO blocks — the answer to that is `O_NONBLOCK`, not avoiding the open, which `ts-module-resolution.ts` already knew and I missed on first read. Worth confirming the fd shape here matches that one's, since it is now the pattern in two places.

Both guards are checked by removal rather than asserted: without the cap the oversized test fails; without `isFile()` the suite hangs on the `/dev/zero` case until it is killed. The device and FIFO tests skip on Windows.

josephismikhail (approved):
Reviewed across three rounds, with the fixes pushed to this branch. The approach was right; two real defects sat inside the guard itself.

**The cap did not bind.** `fs.readFileSync(handle)` re-stats the handle and reads *that* size, not the one the fstat checked. Measured directly: an fstat that saw 50 bytes followed by a `readFileSync` on the same handle that returned 550. The size-0 case is the same mechanism — `/proc/meminfo` is a regular file, `isFile()` true, `stat` size 0, 1503 bytes of content, so it passed both guards and was read whole. The read is now performed here and capped, refusing rather than truncating.

**The guard is now shared.** `readObject` in `ts-module-resolution.ts` had the same defect and the same open/fstat/type/size sequence, so `bounded-read.ts` holds one copy and both callers use it. Containment stays with `readObject`, supplied as an `accept` predicate that runs on the same open handle before any content is read.

**One thing I added and then took back out.** I first gave the manifest path the containment check `readObject` has. That was wrong: rooted at the scanned directory it refuses manifests a repo legitimately symlinks within itself, and it does not refuse the case worth refusing, since a symlinked member directory resolves root and file to the same tree. Rooted at the mapped root instead, it refuses every symlinked member repo — which `detectSystem` has admitted on purpose since Ix#225. Either way a package silently leaves the registry, and in `hasWorkspaceMarker` a silent refusal shreds a monorepo into N member repos. So the manifest path keeps exactly the containment it had before this PR, which is none, and the comment now says why. Confining what a scanned repo may name is still worth doing — it needs a diagnostic and a way out, not a silent default bolted onto a read.

**Tests.** Every guard is mutation-checked, and the checks were moved off Linux-only footing. `accept` runs between the fstat and the read, so growing the file from inside it reproduces the exact defect through the public API: that one case kills both the revert to `readFileSync` and a reordering that reads before `accept` decides, on all four platforms (both previously survived on Windows). `isFile()` is pinned with `/dev/null`, the one non-regular file the cap cannot catch. The multi-byte test assumed the first read returns exactly 65536 bytes and would have passed vacuously on FUSE and network mounts; it now uses 3-byte characters. The FIFO test used to `return` when `mkfifo` was absent, reporting a pass having asserted nothing.

Not addressed here, worth a follow-up: `ingest.ts:1730` has the same `readFileSync`-re-stats shape (its size-0 variant is covered there only by accident, via an adjacent `st.size === 0` skip), and the FIFO test still hangs the runner rather than failing if `O_NONBLOCK` is dropped — confirmed, exit 124 — which is documented in the test but not designed away.

1015 green on Linux, all 20 checks green including all four Test platforms.

PR #466, opened 2026-08-19
fix(upgrade): stop telling people to upgrade a backend that is current

KageBinary:
`ix doctor` printed

```
✓ Backend is the released image — running the released image
```

while every command above it printed

```
  Backend update available
  Run: ix upgrade
```

## Why

`.backend-version` is written at install time (`install.sh`, `install.ps1`) and by `ix upgrade` — but **not by `ix docker start`**, which pulls `:latest` on every cold start. So anyone who took a newer image that way — or through docker compose — keeps a file naming the release they installed while running a later one, and is nagged on every command for ever.

## What this does

Records the version **at the pull**, where it is known, instead of trying to infer it afterwards. Right after `docker compose up -d --pull always` succeeds, the running image *is* the current backend release, so that release is simply what to write.

**This replaces an earlier version of this PR that inferred it from the running container.** That approach does not work, and the way it fails is worse than the bug: `checkBackendImage()` returning `ok` compares the container against the **locally pulled** `:latest`, and `docker image inspect` never contacts a registry. So `ok` proves only that nothing has been pulled since — equally true of a months-old image. On the machine this was found on, local `:latest` was built 2026-06-12 against a current release of 1.0.16: the notice would have been silenced *and* `.backend-version` stamped 1.0.16, after which `ix upgrade` reports "Backend already on the latest version" and never pulls again. That is the "current file in front of a stale container" case this PR exists to prevent, made permanent and silent, and it lands on the ordinary user who ran `ix upgrade` once and has not since.

No local signal proves currency: `/v1/health` reports `schema_version` and no release version, and the image's only label is the Ubuntu base's `22.04`.

Dropping the inference also removes what it cost. `checkForUpdate`'s cached branch has no `await` before it returns, so the deliberately-unawaited call in `main.ts` ran it **synchronously** on the startup path — up to five `docker` subprocesses, one of them `docker info` at a 10s timeout, before argv was parsed. That fired on every command for anyone whose tracked version read behind, which is everyone who has never run `ix upgrade` (the stamp defaults to `0.0.0`), and nothing repaired it for the non-`ok` states, so it repeated for ever.

## Where the errors fall

If the release cannot be fetched, the stamp is left alone: the user keeps a notice they may not need. The file can no longer claim a backend is current when it is not — which is the only direction worth being wrong in, since a stamp that reads new both hides a stale backend and stops `ix upgrade` from fetching it.

Someone driving raw `docker compose` still sees the notice. That is the honest answer — nothing here can see what they pulled — and the real cure is for the backend to report its release version in `/v1/health`, which would retire this file entirely. Worth a follow-up.

The existing `ix upgrade` stamp write now goes through the same helper, which adds the `mkdirSync` the other path was missing.

## Tests

The stamp writer is pinned where it can actually fail: it creates a missing `IX_HOME`, refuses to write a version it does not know (leaving an older stamp in place), and never throws into the command it is attached to. All mutation checked, plus a drift guard on the call site that fails both if the call is deleted and if it is moved to before the pull.

1002 tests green on Linux, lint, typecheck and knip clean.


KageBinary:
@josephismikhail — review please when you get a chance. Green against the new main (verified merged, 1102 tests), and the second of the last two before 0.10.0 GA.

The judgement calls worth your eye:

1. **`ok` is treated as the only state that proves anything.** `docker-unavailable`, `not-running` and `latest-not-pulled` all leave the question open, so the tracked version stays the answer there and the notice behaves as it always did. `local-build` and `digest-mismatch` still get the notice. If you think an open question should suppress rather than nag, that is a one-line change and the pure function is where it lives.

2. **It writes to `~/.ix/.backend-version` as a side effect of a *check*.** Finding the container already on `:latest` repairs the file it disagreed with, so the docker calls are paid once rather than on every command. That is the part I would push back on if I were reviewing: a read path that writes. My argument for it is that the file has exactly one writer today and it is wrong, and the repair value is not a guess — the running image *is* the latest release. Yours may differ.

3. **Cost.** The docker calls happen only once the tracked version has already said "behind", so a correct tracker costs nothing. `checkForUpdate` is called on every command, hence the ordering.

The motivating case is verified end to end on my machine — tracked `1.0.13`, container on `:latest`, every command nagging while `ix doctor` two lines later said "running the released image". Afterwards: no nag, file reads `1.0.16`. The reverse case is the one I actually care about though: a current file in front of a stale container reported "up to date" while an old build served the graph.

josephismikhail (approved):
Reviewed across thirteen rounds, with the fixes pushed to this branch. The diagnosis in the original description was right; the mechanism was not, and replacing it took most of the work.

**Why the original approach could not stand.** `checkBackendImage()` returning `ok` compares the running container against the **locally pulled** `:latest` — `docker image inspect` never contacts a registry — so `ok` proves only that nothing has been pulled since, which is equally true of a months-old image. Measured on the machine this was found on: local `:latest` built 2026-06-12 against a current release of 1.0.16. That container reports `ok`, so the notice would be silenced *and* `.backend-version` stamped 1.0.16, after which `ix upgrade` says "already on the latest version" and never pulls. The "current file in front of a stale container" case this PR set out to prevent, made permanent and silent, landing on the ordinary user who ran `ix upgrade` once.

Dropping the inference also removed what it cost: `checkForUpdate`'s cached branch has no `await` before it returns, so the deliberately-unawaited call in `main.ts` ran it **synchronously** — up to five `docker` subprocesses, one of them `docker info` at a 10s timeout — before argv was parsed, on every command, for everyone whose stamp read behind.

**What ships instead.** Record the version at the pull, where it is known. `ix docker start` runs `--pull always`, so at that instant the running image *is* the current release. 65 lines, gated on the compose actually declaring `ix-memory-layer:latest` — `findComposeFile()` falls back to any `docker-compose.yml` in the working directory, so without that gate a pinned tag, a digest or a local build would be stamped as the latest release, which is the same inversion from the other side.

Errors fall the safe way: a failed fetch leaves the stamp alone and the user keeps a notice they may not need. The file can no longer claim a backend is current when it is not.

**Corrections to the description, since they were load-bearing.** "`.backend-version` is written by `ix upgrade` and by nothing else" is false — `install.sh:1069` and `install.ps1:539` both stamp at install time. The real gap is narrower: no path recorded a pull that was not `ix upgrade`.

**Not fixed here, worth follow-ups:**
- `ix docker start` on an already-healthy backend returns before pulling, so it does not stamp — correct, but the nag then clears only via `ix upgrade`.
- The awaited stamp can spend up to 10s silently between the compose output and "Waiting for services…" on a network that blackholes GitHub.
- `ix upgrade` stamps after `docker pull` but restarts only conditionally, so a non-standard compose project can advance the stamp without the container being recreated.
- The stamp write is not atomic, and the fetch does not seed the version cache, so a cold start spends a request `checkForUpdate` then spends again.
- `mcp-install.test.ts` and `config-file-mode.test.ts` each hand-roll an unmeasured version of the file-mode predicate this PR measures.

**The real cure is upstream.** `/v1/health` reports `schema_version` and no release version, which is the only reason any of this inference exists. If the backend reported its own version, `ix doctor` and the notice would both just ask it, every population is fixed including raw `docker compose` users, and `.backend-version` becomes vestigial.

47 mutations across the review rounds, each killed by an identified test. 1036 green on Linux; the file's own 41 run on Windows too, and skip with a reported reason under root.

PR #468, opened 2026-08-20
chore(release): 0.10.1

KageBinary:
Bumps `ix-cli` to **0.10.1**.

Retargeted from 0.10.0, which this PR originally carried: v0.10.0 shipped while it sat in review. The release workflow runs `npm version "$VERSION"` at build time (`release.yml:205`), so that tag still produced a correct artifact — what an unbumped tree costs is a *source checkout* of `main`, which has reported 0.9.3 since GA.

## The trip-wire this also removes

`watch-dedup.test.ts` asserted the literal `"0.9.3"` from `ix --version`, so **every** `chore(release)` bump turned it red — including this one. It reads `package.json` now.

That test is about whether the watch child *starts*; the number it prints is incidental, which is why reading the value is not a weaker assertion than pinning it. The same value seeds `.version-check.json` so the update notice cannot appear on stdout and be mistaken for the version — see the note in the diff.

The change needed no retargeting of its own, which is the point of it: it is version-agnostic by construction, so 0.10.2 will not reopen this.

## Verified

- `watch-dedup.test.ts` — 15/15 green against the 0.10.1 tree
- The bump is `npm version`'s own output, so `package-lock.json` tracks it

## Note on the two commits

The retarget is a second commit rather than an amend, because rewriting the branch was not available to me here. It squash-merges to a single `chore(release): 0.10.1`, so nothing about the merged history changes.

Not a blocker for tagging v0.10.1 — a tag alone is sufficient, as v0.10.0 demonstrated. This is about `main` telling the truth about itself.

### DIFF
--- a/ix-cli/package-lock.json
+++ b/ix-cli/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@ix/cli",
-  "version": "0.10.0",
+  "version": "0.10.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@ix/cli",
-      "version": "0.10.0",
+      "version": "0.10.1",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.30.0",
         "chalk": "^6.0.0",
--- a/ix-cli/package-lock.json
+++ b/ix-cli/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@ix/cli",
-  "version": "0.9.3",
+  "version": "0.10.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@ix/cli",
-      "version": "0.9.3",
+      "version": "0.10.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.30.0",
         "chalk": "^6.0.0",
--- a/ix-cli/package.json
+++ b/ix-cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ix/cli",
-  "version": "0.10.0",
+  "version": "0.10.1",
   "type": "module",
   "description": "Ix Memory — Persistent Memory for LLM Systems",
   "main": "dist/index.js",
--- a/ix-cli/package.json
+++ b/ix-cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ix/cli",
-  "version": "0.9.3",
+  "version": "0.10.0",
   "type": "module",
   "description": "Ix Memory — Persistent Memory for LLM Systems",
   "main": "dist/index.js",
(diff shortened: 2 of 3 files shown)


KageBinary:
Re-verified: green, and the `watch-dedup` change is the part worth keeping regardless of which version ships. Reading `version` from `package.json` instead of the `"0.9.3"` literal is what stops every future `chore(release)` bump failing a test that is about whether the child process *starts*.

**One thing to settle before tagging, though:** this bumps to **0.10.1**, which was decided when 0.10.1 was going to be #477 alone. There are now nine other PRs open and reviewed-clean (#480, #482, #484, #486, #488, #490, #491, #492, #475). If they land too, 0.10.1 is a much larger release than the version number advertises — #488 in particular changes what `ix context` returns on the default path from 1 entity to 41.

That's a call for @ix-infrastructure rather than a defect in this PR: either land #477 alone as 0.10.1 as planned and re-bump afterwards, or land the batch and make this 0.11.0. Flagging it because the number is easy to merge without re-deciding.

Verified with everything merged together: 1348 passed / 2 skipped, typecheck clean, lint 0 errors, knip 0.

KageBinary:
Second pass — I audited whether this bump is **complete**, since a half-bumped release is the failure mode that matters here.

Grepped every `*.json`/`*.ts`/`*.sh`/`*.ps1`/`*.yml`/`*.rb`/`*.md` outside `node_modules` and `dist` for version literals. Everything that remains is either a deliberate test fixture (`upgrade-compass-stamp`, `upgrade-check-closing-line`, `view-server`), a workflow comment, or `Formula/ix.rb` — which tracks the *released* tag and is #475's job, not this one.

Every version literal in non-test source is the `"0.0.0"` sentinel:
```
main.ts:30        let cliVersion = "0.0.0";
main.ts:34        cliVersion = pkg.version || "0.0.0";
upgrade.ts:69     return pkg.version || "0.0.0";
mcp/server.ts:494 startIxMcpServer(version = "0.0.0")
```

So the version is read from `package.json` at runtime and **`package.json` + `package-lock.json` are genuinely the only two files that need to change**. No install script pins a version. The bump is complete.

The `watch-dedup` change is the durable part: reading `version` from `package.json` instead of the `"0.9.3"` literal is what stops every future `chore(release)` failing a test that is about whether the child process *starts*.

**Version question, now settled:** `v0.10.0` is the newest tag and nothing was cut after it, so **0.10.1 is free and this PR already sets it** — no extra bump commit needed even if the whole batch lands. Every PR in the batch is a bug fix, so patch is right; 0.11.0 would be wrong, since #488 restores output that was supposed to be there rather than adding a feature.

Verified with the full batch merged: 1367 passed / 2 skipped, typecheck clean, lint 0 errors, knip 0.

PR #474, opened 2026-08-21
fix(view): prevent __ix routes from serving HTML for API requests

### LINKED ISSUE #473: Read, Explain, Inventory and Help do nothing in `ix view`
The detail panel's **Read**, **Explain**, **Inventory** and **Help** actions have never worked
in a released build. Clicking them does nothing and reports nothing.

## What happens

Compass posts to `/__ix/read`, `/__ix/explain`, `/__ix/inventory`, `/__ix/help`
(`system-compass/src/lib/ix-client.ts`). Those routes are implemented **only** in
`system-compass/vite.ix-local-bridge.ts`, which `vite.config.ts` registers behind
`mode === "development"`:

```js
plugins: [
  react(),
  mode === "development" && componentTagger(),
  mode === "development" && ixLocalBridgePlugin(__dirname),
].filter(Boolean),
```

The server in `ix-cli/src/cli/commands/view.ts` has exactly three route branches — `/v1*`
proxied to the backend, `/__ix/remap`, and the static file handler. So the other four fall
through to static and are answered by `index.html`.

Verified against a live `ix view` on rc.15:

```
POST /__ix/read       -> 200   (body is <!doctype html>…)
POST /__ix/explain    -> 200
POST /__ix/help       -> 200
POST /__ix/inventory  -> 200
POST /__ix/remap      -> handled
```

**200, not 404**, which is what makes it silent: `post<IxReadResponse>()` gets a successful
response, `JSON.parse` throws on the HTML, and `handleRead`'s `.catch()` sets `readData = null`
without surfacing anything. The user sees a button that does nothing.

## Not a regression

Broken in every release including 0.9.x — `ix view` never implemented these. It works under
`npm run dev`, which is presumably why it shipped.

## Fixing it

Port the four handlers out of the Vite plugin into `view.ts`. They shell out to the CLI, so
they need the same containment `ix read` got in #424 — the view server is reachable from the
browser and the node id, path and line range all arrive from the page.

Worth doing at the same time: the static handler should not answer `/__ix/*`. A missing route
there must 404 so a client can tell the difference between "not implemented" and "here is a
web page". A 200 of HTML is the reason this went unnoticed.

Found while testing v0.10.0-rc.15. Deferred out of 0.10.0 deliberately.

Alot1z:
## Summary

Fixes #473 — `/__ix/read`, `/__ix/explain`, `/__ix/inventory`, and `/__ix/help` routes in the view server now return 404 with JSON instead of silently serving `index.html`.

## What happens

The Compass detail panel sends POST requests to `/__ix/*` routes. In the production view server (`view.ts`), these routes fell through to the SPA static handler, which served `index.html` with a 200 status. The client tried to `JSON.parse` the HTML, caught the error silently, and the user saw nothing — buttons that did nothing.

## Root cause

`view.ts` had exactly three route branches:
- `/v1*` — proxied to backend
- `/__ix/remap` — handled (POST only)
- Static file handler — serves `index.html` for SPA fallback

The four action routes had no handler and fell through to the SPA fallback, which served HTML for any unrecognized path.

## Fix

Added a 404 handler between the `/__ix/remap` POST handler and the static file handler. Any `/__ix/*` route that isn't `POST /__ix/remap` now returns:

```json
{"ok": false, "error": "not found: /__ix/read"}
```

with `404` status and `Cache-Control: no-store`.

## Changes

- **`ix-cli/src/cli/commands/view.ts`** — Added `/__ix/*` 404 handler (8 lines)
- **`ix-cli/src/cli/__tests__/view-routes.test.ts`** — New test file with 20 tests covering route matching, input validation, response format, edge cases, and SPA fallback
- **`ix-cli/src/cli/__tests__/view-server.test.ts`** — Updated existing test to expect 404 for GET `/__ix/remap` (was incorrectly expecting SPA fallback)

## Testing

- 20 new tests in `view-routes.test.ts` — all pass
- 1241 total tests pass, 0 failures, 18 skipped (pre-existing)
- Full test suite: 74/74 files pass

## Validation

- npm run typecheck — clean
- npx vitest run — 1241 passed, 0 failed

### DIFF
--- a/ix-cli/src/cli/__tests__/context-mode-conflict.test.ts
+++ b/ix-cli/src/cli/__tests__/context-mode-conflict.test.ts
@@ -7,11 +7,14 @@ import { Command } from "commander";
 import {
   detectContextModeConflict,
   registerContextCommand,
+  sanitizeId,
+  mergeDiffOptions,
 } from "../commands/context.js";
 import {
   detectDiffModeConflict,
   registerDiffCommand,
 } from "../commands/diff.js";
+import { reportFailure } from "../ui.js";
 
 /**
  * C-1..C-4 silent-ignore flag gaps in `ix context` and C-5 in `ix diff` are now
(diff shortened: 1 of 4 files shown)


KageBinary:
Thanks — the change in `view.ts` itself is sound, and the tests for it are honest. I mutation-checked them: delete the handler and 6 of the 20 go red, which is what a test file should do. Returning a JSON 404 is strictly better than serving `index.html` to an API path.

Two things before it lands, one of them about the framing rather than the code.

## It doesn't fix #473, and shouldn't close it

The PR is titled as a fix for the dead Read/Explain/Inventory/Help actions, but the routes still aren't implemented — this changes *how they fail*, not *whether they work*. I traced all four through the Compass client. `post()` already throws `IxApiError` on `!res.ok`, and `errorMessageFrom` already has a comment anticipating "an HTML body from the SPA fallback", so the client was half-ready for this:

| action | today (200 + HTML) | with this PR |
|---|---|---|
| Explain | JSON parse error shown in the panel | `not found: /__ix/explain` shown |
| Inventory | JSON parse error shown in the panel | `not found: /__ix/inventory` shown |
| **Read** | `.catch(() => {})` — silent, nothing | **still silent, nothing** |
| **Help** | `.catch(() => {})` — static fallback list | **unchanged** |

So two of the four get a better message, and the two that are *genuinely* silent — the ones the description is describing when it says the user "saw nothing" — are exactly the two this doesn't reach. That's not a reason to reject the change; it's a reason to retitle it to what it does (make the failure visible and diagnosable) and leave #473 open for the actual route implementations. Please drop the `Fixes #473` line.

If you want to take the real fix too, that's a bigger and more welcome piece of work: the four routes exist only in the Vite dev middleware and need equivalents in the static server that `ix view` runs.

## It carries #472's diff

`context-mode-conflict.test.ts` comes along with this branch, including the three `ix diff edge-case integration` tests that make live HTTP calls. Measured on a checkout of this branch with a backend running: **100/103, three 5s timeouts**. Details are in my review on #472 — this branch inherits all of it. Once #472 is sorted, rebase this on top so the diff here is just `view.ts` and `view-routes.test.ts`.

## Please test this heavily before pushing again

The unit tests here are good, but they only prove the server returns a 404. What matters to a user is what the Compass UI does with it, and that's untested:

1. **Exercise it in a real `ix view`,** not just vitest. Build the CLI, run `ix view` against a mapped project, and click Read, Explain, Inventory, and Help. Confirm what each one actually shows now — I'd expect a visible error on two and unchanged silence on two, and if you see something different than the table above I want to know.
2. **Run the suite both ways** — with a backend reachable and with `IX_ENDPOINT=http://127.0.0.1:9`. Our CI only ever sees the second, so a test that depends on the first is green upstream forever and red on your machine.
3. **Mutation-check anything you add.** Break the branch the test names and confirm it goes red. Your existing `view-routes.test.ts` already passes this bar — keep that standard for whatever comes next.
4. **Check `POST /__ix/remap` still works** end to end after the handler change, since it's the one `/__ix/` route that does have an implementation and it now sits next to a catch-all 404.

The `view.ts` change is worth having on its own merits. Retitle, rebase, and verify in the running app and I'm happy to take it.

KageBinary:
Reviewed. The `view.ts` change is **correct and worth having** — I've rebased it onto `main` as #492, carrying your commit `e6f4a86` unchanged (I can't push to your fork from here, so it had to be a separate PR; the credit is yours). Two scope problems kept it from merging as-is.

**1. `Fixes #473` is not accurate, and merging on that basis would close an unfixed bug.**

`view.ts` implements exactly one `/__ix/*` route:

```
427:  if (pathname === "/__ix/remap" && req.method === "POST") {
```

`/__ix/read`, `/__ix/explain`, `/__ix/inventory` and `/__ix/help` still do not exist. Your change makes them return a JSON 404 instead of `index.html`, so the client's `JSON.parse` fails loudly instead of throwing into a silent catch. That is real hardening and I want it — the four buttons now fail *visibly* rather than doing nothing — but they still don't work, so #473 has to stay open. #492 drops the closing keyword and says so.

**2. It carried an older copy of #472's tests.**

Commits `f9883ae` and `2e79fa8` duplicate the conflict-detection tests that #472 also has, at an earlier revision (403 lines here vs 557 there). Merging both would strand or revert the newer work. #492 is the view fix and its tests alone, so it no longer depends on #472 and the two can land in any order.

**Verification:** mutation-checked — disabling the `startsWith("/__ix/")` guard turns 16 tests in `view-routes.test.ts` red, so the coverage is pinned to the branch and not to the response body. I also noticed the `GET /__ix/remap` test flipped from asserting a 200 SPA fallback to a 404; that's a genuine meaning change rather than a test bent to pass, since `remap` is POST-only and a GET of it is exactly the unmatched case.

KageBinary:
Superseded by #492, which returns a JSON 404 for unimplemented `__ix` routes instead of letting the static handler answer them with index.html.

Same defect, and #492 additionally covers the Vite-dev-only route gap that made the JSON parse throw into a silent catch. Closing in favour of #492.

Alot1z:
Agreed with the supersession — and the framing lesson is fair: the change made the four dead actions fail visibly instead of doing nothing, it did not implement them, so closing #473 on that basis would have been wrong. Keeping the issue open and covering the Vite-dev-only route gap in #492 was the correct scope. Thanks for carrying `e6f4a86`.

PR #477, opened 2026-08-21
fix(upgrade): stop `--check` calling an out-of-date install up to date

### LINKED ISSUE #476: `ix upgrade --check` says "ix is up to date" when it is not
## What happens

On a CLI that is behind, `ix upgrade --check` reports the new version and then contradicts itself:

```
$ ix upgrade --check
Current version: 0.9.2
Checking for updates...
New CLI version available: 0.10.0
[ok] Backend already on the latest version (1.0.17)
[ok] Compass already on the latest version (0.9.2)

[ok] ix is up to date        <-- it is not
```

Reproduced against the shipped **v0.10.0** binary as well, by pointing it at a
`package.json` reporting an older version — so this is the release's own code,
not only the older install that surfaced it:

```
Current version: 0.9.0
New CLI version available: 0.10.0
...
[ok] ix is up to date
```

## Why

`upgrade.ts:1719` prints the closing line unconditionally at the end of the
action, with no reference to `cliUpToDate` (computed at :1328) or to whether
anything was installed:

```ts
console.log("");
console.log("[ok] ix is up to date");
```

In the non-`--check` path the line is correct — by then the upgrade has actually
run. It is only wrong for `--check`, whose entire job is to answer this question.

## Why it matters now

Not a v0.10.0 regression; 0.9.x has it too. But GA moved `/releases/latest` to
v0.10.0, so this is the first time a large number of installs are behind — which
makes `--check` the command they will run, and the last line the one they will
believe.

## Suggested fix

Make the closing line reflect state rather than reaching the end of the function:
say "up to date" only when the CLI, backend and compass were all already current,
and otherwise close with what is available and how to get it (`Run: ix upgrade`).
A test that runs `--check` against a stubbed newer release and asserts the
absence of "up to date" would have caught it.

KageBinary:
Closes #476.

## The bug

`ix upgrade --check` closed with `[ok] ix is up to date` unconditionally — two
lines under `New CLI version available: 0.10.0`:

```
Current version: 0.9.2
New CLI version available: 0.10.0
[ok] Backend already on the latest version (1.0.17)
[ok] Compass already on the latest version (0.9.2)

[ok] ix is up to date        <-- it is not
```

`--check` exists to answer one question — is there anything to install? — and it
answered it wrong for every user who was behind.

## After

```
Current version: 0.9.3
New CLI version available: 0.10.0
Backend update available: none → 1.0.17
Compass update available: none → 0.3.0

[!!] Update available: CLI 0.9.3 → 0.10.0, backend none → 1.0.17, Compass none → 0.3.0
  Run: ix upgrade
```

A current install is unchanged: `[ok] ix is up to date`. Both verified against
the live v0.10.0 release, not a stub.

## The install path deliberately keeps the flat line

`ix upgrade` without `--check` still closes with `[ok] ix is up to date`, and
that is not an oversight. Every branch that adds to `outstanding` has already
acted on it by the time the line prints, so reusing the list there would report
the work that just succeeded as still pending.

## Tests

Five, in `upgrade-check-closing-line.test.ts`, covering the rule (`closingStatus`)
and the wiring (the real command, driven through commander with `fetch` stubbed
and a throwaway `IX_HOME`).

Mutation-checked both ways:

| mutation | result |
|---|---|
| restore the unconditional closing line | ✗ red |
| keep the fix, drop the CLI branch's `outstanding.push` | ✗ red |

The second one matters. The first version of that test asserted only
`toContain("Update available")` and **passed** with the CLI branch broken — a
throwaway `IX_HOME` has no backend or compass stamp either, so those two are
outstanding regardless and kept the line on screen. It now matches the closing
line itself and requires the CLI named in it, keyed to the fetched version rather
than the current one so a release bump does not rot it.

## Not a 0.10.0 regression

0.9.x has the same bug. GA moving `/releases/latest` to v0.10.0 is what turned it
from a rare case into the one most installs now hit, which is why it is worth a
patch release rather than waiting.

### DIFF
--- a/ix-cli/src/cli/__tests__/upgrade-check-closing-line.test.ts
+++ b/ix-cli/src/cli/__tests__/upgrade-check-closing-line.test.ts
@@ -1,0 +1,99 @@
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import { closingStatus } from "../commands/upgrade.js";
+
+/**
+ * `ix upgrade --check` used to close with `[ok] ix is up to date` no matter what
+ * it had just found, two lines under `New CLI version available: 0.10.0` (#476).
+ *
+ * The decision is tested twice over, and deliberately so. `closingStatus` covers
+ * the rule; the command test below covers the wiring, because a branch that
+ * forgets to record what it found leaves the rule correct and the output wrong —
+ * which is the shape of the original bug.
+ */
+describe("closingStatus", () => {
+  it("refuses to call an out-of-date install up to date under --check", () => {
+    const status = closingStatus(true, ["CLI 0.9.2 → 0.10.0"]);
+    expect(status.upToDate).toBe(false);
+    expect(status).toHaveProperty("summary", "CLI 0.9.2 → 0.10.0");
+  });
+
+  it("lists every outstanding component, not just the first", () => {
+    const status = closingStatus(true, ["CLI 0.9.2 → 0.10.0", "backend 1.0.16 → 1.0.17"]);
+    // Reporting only the CLI would still be "not up to date" and still pass a
+    // looser assertion, while hiding the two components the user must also pull.
+    expect(status).toHaveProperty("summary", "CLI 0.9.2 → 0.10.0, backend 1.0.16 → 1.0.17");
+  });
+
+  it("says up to date under --check when nothing is outstanding", () => {
+    expect(closingStatus(true, []).upToDate).toBe(true);
+  });
+
+  it("keeps the flat line on an install run, which has already acted", () => {
+    // Not a shortcut: on this path every entry in the list has been installed by
+    // the time the line prints, so reporting them as outstanding would describe
+    // the work that just succeeded as still pending.
+    expect(closingStatus(false, ["CLI 0.9.2 → 0.10.0"]).upToDate).toBe(true);
+    expect(closingStatus(undefined, ["CLI 0.9.2 → 0.10.0"]).upToDate).toBe(true);
+  });
+});
+
+describe("ix upgrade --check output", () => {
+  let home: string;
+  let logs: string[];
+
+  beforeEach(() => {
+    home = mkdtempSync(join(tmpdir(), "ix-upgrade-check-"));
+    process.env.IX_HOME = home;
+    logs = [];
+    // console, not process.stdout.write: vitest replaces the console object, so
+    // patching the stream underneath it captures nothing and the assertion can
+    // never fail.
+    vi.spyOn(console, "log").mockImplementation((...args: unknown[]) => {
+      logs.push(args.map(String).join(" "));
+    });
+  });
+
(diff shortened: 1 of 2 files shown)


KageBinary:
This gates **v0.10.1** — the tag goes out as soon as it is approved, and nothing else is queued for that release.

Worth a look at in review: the second mutation check in the description. The first version of the command test asserted only `toContain("Update available")` and passed with the CLI branch's `outstanding.push` deleted, because a throwaway `IX_HOME` has no backend or compass stamp either and those two kept the line on screen. It now matches the closing line itself and requires the CLI named in it.

CI is 20/20 green. Verified against the live v0.10.0 release rather than a stub, in both directions: an install that is behind, and one that is current.

KageBinary:
Re-verified on the current tip, and confirmed the fix end-to-end against a cold `IX_HOME` rather than only in tests:

```
$ IX_HOME=$(mktemp -d) ix upgrade --check
[ok] CLI already on the latest version (0.10.1)
Backend update available: none → 1.0.17
Compass update available: none → 0.3.0

[!!] Update available: backend none → 1.0.17, Compass none → 0.3.0
  Run: ix upgrade
```

That is #476 gone: the closing line no longer contradicts the two lines above it.

## ⚠️ This conflicts with #480 — here is the resolution

Both PRs rewrite the same closing block, so whichever merges second hits a conflict in `upgrade.ts`. They're complementary. I merged both, resolved it, and ran the suite; use this rather than guessing.

**Hunk 1** — keep #480's two predicates and put #477's `outstanding.push` inside:

```ts
const backendUpgradeNeeded = !!backendLatest && backendUpdateAvailable(backendLatest, backendCurrent);
const backendUpgradeSkipped = !opts.check && !backendComposeAllowsUpgrade(backendComposeFile);

if (backendUpgradeNeeded) {
  // Recorded even when the compose file is pinned: `outstanding` is only
  // read under --check, and --check never sets backendUpgradeSkipped, so
  // the two never disagree about the same run.
  outstanding.push(`backend ${backendCurrent === "0.0.0" ? "none" : backendCurrent} → ${backendLatest}`);
  console.log(
    `Backend update available: ${backendCurrent === "0.0.0" ? "none" : backendCurrent} → ${chalk.green(backendLatest)}`
  );
}
```

**Hunk 2** — three-way sign-off, and **the order is load-bearing**:

```ts
const closing = closingStatus(opts.check, outstanding);
if (backendUpgradeSkipped) {
  console.log("[!!] ix upgrade finished with the backend unchanged");
} else if (closing.upToDate) {
  console.log("[ok] ix is up to date");
} else {
  console.log(chalk.yellow(`[!!] Update available: ${closing.summary}`));
  console.log(chalk.dim("  Run: ix upgrade"));
}
```

`backendUpgradeSkipped` must be tested **first**. It can only be true on an install run, where `closingStatus` returns `upToDate: true` by design — so testing it second silently swallows #480's line, and #480's own test (`expect(output).not.toContain("[ok] ix is up to date")`) is what catches you.

**Verified:** with both merged and resolved this way, all 138 upgrade tests from both PRs pass, and the full suite is 1348 passed / 2 skipped with every other open PR merged in too.

PR #480, opened 2026-08-22
fix(upgrade): respect pinned backend compose

### LINKED ISSUE #479: [bug] ix upgrade stamps latest while pinned compose restarts an older backend
## Description

`ix upgrade` always pulls `ix-memory-layer:latest` and advances `.backend-version`, even when the existing backend compose file pins an older tag or digest. It then restarts that pinned image and reports that the latest backend is running.

This is not a regression in the v0.10.0 stock compose, which tracks `:latest`. It affects supported user-edited compose files, a case `ix docker start` already detects before stamping.

## Steps to reproduce
1. Put a fixed tag or digest for `memory-layer` in `~/.ix/backend/docker-compose.yml`.
2. Set `.backend-version` behind the current backend release and start the backend.
3. Run `ix upgrade`.
4. Compare the running image ID and health response with the version stamp.

## Expected behavior

Ix should not advance the backend stamp or claim a latest-image restart when the compose file does not track the released `:latest` image. Existing affected installs whose stamp is already ahead should also not be reported as fully up to date.

## Actual behavior

The latest tag is pulled and stamped, but Docker Compose recreates the service from the pinned image. Subsequent update checks trust the advanced stamp.

## CLI output

```text
Backend update available: 1.0.16 → 1.0.17
[ok] Backend image updated to 1.0.17
[ok] Backend restarted with latest image
[ok] ix is up to date
```

## Environment
- OS: macOS
- Node: v26.7.0
- Ix version: v0.10.0
- Backend releases tested: 1.0.16 and 1.0.17

Hiro-Chiba:
Closes #479.

## Summary
Respect a user-managed backend compose file when `ix upgrade` encounters a pinned tag or digest.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Gate backend pulls, stamp updates, and restart success claims on the compose file tracking the released `:latest` image.
- Keep pinned or unreadable compose files unchanged and report that the backend was not upgraded.
- Avoid reporting an already-misstamped pinned installation as fully up to date.
- Cover pinned tags, pinned digests, the already-advanced stamp case, and the managed `:latest` path.

## Validation
- Targeted upgrade tests
- `npm test`
- `npm run typecheck`
- ESLint on changed files
- `git diff --check`

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/upgrade-pinned-compose.test.ts
+++ b/ix-cli/src/cli/__tests__/upgrade-pinned-compose.test.ts
@@ -1,0 +1,145 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+import { Command } from "commander";
+
+const mocks = vi.hoisted(() => ({
+  execFileSync: vi.fn((..._args: unknown[]) => ""),
+}));
+
+vi.mock("child_process", () => ({ execFileSync: mocks.execFileSync }));
+vi.mock("node:child_process", () => ({ execFileSync: mocks.execFileSync }));
+
+const TEST_DIR = dirname(fileURLToPath(import.meta.url));
+const CURRENT_CLI_VERSION = JSON.parse(
+  readFileSync(join(TEST_DIR, "..", "..", "..", "package.json"), "utf-8"),
+).version as string;
+const BACKEND_IMAGE = "ghcr.io/ix-infrastructure/ix-memory-layer";
+
+let home: string;
+let originalIxHome: string | undefined;
+
+beforeEach(() => {
+  vi.resetModules();
+  vi.clearAllMocks();
+
+  originalIxHome = process.env.IX_HOME;
+  home = mkdtempSync(join(tmpdir(), "ix-upgrade-pinned-compose-"));
+  process.env.IX_HOME = home;
+  mkdirSync(join(home, "backend"), { recursive: true });
+  writeFileSync(join(home, ".backend-version"), "1.0.16");
+
+  vi.stubGlobal(
+    "fetch",
+    vi.fn(async (input: string | URL | Request) => {
+      const url = String(input);
+      if (url.includes("/ix-memory-layer-dist/")) {
+        return new Response(JSON.stringify({ tag_name: "v1.0.17" }), { status: 200 });
+      }
+      if (url.includes("/ix-compass-dist/")) {
+        return new Response("", { status: 404 });
+      }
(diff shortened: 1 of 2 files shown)


KageBinary:
Reviewed. Failing closed on an unreadable compose file is the right call, and I confirmed the normal path is unaffected — both `docker-compose.standalone.yml` and the installed `~/.ix/backend/docker-compose.yml` declare `image: ghcr.io/ix-infrastructure/ix-memory-layer:latest`, so `composeTracksLatestBackend` is true for every stock install and only a hand-pinned file takes the new branch.

Suppressing `[ok] Backend already on the latest version` for a pinned compose is the subtle half and it's right: the stamp records what `ix upgrade` last pulled, which says nothing about a compose that pulls something else.

**Mutation-checked**: forcing `backendUpgradeSkipped = false` turns 3 of the 5 new tests red. The two that survive are the `--check` case (skip is `!opts.check`-gated by design) and the `:latest` control — both correct.

⚠️ **Merge-order note: this conflicts with #477.** Both rewrite the same closing block. They're complementary, not competing, and I've resolved and tested the combination — the resolution is in a comment on #477. Whichever lands second needs that hunk.

Minor, non-blocking: a user who pins deliberately now ends every `ix upgrade` on `[!!] ix upgrade finished with the backend unchanged` even when nothing was due. That's truthful and exit code stays 0, so I'd leave it.

138/138 upgrade tests pass. No changes requested.

KageBinary:
Second, heavier pass — I ran a **real `ix upgrade`** (not `--check`, not a mock) against a throwaway `IX_HOME` holding a digest-pinned compose, with the live backend running:

```
Backend update available: 1.0.16 → 1.0.17
[!!] Backend not upgraded: …/backend/docker-compose.yml does not track ghcr.io/ix-infrastructure/ix-memory-layer:latest.
     Left the compose file and backend version stamp unchanged.
[ok] Compass already on the latest version (0.9.2)

[!!] ix upgrade finished with the backend unchanged
```

Verified afterwards:
- compose file **byte-identical**
- `.backend-version` still `1.0.16`
- **the running container was not restarted** — `Up 3 hours (healthy)` before and after

That last one is the property that actually matters, and it is the one a mocked `execFileSync` cannot demonstrate.

I also ran `--check` against the same pinned home, because that is where the merge with #477 could go wrong:

```
[!!] Update available: backend 1.0.16 → 1.0.17
  Run: ix upgrade
```

Correct: `--check` never sets `backendUpgradeSkipped`, so it reports what is outstanding and does **not** print the backend-unchanged line. Both PRs' behaviours coexist under the resolution posted on #477, confirmed with real network calls rather than fixtures.

**Deliberately not run:** the `:latest` counterpart. `$IX_HOME/backend` has basename `backend`, so `docker compose -f …/backend/docker-compose.yml up -d` would derive the project name `backend` — the same project as the running stack — and restart it. That path is covered by the mutation-checked unit test and is unchanged from `main`.

Still no changes requested.

PR #482, opened 2026-08-22
fix(doctor): inspect backend behind compose proxy

### LINKED ISSUE #481: [bug] ix doctor inspects a compose proxy instead of memory-layer
## Description

`ix doctor` identifies the first container publishing port 8090 as the memory-layer backend. In a compose setup where nginx publishes the loopback ports and `memory-layer` stays on the internal network, doctor inspects nginx and reports it as an unreleased local backend image.

The stock v0.10.0 compose publishes memory-layer directly. This report concerns proxy-fronted compose files.

## Steps to reproduce
1. Run a compose project with a `loopback-proxy` service publishing 8090 and an internal `memory-layer` service.
2. Ensure the memory-layer container uses `ghcr.io/ix-infrastructure/ix-memory-layer:latest`.
3. Run `ix doctor --format json`.

## Expected behavior

Doctor should follow the publishing container's compose project to the `memory-layer` service and compare that image with the released image.

## Actual behavior

Doctor compares the nginx image with the memory-layer release and emits a false local-build warning.

## CLI output

```text
Backend is the released image: warning
running a local build (nginx@sha256:...), not the released image
```

## Environment
- OS: macOS
- Node: v26.7.0
- Ix version: v0.10.0
- Backend: 1.0.17

Hiro-Chiba:
Closes #481.

## Summary
Make backend image checks inspect the memory-layer service behind a compose port proxy.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Prefer a directly published released backend when present.
- When 8090 is published by another service, follow its compose project to `com.docker.compose.service=memory-layer`.
- Preserve local-build warnings for directly published development backends.
- Add a regression test for a proxy-fronted compose project.

## Validation
- Backend container inspection and backend status tests
- `npm test`
- `npm run typecheck`
- ESLint on changed files
- `git diff --check`

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/backend-container-inspection.test.ts
+++ b/ix-cli/src/cli/__tests__/backend-container-inspection.test.ts
@@ -1,0 +1,72 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const execFileSync = vi.hoisted(() => vi.fn());
+
+vi.mock("node:child_process", () => ({ execFileSync }));
+
+import { BACKEND_IMAGE, checkBackendImage, inspectBackendContainer } from "../backend-status.js";
+
+describe("inspectBackendContainer", () => {
+  beforeEach(() => {
+    execFileSync.mockReset();
+  });
+
+  it("checks the memory-layer image behind a compose port proxy", () => {
+    execFileSync.mockImplementation((_command: string, args: string[]) => {
+      if (args[0] === "info") return "ok";
+      if (args[0] === "ps" && args.includes("publish=8090")) return "proxy-id";
+      if (args[0] === "inspect" && args[1] === "proxy-id") {
+        return "sha256:proxy|::|nginx:stable|::|ix|::|/tmp/docker-compose.yml";
+      }
+      if (args[0] === "image" && args[2] === "sha256:proxy") {
+        return '["nginx@sha256:proxy"]';
+      }
+      if (args[0] === "ps" && args.includes("label=com.docker.compose.service=memory-layer")) {
+        return "backend-id";
+      }
+      if (args[0] === "inspect" && args[1] === "backend-id") {
+        return `sha256:backend|::|${BACKEND_IMAGE}:latest|::|ix|::|/tmp/docker-compose.yml`;
+      }
+      if (args[0] === "image" && args[2] === "sha256:backend") {
+        return `["${BACKEND_IMAGE}@sha256:backend"]`;
+      }
+      if (args[0] === "image" && args[2] === `${BACKEND_IMAGE}:latest`) {
+        return "sha256:backend";
+      }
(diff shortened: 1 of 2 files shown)


KageBinary:
Reviewed. Correct, and the fallback chain is the part that makes it safe: direct released backend → compose-project sibling → `publishers[0]`. That last line is what preserves local-build detection for a dev image that doesn't use the released repository name, so the existing `local-build` warning path is untouched.

Confirmed `memory-layer` is the right service label — it's the service name in `docker-compose.standalone.yml` and in the installed `~/.ix/backend/docker-compose.yml`.

**Mutation-checked**: forcing the compose-project follow to `continue` turns the proxy test red, so the new branch is genuinely pinned. 9/9 tests pass across this file and `backend-status.test.ts`.

Minor, non-blocking: this can now issue up to N `docker inspect` calls where it previously issued one. Bounded by the number of containers publishing 8090, so in practice 1–2, and each is already timeout-guarded by the `docker()` helper. Not worth changing.

No changes requested.

KageBinary:
Second, heavier pass — I re-verified this against a **real `docker` process** rather than the vitest mock, by putting a scripted `docker` binary on PATH and calling the compiled `inspectBackendContainer`/`checkBackendImage`. That exercises the real `execFileSync` call and the real Go-template parsing, which the mock does not.

**Against the actual daemon on this box** (memory-layer published directly, no proxy) — unchanged behaviour, first branch:
```
containerId ghcr.io/ix-infrastructure/ix-memory-layer:latest  composeProject=backend  checkBackendImage=ok
```

**Proxy scenario** (nginx on 8090, memory-layer internal to project `ix`): resolves to the memory-layer container, not nginx. And the result that matters most — an **old digest behind the proxy is now correctly reported as `digest-mismatch`** instead of being invisible. That is the actual user-facing win here and it works.

Fallback branches, all correct:

| scenario | result |
|---|---|
| dev backend published directly (`ix-memory-layer-dev:latest`) | `local-build` preserved |
| publisher has no compose-project label | skips the follow, falls back to the publisher |
| docker daemon unavailable | `null` / `docker-unavailable` |
| two publishers, released backend is the **second** | picks the released one |

That last one is worth calling out: it works because the code uses `.find(isReleasedBackendRef)` rather than `publishers[0]`, so it does not depend on `docker ps` ordering.

**One pre-existing wart this does not fix** (not a blocker, not a regression): a proxy on 8090 with *no* compose-project label and a non-released image is still reported as `local-build`, so `ix doctor` says "local build" when it actually just failed to find the backend. `main` does the same thing. Only worth a follow-up if bare-`docker run` proxies turn out to be a real deployment shape.

Still no changes requested.

PR #484, opened 2026-08-22
fix(ingest): split oversized bulk commits

### LINKED ISSUE #483: [bug] oversized bulk commits fall back to serial per-file ingestion
## Description

Bulk patch requests are grouped only by file count. A proxy or backend request-size limit can therefore reject a batch with HTTP 413. Ix then sends the entire failed batch one patch at a time behind a shared serial fallback lock, turning a recoverable size mismatch into an extremely slow map.

The local reproduction used a proxy with a 1 MiB request limit, but the problem applies to any ingress or backend payload cap.

## Steps to reproduce
1. Put the Ix backend behind an HTTP request-size limit.
2. Map a repository whose serialized patch batch exceeds that limit.
3. Observe the `/v1/patches/bulk` response and fallback timing.

## Expected behavior

Ix should reduce an oversized bulk request and retain bulk throughput for the smaller requests. Unrelated bulk failures should keep the existing per-file fallback.

## Actual behavior

One 413 causes every patch in the rejected chunk to be committed serially through `/v1/patch`.

## CLI output

```text
[bulk failed, falling back to per-file] ... 413 Request Entity Too Large
```

## Environment
- OS: macOS
- Node: v26.7.0
- Ix version: v0.10.0
- Backend: 1.0.17
- Repository size: 7,239 discovered files

Hiro-Chiba:
Closes #483.

## Summary
Split oversized bulk patch requests instead of serializing the entire rejected batch.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Recognize HTTP 413 and equivalent payload-limit errors.
- Bisect oversized batches recursively until the smaller bulk requests succeed.
- Keep the existing per-file path for a rejected single patch and for non-payload bulk failures.
- Add focused regression coverage for splitting and fallback behavior.

## Validation
- Ingest retry and bulk split tests
- `npm test`
- `npm run typecheck`
- ESLint on changed files
- `git diff --check`

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/ingest-retry.test.ts
+++ b/ix-cli/src/cli/__tests__/ingest-retry.test.ts
@@ -1,5 +1,10 @@
-import { describe, expect, it } from 'vitest';
-import { isAbortError, isRetryableCommitConflict } from '../commands/ingest.js';
+import { describe, expect, it, vi } from 'vitest';
+import {
+  commitBulkWithPayloadSplit,
+  isAbortError,
+  isPayloadTooLargeError,
+  isRetryableCommitConflict,
+} from '../commands/ingest.js';
 
 describe('isAbortError', () => {
   it('detects AbortError and TimeoutError by name', () => {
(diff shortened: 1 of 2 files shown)


KageBinary:
Reviewed. The recursive bisect is the right shape, and two details make it safe rather than merely plausible:

- **The recursion is sequential** (`await left; await right`), so splitting a chunk never puts two bulk commits in flight against `revisions.current` at once. Getting this wrong would have reintroduced the write-write conflict the per-file fallback exists to avoid.
- **`retryOnConflict` doesn't swallow the 413** — it only retries commit conflicts, so the payload error reaches the splitter on the first attempt.

I checked the error shape against the real client rather than the fixture: `commitPatchBulk` throws `new Error(`${resp.status}: ${text}`)`, so a real 413 arrives as `Error: 413: …` and `/^(?:error:\s*)?413(?:\b|:)/` matches it after `String(err).toLowerCase()`. The `patch 413 failed validation` negative test is the one that keeps that regex honest.

`commitMsPerChunk[ci] += chunkMs` (was `=`) is correct — the array is `new Array(n).fill(0)`, so there's no `undefined + n` NaN, and accumulating means a split chunk reports its true total including the rejected attempt.

**Mutation-checked twice**: forcing the split branch off, and dropping the `status === 413` check, each turn the relevant tests red. 54/54 ingest tests pass.

No changes requested.

PR #486, opened 2026-08-22
fix(status): flag workspaces without a completed map

### LINKED ISSUE #485: [bug] ix status reports partial first maps as up to date
## Description

The workspace ingest baseline is written only after a clean map. If the first map is interrupted after some bulk commits, the partial graph remains queryable but `ix status` treats the missing baseline as a clean revision-zero workspace.

## Steps to reproduce
1. Start the first `ix map` for a repository with no ingest baseline.
2. Interrupt it after some patch batches have committed.
3. Confirm that search or stats can see the partial graph.
4. Run `ix status` and `ix status --format llm`.

## Expected behavior

Without a completed map baseline, status should say that no trustworthy map has completed and recommend remapping. Machine-readable output should not claim `stale=false`.

## Actual behavior

Text output says `Graph is up to date.` and LLM output says `stale=false`, despite the graph being partial.

## CLI output

```text
currentRev: 0
lastIngestAt: null
staleFiles: 0
Graph is up to date.
```

## Environment
- OS: macOS
- Node: v26.7.0
- Ix version: v0.10.0
- Backend: 1.0.17

Hiro-Chiba:
Closes #485.

## Summary
Stop reporting a workspace as current when no map has completed successfully.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Carry explicit map-completion state from the workspace ingest baseline.
- Expose `mapCompleted` in JSON and `map_complete` in LLM output.
- Mark a workspace with no completed map as stale and show actionable text guidance.
- Add regression coverage for unmapped and completed workspaces.

## Validation
- Workspace staleness and LLM status tests
- `npm test`
- `npm run typecheck`
- ESLint on changed files
- `git diff --check`

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/llm-tier5.test.ts
+++ b/ix-cli/src/cli/__tests__/llm-tier5.test.ts
@@ -137,6 +137,7 @@ describe("read --format llm", () => {
 describe("status --format llm", () => {
   it("answers 'can I trust the graph' in one field", () => {
     const lines = renderStatusLlm("ok", "http://localhost:8090", {
+      mapCompleted: true,
       currentRev: 11, lastIngestAt: "2026-08-09T12:00:00.000Z",
       staleFiles: 2, sampleChangedFiles: ["src/a.ts", "src/b.ts"],
     });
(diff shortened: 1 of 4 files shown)


KageBinary:
Reviewed. The bug is real and I confirmed it on `main`: with no baseline, `detectStaleFiles` returns `staleFiles: 0`, and `status.ts` only branches on `staleFiles > 0`, so a workspace whose first `ix map` was interrupted falls through to `Graph is up to date.` — exactly what #485 describes.

Deriving `mapCompleted` from `loadIngestBaseline(root) !== null` is the right signal: the baseline is written at the *end* of a clean ingest pass, so its absence is precisely "no map has completed".

Two things I checked because they're the usual way a change like this goes wrong:

- **`--format llm` is key-addressed, not positional** (`llmLine` emits `key=value` pairs and drops nullish values), so inserting `map_complete` ahead of `rev` can't break an existing parser. `docs/llm-format.md` doesn't enumerate the status fields, so there's no doc to update.
- **`stale` now ORs in `!mapCompleted`**, which is the field an agent actually keys off. Getting that right matters more than the new field itself.

31/31 tests pass across `stale-workspace` and `llm-tier5`; typecheck clean. No changes requested.

PR #488, opened 2026-08-22
fix(context): consume summarized graph results

### LINKED ISSUE #487: [bug] ix context drops summarized graph relationships
## Description

The backend's `compact` and `standard` context modes return graph data as `nodeSummaries` and `edgeSummaries`, while the v0.10.0 CLI only reads `nodes` and `edges`. The default context bundle therefore collapses to its target and loses known relationships. The CLI also accepts numeric `--depth` values even though the backend depth vocabulary is `compact`, `standard`, and `full` with legacy aliases.

## Steps to reproduce
1. Map code containing a function with known callees.
2. Confirm the relationships with `ix explain <target>`.
3. Run `ix context <target> --format json`.
4. Also run `ix context <target> --depth 2 --format json`.

## Expected behavior

The context bundle should include the summarized entities and relationships returned by the backend. Unsupported depth values should fail at argument parsing.

## Actual behavior

The bundle contains only the resolved target and an empty `relationships` array. `--depth 2` is accepted but silently falls back backend-side.

## CLI output

```json
{
  "entities": [{ "name": "loadAccount" }],
  "relationships": []
}
```

`ix explain loadAccount` reported three callees for the same graph.

## Environment
- OS: macOS
- Node: v26.7.0
- Ix version: v0.10.0
- Backend: 1.0.17

Hiro-Chiba:
Closes #487.

## Summary
Consume summarized context graph results and validate the backend depth vocabulary.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Add the backend `nodeSummaries` and `edgeSummaries` response shapes to the client types.
- Use summaries when full `nodes` or `edges` arrays are empty.
- Validate `--depth` against compact, standard, full, shallow, and deep.
- Add bundle and CLI argument regression tests.

## Validation
- Context bundle and option validation tests
- `npm test`
- `npm run typecheck`
- ESLint on changed files
- `git diff --check`

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/context-pick-validation.test.ts
+++ b/ix-cli/src/cli/__tests__/context-pick-validation.test.ts
@@ -148,6 +148,27 @@ describe("ix context --as-of-rev validation", () => {
   });
 });
 
+describe("ix context --depth validation", () => {
+  beforeEach(() => {
+    resolveFileOrEntity.mockReset().mockResolvedValue(undefined);
+  });
+
+  it.each(["2", "wide", "standard-ish"])("rejects unsupported depth %j before resolving", async (depth) => {
+    const result = await runContext(["--depth", depth]);
+
+    expect(result.error?.exitCode).toBe(1);
+    expect(result.stderr).toContain("must be one of: compact, standard, full, shallow, deep");
+    expect(resolveFileOrEntity).not.toHaveBeenCalled();
+  });
+
+  it.each(["compact", "standard", "full", "shallow", "deep", "FULL"])("accepts depth %j", async (depth) => {
+    const result = await runContext(["--depth", depth]);
+
+    expect(result.error).toBeUndefined();
+    expect(resolveFileOrEntity).toHaveBeenCalledOnce();
+  });
+});
+
 describe("ix context refusals reach the caller that asked for records", () => {
   beforeEach(() => {
     resolveFileOrEntity.mockReset().mockResolvedValue(undefined);
--- a/ix-cli/src/cli/__tests__/context.test.ts
+++ b/ix-cli/src/cli/__tests__/context.test.ts
@@ -3,9 +3,11 @@ import { describe, expect, it } from "vitest";
 import type {
   ConflictReport,
   DecisionReport,
+  EdgeSummary,
   GraphEdge,
   GraphNode,
   IntentReport,
+  NodeSummary,
   ScoredClaim,
 } from "../../client/types.js";
 import type { EntityFacts } from "../explain/facts.js";
@@ -58,6 +60,8 @@ function makeContext(overrides: Partial<{
   intents: IntentReport[];
   nodes: GraphNode[];
   edges: GraphEdge[];
+  nodeSummaries: NodeSummary[];
+  edgeSummaries: EdgeSummary[];
 }> = {}) {
   return {
     claims: overrides.claims ?? [makeClaim("renders to DOM", 0.9)],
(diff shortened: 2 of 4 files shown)


KageBinary:
Reviewed and **verified against a live 1.0.17 backend**, not just the mocks. This is the highest-value PR of the six and the bug is worse than the description suggests: it hits the **default** `ix context` path.

The CLI sends no `depth` unless `--depth` is passed, and `ContextService` defaults to `standard`:

```scala
effectiveDepthRaw = depth.getOrElse("standard").toLowerCase
...
nodesOut = if (effectiveDepth == "full") allNodes else Vector.empty
edgesOut = if (effectiveDepth == "full") edgesTrimmed else Vector.empty
```

So every default `ix context` was reading two arrays the backend deliberately empties. `POST /v1/context` here:

| depth | nodes | edges | nodeSummaries | edgeSummaries |
|---|---|---|---|---|
| compact | 0 | 0 | 8 | 10 |
| **standard (default)** | **0** | **0** | **40** | **106** |
| full | 84 | 106 | 84 | 106 |

End-to-end through the built CLI, same target, same backend:

| | entities | relationships |
|---|---|---|
| `main` | 1 | 0 |
| this PR | **41** | **100** |

One entity and no relationships means the bundle was target-only — the feature was returning almost nothing by default.

**The two claims I checked rather than took on trust:**

- **Field shapes.** `NodeSummary(id, kind, name, rev, sourceUri: Option[String])` and `EdgeSummary(id, src, dst, predicate, rev)` in `StructuredContext.scala` match your TS types exactly, including `sourceUri` arriving as literal `null` — which `?? undefined` handles.
- **The depth vocabulary.** `compact|standard|full|shallow|deep` is exactly what the backend normalizes (`case "shallow" => "compact"; case "deep" => "full"`), so the new validator rejects nothing the backend would have accepted. Worth saying because a too-narrow allowlist here would have been a regression that no unit test could see.

The `context.nodes.length > 0 ? nodes : summaries` shape is right rather than merely convenient: at `full` both are populated and `nodes` wins, which keeps the richer `provenance.sourceUri`.

129/129 context tests pass; typecheck clean. No changes requested.

KageBinary:
The graph-summaries fix itself is verified and I want it — see the numbers below. **#493** carries your commit `13ae4f4` unchanged plus one fix commit; I couldn't push to your fork from this environment, so it had to be a separate PR. The credit is yours.

**The one thing that needed changing: the `--depth` allowlist is a breaking change.**

`ContextService` picks its limits with a match ending in `case _ => standard`, so a value outside the vocabulary was never an error. `ix context X --depth 2` ran a standard-depth query and exited 0. This PR turns that into:

```
error: option '--depth <depth>' argument '2' is invalid. must be one of: compact, standard, full, shallow, deep
```

Anyone with a numeric or stale depth in a script starts failing on upgrade — inside a **patch** release, for a flag whose wrong values had been harmless. #493 warns and falls back to `standard` instead, on stderr so `--format json`/`--format llm` stay parseable:

```
$ ix context README.md --depth 2 --format json
  Warning  --depth 2 is not one of compact, standard, full, shallow, deep; using standard.
$ echo $?  →  0
```

The validation is still real and still tested — `parseContextDepthOption` is now exported and tested directly, since the command-level harness returns before a depth would reach the client. Deliberately unlike `--pick`/`--as-of-rev` beside it, which do throw: those reject values with no defined meaning, whereas this one's meaning is defined and it is `standard`.

**On the main fix — I verified it against a live 1.0.17 backend rather than the mocks, and it's bigger than the description says.** The CLI sends no `depth` by default and the backend defaults to `standard`, which is one of the depths that empties the full arrays. So the *default* path was reading two deliberately-empty arrays:

| depth | nodes | edges | nodeSummaries | edgeSummaries |
|---|---|---|---|---|
| **standard (default)** | **0** | **0** | **40** | **106** |
| full | 84 | 106 | 84 | 106 |

End-to-end, same target: `main` → **1 entity / 0 relationships**; this → **41 / 100**.

Also confirmed the depth vocabulary itself is exactly right (`case "shallow" => "compact"; case "deep" => "full"`), so the allowlist rejects nothing the backend would have accepted — worth saying, because a too-narrow list there would have been a second regression no unit test could see.

KageBinary:
Superseded by #493, which consumes the summarized graph results *and* warns rather than throwing on an unrecognized depth.

#493 is a strict superset of this change. Closing in favour of it.

PR #490, opened 2026-08-22
fix(mcp): flag semantic CLI errors

### LINKED ISSUE #489: [bug] MCP tools return semantic error records without isError
## Description

Formatted CLI commands can return a structured `error code=...` record while exiting successfully. The MCP `runFormatted` wrapper currently treats only a nonzero runner result as failure, so `ix_locate` returns `unknown_target` without MCP `isError: true`.

## Steps to reproduce
1. Start `ix mcp` against a mapped graph.
2. Call `ix_locate` with a target that does not exist.
3. Inspect the MCP tool result.

## Expected behavior

A leading structured CLI error record should be returned as an MCP tool error with `isError: true`, while preserving the text payload. Error-like lines contained inside successful source output should remain data.

## Actual behavior

The text says `error code=unknown_target`, but the MCP result has no `isError` field and appears successful to clients.

## CLI output

```text
error code=unknown_target message="No graph entity found for \"Missing\"."
```

## Environment
- OS: macOS
- Node: v26.7.0
- Ix version: v0.10.0
- MCP SDK: 1.30.x

Hiro-Chiba:
Closes #489.

## Summary
Mark leading structured CLI error records as MCP tool errors.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Inspect successful `runFormatted` output for a leading `error code=` record.
- Add `isError: true` without changing the CLI text payload.
- Keep error-like source lines after a successful leading record as ordinary data.
- Add MCP regression coverage for both cases.

## Validation
- MCP tool tests
- `npm test`
- `npm run typecheck`
- ESLint on changed files
- `git diff --check`

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/mcp.test.ts
+++ b/ix-cli/src/cli/__tests__/mcp.test.ts
@@ -108,7 +108,41 @@ describe("ix mcp", () => {
     });
 
     expect(calls).toEqual([["locate", "--format=llm", "--", "Widget"]]);
+    expect(result.isError).toBeFalsy();
     expect(result.content).toEqual([{ type: "text", text: "match name=Widget" }]);
+  });
+
+  it("marks a structured LLM error as an MCP error when the CLI exits successfully", async () => {
+    const error = 'error code=unknown_target message="No graph entity found for \\"Missing\\"."';
+    const client = await connect(async () => ({
+      ok: true,
+      stdout: error,
+      stderr: 'No entity found matching "Missing".',
+    }));
+
+    const result = await client.callTool({
+      name: "ix_locate",
+      arguments: { symbol: "Missing" },
+    });
+
+    expect(result.isError).toBe(true);
+    expect(result.content).toEqual([{ type: "text", text: error }]);
+  });
+
+  it("keeps error-like source lines after a successful LLM record as data", async () => {
+    const output = [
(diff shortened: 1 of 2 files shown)


KageBinary:
Reviewed, and I checked the premise against the real CLI rather than the fixture:

```
$ ix locate ZzNotARealSymbolQq --format llm
error code=unknown_target message="No graph entity found for \"ZzNotARealSymbolQq\"."
$ echo $?
0
```

Exit 0 with a structured error on stdout — so the MCP server was reporting these as successful tool calls, and your test fixture matches the real output byte for byte.

The "only the leading record governs status" rule is right, and it works because `runCommand` puts the whole stdout in a single text block, so `startsWith` is genuinely testing the first line. The `ix_read` case you cover is the one that would have broken a naive `includes`.

**Mutation-checked**: forcing `semanticError` to `false` turns exactly the new error test red. 22/22 mcp tests pass.

Non-blocking note: a file whose *first* line is literally `error code=` would be misreported by `ix_read`. `ix read --format llm` always emits a `content` record first, so it can't happen today — flagging it only so the invariant is written down somewhere. No changes requested.

PR #493, opened 2026-08-22
fix(context): consume summarized graph results, and warn rather than fail on unknown --depth

KageBinary:
Supersedes #488. Carries @Hiro-Chiba's commit `13ae4f4` unchanged, plus one fix
commit. Opened here only because pushing to the fork is not possible from this
environment — please credit @Hiro-Chiba, and close #488 in favour of this once
it lands.

## Why #488 matters (verified against a live 1.0.17 backend, not the mocks)

The CLI sends no `depth` unless `--depth` is passed, and `ContextService`
defaults to `standard` — which is one of the depths that deliberately empties
the full arrays:

```scala
nodesOut = if (effectiveDepth == "full") allNodes else Vector.empty
edgesOut = if (effectiveDepth == "full") edgesTrimmed else Vector.empty
```

So **every default `ix context` was reading two arrays the backend empties on
purpose.** `POST /v1/context` against this box:

| depth | nodes | edges | nodeSummaries | edgeSummaries |
|---|---|---|---|---|
| compact | 0 | 0 | 8 | 10 |
| **standard (default)** | **0** | **0** | **40** | **106** |
| full | 84 | 106 | 84 | 106 |

End-to-end through the built CLI, same target, same backend:

| | entities | relationships |
|---|---|---|
| `main` | 1 | 0 |
| this PR | **41** | **100** |

`NodeSummary(id, kind, name, rev, sourceUri: Option[String])` and
`EdgeSummary(id, src, dst, predicate, rev)` in `StructuredContext.scala` match
the added TS types exactly, including `sourceUri` arriving as literal `null`.

## What this PR adds: the `--depth` allowlist no longer exits 1

This is the only breaking change in an otherwise pure-bugfix batch, and it broke
a case that had always worked.

The backend picks its limits with a match ending in `case _ => standard`, so a
value outside the vocabulary was never an error — `ix context X --depth 2` ran a
standard-depth query and exited 0. #488 turned that into:

```
error: option '--depth <depth>' argument '2' is invalid. must be one of: compact, standard, full, shallow, deep
```

Anyone with a numeric or stale depth in a script would start failing on upgrade,
inside a **patch** release, for a flag whose wrong values had been harmless.

Now it warns and does what the backend already did:

```
$ ix context README.md --depth 2 --format json
  Warning  --depth 2 is not one of compact, standard, full, shallow, deep; using standard.
$ echo $?
0
```

stdout stays valid JSON with the full 41/100 standard-depth result. The warning
goes to **stderr** via `renderWarningErr`, so `--format json` and `--format llm`
stay parseable by whatever is reading stdout — the entire audience for those two
formats.

Deliberately unlike `--pick` and `--as-of-rev` beside it, which do throw
`InvalidArgumentError`: those reject values with no defined meaning, whereas this
one has a defined meaning and it is `standard`. The typo still gets surfaced;
nobody's pipeline starts exiting 1.

## Verification

- Full suite **1201 passed**, 2 skipped; `typecheck` clean; `lint` **0 errors**
  (42 pre-existing warnings, same as `main`); `knip` exit 0
- **Mutation-checked both directions**: restoring the `throw` fails 6 tests;
  falling back to `full` instead of `standard` fails 3
- The `--depth` validation is still real — `parseContextDepthOption` is exported
  and tested directly, since the command-level harness returns before a depth
  would reach the client

One caveat worth recording: across 6 full-suite runs, one run had a single
unidentified test failure that did not reproduce in the other 5, including a
deliberate rebuild-then-run to test the obvious hypothesis. Nothing in this PR
touches processes, ports, or timing — it is a pure function plus its tests — so
I do not believe it is from this change, but I would rather flag it than report
6/6 green.

### DIFF
--- a/ix-cli/src/cli/__tests__/context-pick-validation.test.ts
+++ b/ix-cli/src/cli/__tests__/context-pick-validation.test.ts
@@ -148,6 +148,27 @@ describe("ix context --as-of-rev validation", () => {
   });
 });
 
+describe("ix context --depth validation", () => {
+  beforeEach(() => {
+    resolveFileOrEntity.mockReset().mockResolvedValue(undefined);
+  });
+
+  it.each(["2", "wide", "standard-ish"])("rejects unsupported depth %j before resolving", async (depth) => {
+    const result = await runContext(["--depth", depth]);
+
+    expect(result.error?.exitCode).toBe(1);
+    expect(result.stderr).toContain("must be one of: compact, standard, full, shallow, deep");
+    expect(resolveFileOrEntity).not.toHaveBeenCalled();
+  });
+
+  it.each(["compact", "standard", "full", "shallow", "deep", "FULL"])("accepts depth %j", async (depth) => {
+    const result = await runContext(["--depth", depth]);
+
+    expect(result.error).toBeUndefined();
+    expect(resolveFileOrEntity).toHaveBeenCalledOnce();
+  });
+});
+
 describe("ix context refusals reach the caller that asked for records", () => {
   beforeEach(() => {
     resolveFileOrEntity.mockReset().mockResolvedValue(undefined);
--- a/ix-cli/src/cli/__tests__/context-pick-validation.test.ts
+++ b/ix-cli/src/cli/__tests__/context-pick-validation.test.ts
@@ -8,7 +8,7 @@ vi.mock("../resolve.js", async (importOriginal) => ({
   resolveFileOrEntity,
 }));
 
-import { registerContextCommand } from "../commands/context.js";
+import { parseContextDepthOption, registerContextCommand } from "../commands/context.js";
 
 async function runContext(args: string[]): Promise<{ error?: CommanderError; stderr: string }> {
   const stderr: string[] = [];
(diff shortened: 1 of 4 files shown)


KageBinary:
Heavier pass on the graph-summaries half. The fix holds, but it surfaces a **pre-existing backend behaviour** that is worth filing separately — not a blocker for this PR, and not something this PR causes.

**Every relationship in the bundle references an entity the bundle does not contain.** Measured on the produced bundle:

```
entities: 41   relationships: 100
relationships whose src or dst is NOT in entities: 100
```

I checked whether that was this PR's construction or the backend's, by querying `/v1/context` with the *same* query the CLI sends:

| depth | nodeSummaries | edgeSummaries | edges with BOTH ends in the node set |
|---|---|---|---|
| compact | 8 | 10 | 1 |
| **standard (default)** | **40** | **120** | **0** |
| full | 84 | 106 | **106** |

It is the backend. `ContextService` applies `nodeLimit` and `edgeLimit` to the two collections independently, so at `standard` it returns 120 edges whose endpoints are mostly outside the 40 nodes it sent. At `full`, where `nodesOut = allNodes`, the graph is referentially complete. This PR faithfully surfaces what it is given — the `nodes.length > 0 ? nodes : summaries` fallback is not at fault.

**Why it is worth a follow-up anyway:** those relationships reach the user as evidence lines made of two opaque UUIDs —

```
[30] relationship — 00579c85-38b5-1e2b-11cc-f1375286d031 --CONTAINS--> 6d1ed72f-64d5-0cfe-452b-299ccb1083d1
       graph relationship from the context service
```

— and they are **competing for the evidence budget**: the same bundle reports `relationshipsTruncated: 20, evidenceTruncated: 32`. So unusable UUID-pair evidence is displacing 32 items that might have been readable.

The fix belongs in `ContextService`: either trim edges to the returned node set, or include the endpoints of returned edges in `nodeSummaries`. Doing it CLI-side by filtering unresolvable relationships would be wrong — at `standard` it would drop all 100 and put us back where `main` is.

To be clear about the comparison, because it is what decides whether this PR is an improvement: `main` gives **1 entity / 0 relationships** on this query. This gives 41 entities — all of them named and resolvable — plus 100 relationships that are currently only machine-identifiable. Strictly better, with a known next step.

Happy to open the backend issue if that is useful.

KageBinary:
Followed up on the dangling-relationship finding: filed as **ix-infrastructure/Ix-memory#158** and fixed in **ix-infrastructure/Ix-memory#159**.

Root cause was two independent takes in `ContextService`:

```scala
allNodes     = allNodesAll.take(nodeLimit)
edgesTrimmed = expanded.edges.take(edgeLimit)
```

`allNodesAll` is `seeds ++ expanded`, so its head is seeds, while edges run from a seed *out* to an expansion node further down — the node take removed exactly the endpoints the edge take needed.

Worth recording because it was my first instinct and it is wrong: **keeping only the edges induced on the first N nodes yields zero edges** at standard depth on this query. The node set has to be grown *from* the edges instead. After the fix, standard goes from 120 edges / 0 resolvable to **27 edges / 27 resolvable**, and the invariant now holds at `full` too — which was never sound either, it just happened to have 151 nodes against a 200 limit.

**Nothing is needed in this PR.** It consumes what the backend sends and will simply start receiving coherent data once the backend ships. Two notes for sequencing:

- The backend fix reaches nobody until a `v*` tag builds the image — merging it alone ships nothing.
- Until then this PR is still the right change: 41 named entities beats `main`'s 1, and the relationships were already being returned, just not shown.

PR #507, opened 2026-08-24
fix(stale): mark unmapped graph results stale

### LINKED ISSUE #506: [bug] partial first-map graph results are reported as current
## Description

When an initial `ix map` commits source graph data but does not produce a successful map baseline, read commands can still return those partial graph results as current. `ix status` correctly reports `mapCompleted: false`, but `ix context` reports `classification: current` because the stale probe treats a missing baseline as fresh.

This can make agents trust a graph whose initial map never completed.

## Steps to reproduce
1. Create a fresh workspace with a source file containing a class.
2. Run ingestion so that the entity is committed.
3. Remove the local ingest baseline, reproducing the state left when an initial map fails after committing graph patches.
4. Run `ix status --format json`.
5. Run `ix context <ClassName> --kind class`.

## Expected behavior

Graph results should be marked stale or unverified whenever there is no completed map baseline.

## Actual behavior

`ix status` reports that the map is incomplete, while `ix context` classifies the returned partial graph as current.

## CLI output

```text
$ ix status --format json
{
  "mapCompleted": false,
  "currentRev": 0,
  "lastIngestAt": null
}

$ ix context RuntimeStaleWidget --kind class
Context: RuntimeStaleWidget
  classification:current
```

## Environment
- OS: macOS 26.6.2
- Node: v26.7.0
- Ix version: 0.10.2
- Backend version: 1.0.26

Hiro-Chiba:
Closes #506.

## Summary
Treat graph results as stale when their workspace has no completed ingest baseline.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [x] Test
- [ ] CI

## Changes
- Make the shared stale probe treat a missing baseline as unverified instead of current.
- Cover both direct stale checks and the cached probe used by `ix context`.

## Validation
- `npm test`: 82 files passed, 1387 tests passed, 3 skipped.
- `npm run typecheck` passed.
- Parser smoke test and ESLint passed.
- Runtime comparison on 0.10.2/backend 1.0.26 changed an incomplete workspace result from `classification:current` to `classification:stale`.

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/context.test.ts
+++ b/ix-cli/src/cli/__tests__/context.test.ts
@@ -225,6 +225,30 @@ describe("ix context bundle", () => {
     ]);
   });
 
+  it("classifies a workspace with no completed map baseline as unverified", () => {
+    // Ix#506: the backend answers from graph patches an initial map committed
+    // before it failed. `ix status` already reported mapCompleted:false while
+    // the bundle said current, so an agent reading the bundle trusted a graph
+    // the CLI knew was half-built.
+    const bundle = buildBundle({ ...input(), mapCompleted: false });
+
+    expect(bundle.freshness.classification).toBe("unverified");
+  });
+
+  it("does not let an unverified workspace masquerade as stale", () => {
+    // `stale` is a claim that a file changed since ingest. Without a baseline
+    // nothing is known to have changed, so the two states stay distinguishable:
+    // an agent can re-map on unverified without being told its files are dirty.
+    const unverified = buildBundle({ ...input(), mapCompleted: false });
+    const stale = buildBundle({
+      ...input(), mapCompleted: true, facts: makeFacts({ stale: true }),
+    });
+
+    expect(unverified.freshness.classification).toBe("unverified");
+    expect(unverified.freshness.stale).toBe(false);
+    expect(stale.freshness.classification).toBe("stale");
+  });
+
   it("asks about each entity's own staleness instead of copying the target's", () => {
     // `freshness` is the target's, from the facts collector. Every other entity
     // has its own source file and its own answer — stamping the target's onto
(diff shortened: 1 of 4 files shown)


KageBinary:
Reviewed and pushed a follow-up commit that changes the approach. Summary of why, since it moves away from the one-line fix.

**The original change was red on all five test jobs.** `context.test.ts` and `context-investigation.test.ts` both failed: with no baseline in the CI environment, every entity carrying a path flipped to `stale: true`. The PR body reports a green full suite locally — that gap is almost certainly #498, since `stale.ts` resolves against the real `~/.ix`, so a local run finds a baseline where CI finds none. Worth setting `IX_HOME` when running this suite locally.

**Underneath the test failures there was a semantics problem.** A missing baseline has three causes, and only one of them is the bug this PR is about:

1. an initial map that committed graph patches and then failed — the #506 case
2. a workspace ingested through the cloud runner, which never writes a local baseline (`saveIngestBaseline` is only reachable from `commands/ingest.ts`; the `runner.runIngestion` path in `map.ts` does not touch it)
3. any workspace whose ingest has never completed without a parse or commit error, since `persistIngestBaselineIfClean` requires `parseErrors === 0 && commitErrors === 0`

In (2) and (3) the graph is current, so answering `true` reported every result of `ix read`, `ix explain`, `ix locate` and `ix context` as modified. A repo with one unparseable file would never see an accurate staleness flag again.

**The fix uses the state the type already declares.** `freshness.classification` has been `"current" | "stale" | "unverified"` since it was written, and nothing has ever produced `unverified` — a missing baseline is exactly what it describes. So:

- `createStaleProbe` goes back to `false` for a missing baseline. Staleness is a claim that a file changed; without a baseline that question has no answer, which is not the same as the answer being yes.
- `ix context` classifies the bundle `unverified` when the workspace has no completed baseline, via a new `hasCompletedMapBaseline()` in `stale.ts`, injected into `buildBundle` the same way `isStale` already is so the function stays pure under test.

An agent still stops trusting a half-built graph — which is what #506 asked for — without being told its files are dirty when they are not. The two states stay distinguishable, and there is a test pinning that.

`npm test`: 82 files, 1391 passed, 2 skipped. Typecheck, lint (0 errors) and the parser smoke test pass.

KageBinary (approved):
Reworked to use the freshness union's existing `unverified` state rather than marking every file stale, which avoids false positives on cloud-ingested workspaces and any workspace whose ingest never completed cleanly. Verified the merge with main (post-#509) locally: 82 files, 1403 tests green.

PR #509, opened 2026-08-24
fix(map): reject empty hierarchy after ingest

### LINKED ISSUE #508: [bug] ix map exits successfully with an empty hierarchy after ingesting PHP
## Description

`ix map` can ingest supported source files, receive a completed response containing zero files and zero regions, and still exit successfully. On a PHP module with 260 files, the local parser committed all 260 patches, while backend 1.0.26 returned `full_local_completed` with an empty architecture hierarchy.

The CLI then persisted a successful baseline, so `ix status` reported `mapCompleted: true` even though no hierarchy existed.

## Steps to reproduce
1. Start Ix 0.10.2 with backend 1.0.26.
2. Change to a PHP-only module containing supported `.php` source files.
3. Run `ix map --full --silent`.
4. Run `ix status --format json`.

## Expected behavior

Ix should either build a hierarchy for the ingested language or fail clearly and leave the workspace unmapped.

## Actual behavior

The command exits 0 with an empty map and records the workspace as successfully mapped.

## CLI output

```text
$ ix map --full --silent
map: 0 files · 0s/0ss/0m regions · 462ms

$ ix status --format json
{
  "backend": "ok",
  "mapCompleted": true,
  "currentRev": 683,
  "staleFiles": 0
}
```

The local ingest in this run found 260 supported source files and committed 260 patches.

## Environment
- OS: macOS 26.6.2
- Node: v26.7.0
- Ix version: 0.10.2
- Backend version: 1.0.26

Hiro-Chiba:
Closes #508.

## Summary
Reject completed-but-empty architecture maps after a clean local ingest found supported source files.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [x] Test
- [ ] CI

## Changes
- Return the minimal discovery and commit summary from local ingestion to `ix map`.
- Detect completed backend responses that contain no files or regions after a clean, non-empty ingest.
- Exit 1 with a language-support diagnostic and invalidate the completion baseline so `ix status` remains unmapped.
- Preserve successful behavior for genuinely empty workspaces, ingest failures, non-completed outcomes, and non-empty maps.

## Validation
- Targeted map and staleness suites: 32 tests passed.
- Full suite: 1392 tests passed and 3 skipped. One `watch-dedup` test timed out while two full suites were running concurrently; its isolated retry passed in 5.21 seconds.
- `npm run typecheck`, ESLint, and `git diff --check` passed.
- Runtime comparison on 0.10.2/backend 1.0.26 changed a 260-file PHP map from exit 0 plus `mapCompleted: true` to exit 1 plus `mapCompleted: false`.

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/map-auto-skip.test.ts
+++ b/ix-cli/src/cli/__tests__/map-auto-skip.test.ts
@@ -115,6 +115,7 @@ describe('describeEmptyCompletedMap', () => {
     expect(message).toContain('mapped 0 files after local ingest found 260 supported source files');
     expect(message).toContain('(260 patches committed)');
     expect(message).toContain('no architecture hierarchy was created');
+    expect(message).toContain("the next 'ix map' re-parses every file");
   });
 
   it('does not reject an actually empty workspace', () => {
@@ -139,6 +140,28 @@ describe('describeEmptyCompletedMap', () => {
       parseErrors: 0,
       commitErrors: 1,
     })).toBeUndefined();
+  });
+
+  it('ignores an outcome the backend does not actually send', () => {
+    // 'ok' was in the completed set but is not one of the six MapOutcome
+    // labels, so it only ever looked like coverage.
+    expect(describeEmptyCompletedMap({ ...emptyResult, outcome: 'ok' }, {
+      filesDiscovered: 260,
+      patchesApplied: 260,
+      parseErrors: 0,
+      commitErrors: 0,
+    })).toBeUndefined();
+  });
+
+  it('leaves guardrail refusals alone, which carry no regions by design', () => {
+    for (const outcome of ['local_map_too_large', 'local_map_not_recommended']) {
+      expect(describeEmptyCompletedMap({ ...emptyResult, outcome }, {
+        filesDiscovered: 260,
+        patchesApplied: 260,
+        parseErrors: 0,
+        commitErrors: 0,
+      })).toBeUndefined();
+    }
   });
 
   it('requires both an explicitly completed outcome and an entirely empty response', () => {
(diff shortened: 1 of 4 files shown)


KageBinary:
Reviewed — the strongest of the three. I checked its assumptions against the backend rather than the description, and they hold:

- **The outcome set matches reality.** `MapOutcome` (MapTypes.scala) defines exactly six labels; the set covered four and correctly left out the two guardrail refusals, which carry no regions by design.
- **`coupling_unchanged` belongs in it**, which was the entry I most expected to be a regression. It isn't: the backend answers it with `map.copy(outcome = CouplingUnchanged)` — the cached map from the last build's revision, not an empty delta — so an empty one really is an empty hierarchy being served from cache.
- **`file_count === 0` is what keeps the guard narrow.** The backend sets `fileCount = rawGraph.vertices.size`, independent of whether any regions were produced, so a small repo that legitimately clusters into zero regions still has a non-zero `file_count` and won't trip this. That distinction is what makes the check safe to run on every map.
- **`filesDiscovered = filePaths.length`** is assigned before hash-skipping, so the unchanged-retry case behaves as the test claims.

Pushed a small follow-up:

- Dropped `"ok"` from the set — it is not one of the six labels the backend sends, so it only ever looked like coverage. Added a test saying so, plus one pinning the two guardrail refusals as deliberately outside the set.
- Recorded why `coupling_unchanged` is inside it, since it reads like a skip and is the entry most likely to be removed later as a mistake.
- The failure message now mentions that the ingest cache was cleared, so the next `ix map` re-parsing every file is visible to the person who just hit the error rather than only in the code comment.

`npm test`: 82 files, 1397 passed, 2 skipped. Typecheck, lint (0 errors) and the parser smoke test pass.

KageBinary (approved):
Verified the guard's assumptions against the backend: MapOutcome defines exactly six labels, coupling_unchanged returns the cached full map rather than an empty delta, and fileCount = rawGraph.vertices.size is independent of region production — which is what keeps file_count === 0 a narrow, safe trigger. Full matrix green.

PR #511, opened 2026-08-24
fix(doctor): scope graph counts to active workspace

### LINKED ISSUE #510: [bug] ix doctor reports graph counts from outside the active workspace
## Description

`ix doctor` calls the unscoped stats endpoint, so its graph counts include tombstoned nodes and data from outside the active workspace. This disagrees with `ix stats`, which scopes counts to the active workspace or system.

As a result, doctor can report a large healthy graph immediately after the current workspace is reset or when the workspace contains only a few entities.

## Steps to reproduce
1. Use a backend that contains tombstoned data from previous mappings.
2. Create a small isolated workspace and ingest one TypeScript class.
3. Run `ix stats --format json` from that workspace.
4. Run `ix doctor --format json` from the same workspace.

## Expected behavior

Doctor should diagnose the active workspace or system and report the same live graph counts as `ix stats`.

## Actual behavior

`ix stats` reports 5 nodes and 6 edges for the workspace, while `ix doctor` reports 17,213 nodes and 2,528 edges from the unscoped graph.

## CLI output

```text
$ ix stats --format json
nodes.total: 5
edges.total: 6

$ ix doctor --format json
Graph has nodes: 17213 nodes
Graph has edges: 2528 edges
```

## Environment
- OS: macOS 26.6.2
- Node: v26.7.0
- Ix version: 0.10.2
- Backend version: 1.0.26

Hiro-Chiba:
Closes #510.

## Summary
Scope `ix doctor` graph counts to the active workspace or stitched system.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [x] Test
- [ ] CI

## Changes
- Resolve the active read system before fetching doctor stats.
- Fall back to the active workspace when no stitched system is present.
- Keep the existing single stats request shared by the node and edge checks.

## Validation
- `npm test`: 82 files passed, 1389 tests passed, 3 skipped.
- `npm run typecheck`, parser smoke test, ESLint, and `git diff --check` passed.
- Runtime comparison on 0.10.2/backend 1.0.26 changed doctor from unscoped 17,213 nodes/2,528 edges to the active workspace's 5 nodes/6 edges.

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/doctor-stats-once.test.ts
+++ b/ix-cli/src/cli/__tests__/doctor-stats-once.test.ts
@@ -1,17 +1,32 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { Command } from "commander";
 
-const statsCalls = { n: 0 };
+const statsCalls: Array<{ workspaceId?: string; systemId?: string } | undefined> = [];
+const scope = { workspaceId: "workspace-current", systemId: undefined as string | undefined };
 
 vi.mock("../../client/api.js", () => ({
   IxClient: class {
-    async stats() {
-      statsCalls.n += 1;
-      return { nodes: { total: 10 }, edges: { total: 20 } };
+    async stats(opts?: { workspaceId?: string; systemId?: string }) {
+      statsCalls.push(opts);
+      if (opts?.workspaceId === scope.workspaceId ||
+          (scope.systemId !== undefined && opts?.systemId === scope.systemId)) {
+        return { nodes: { total: 3457 }, edges: { total: 10365 } };
+      }
+      return { nodes: { total: 22969 }, edges: { total: 10365 } };
     }
     async conflicts() { return []; }
     async health() { return { status: "ok" }; }
   },
+}));
+
+vi.mock("../bootstrap.js", async (orig) => ({
+  ...(await orig<typeof import("../bootstrap.js")>()),
+  resolveWorkspaceId: () => scope.workspaceId,
+}));
+
+vi.mock("../resolve.js", async (orig) => ({
+  ...(await orig<typeof import("../resolve.js")>()),
+  resolveReadSystemId: async () => scope.systemId,
 }));
 
 // Doctor also inspects the live container and probes the schema. Both are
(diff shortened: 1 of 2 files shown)


KageBinary:
Reviewed; the scoping itself is right and the call is byte-identical to the one in `ix stats`, so doctor and stats can no longer disagree. Pushed a follow-up with two corrections, neither to what it computes.

**The comment's rationale was wrong.** It explained the change as excluding tombstoned nodes. `ArangoStatsStore` applies `FILTER n.deleted_rev == null` in all four of its queries — scoped and unscoped alike — so deleted nodes were never in either count. What the unscoped call actually added was every *live* node belonging to some other workspace on the same backend, which is the whole of the 17,213-vs-5 gap in #510. Left as-is it would have sent the next reader looking for a tombstone bug that isn't there. (#510's title has the same diagnosis, so it's worth editing too.)

**Scoping the count makes it less self-explanatory.** `0 nodes` against a backend holding thousands reads as a broken install until the check says whose nodes it counted, and #510 was opened over a number that surprised its reader in the first place. The detail now carries the scope — `5 nodes in this workspace`, `… in this system`, or `… in all workspaces` for the no-workspace-registered case, where the request really is unscoped and naming a workspace would be the lie.

One thing worth noting for the record, since it makes the change safer than it first looks: `resolveWorkspaceId` falls back to `getDefaultWorkspace()`, so running `ix doctor` from an unregistered directory scopes to the default workspace rather than suddenly reporting zero.

Also tightened the test mock, which compared `opts.workspaceId === scope.workspaceId` — with both sides `undefined` the unscoped call matched the scoped branch, so the third case could not have been tested through it.

`npm test`: 82 files, 1391 passed, 2 skipped. Typecheck, lint (0 errors) and the parser smoke test pass.

KageBinary (approved):
Scoping matches `ix stats` exactly, so the two commands can no longer disagree. Verified the rationale against the backend: /v1/stats already filters deleted_rev == null in every query, so this is a workspace-scoping fix rather than a tombstone one — comment and issue title corrected accordingly. Full matrix green.

PR #529, opened 2026-08-28
fix: skip generated dependency lockfiles

### LINKED ISSUE #523: [bug] Generated dependency lockfiles can make ix map fail with HTTP 413
## Description

Generated dependency lockfiles are parsed as architecture inputs. A lockfile smaller than the CLI's 1 MB source limit can expand into a graph patch larger than nginx's default request-body limit. Bulk bisection eventually reaches one patch, then both the bulk and per-file requests fail with HTTP 413. The remaining source files are mapped, but the command exits non-zero and no completion baseline is written.

Dependency lockfiles are generated snapshots and do not provide useful architectural structure. They should be excluded during discovery.

## Steps to reproduce
1. Create a minimal TypeScript package with one source file.
2. Add `react-scripts@5.0.1` and generate `package-lock.json` with `npm install --package-lock-only --ignore-scripts`.
3. Confirm that the generated lockfile is about 646 KB.
4. Run `ix map --full --silent --verbose`.

## Expected behavior

The generated lockfile is skipped and the source graph maps successfully.

## Actual behavior

The single lockfile patch is rejected with HTTP 413 after bulk bisection reaches the per-file fallback. `ix map` exits 1 and the workspace has no completed baseline.

## CLI output

```text
[bulk failed, falling back to per-file] files 1-1 (1 patches): Error: 413 Request Entity Too Large
[commit error] package-lock.json: Error: 413 Request Entity Too Large
Warning: 1 of 3 file patches failed to commit
map: 2 files
```

## Environment
- OS: macOS
- Node: 22.x
- Ix version: 0.10.4
- Backend release: 1.0.28

Hiro-Chiba:
Fixes #523

## Summary
Skip generated dependency lockfiles during local source discovery so their expanded graph patches cannot exceed the proxy request-body limit.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Exclude `package-lock.json`, `npm-shrinkwrap.json`, and `pnpm-lock.yaml` as generated inputs.
- Apply the exclusion to git discovery, filesystem discovery, and explicit paths.
- Add a regression test covering the supported lockfile names.

## Validation
- `npm run typecheck`
- Re-ran a generic 646 KB lockfile reproduction. Map discovered two architectural inputs, exited 0, and produced no HTTP 413.

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/ingest-discovery.test.ts
+++ b/ix-cli/src/cli/__tests__/ingest-discovery.test.ts
@@ -67,22 +67,6 @@ describe('dedupeDiscoveredFilePaths', () => {
     }
   });
 
-  it('skips generated dependency lockfiles that expand into oversized graph patches', () => {
-    const root = join('/repo', 'project');
-    const discovery = discoverIngestFilePaths([
-      join(root, 'package.json'),
-      join(root, 'package-lock.json'),
-      join(root, 'npm-shrinkwrap.json'),
-      join(root, 'pnpm-lock.yaml'),
-      join(root, 'src', 'index.ts'),
-    ], root, (filePath) => filePath);
-
-    expect(discovery.files).toEqual([
-      join(root, 'package.json'),
-      join(root, 'src', 'index.ts'),
-    ]);
-  });
-
   it('confines a canonical path to the discovery root', () => {
     const root = join('/repo', 'project');
     expect(isWithinDiscoveryRoot(root, join(root, 'src', 'app.ts'))).toBe(true);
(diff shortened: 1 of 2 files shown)


KageBinary:

Reviewed the diff, the discovery chokepoints it touches, and the security invariants around `discoverIngestFilePaths`. The lockfile exclusion is the right call for #523; three things needed changing before merge, pushed as `c546948`.

**1. The filter was in the wrong place, and the one thing it changed was a regression.**

`walkFiles` and `tryGitLsFiles` are the only two things that produce a directory listing, and both already call `isGeneratedFile`. Adding the same filter inside `discoverIngestFilePaths` therefore changed nothing for a directory ingest — adding the names to `GENERATED_FILE_NAMES` was already the entire fix.

The one case it *did* change is the case that function documents as out of scope:

> `root` is `undefined` when a single file was named explicitly: that IS the request, and there is nothing for it to escape from.

The filter ran *before* that early return, so `ix ingest src/api.pb.go` — or any explicitly named `.pb.go`, `mock_*`, `zz_generated*` file — silently discovered zero files and reported `Warning: No files found. Check the path and supported extensions.` The path was fine and the extension was supported. Dropped the filter; `GENERATED_FILE_NAMES` carries the fix on its own.

**2. `packages.lock.json` was missing.** Of the lockfile families, only the ones whose extension is in `SUPPORTED_EXTENSIONS` can reach discovery at all. `yarn.lock`, `composer.lock`, `Cargo.lock`, `Gemfile.lock`, `poetry.lock`, `deno.lock` all end in `.lock`, which is not discovered — but NuGet's `packages.lock.json` is `.json`, is discovered, and is exactly as large. Added, with a comment recording why the `.lock` families are deliberately absent.

**3. Case-sensitive match.** `isSupportedSourceFile` lowercases the basename; `GENERATED_FILE_NAMES.has(basename)` did not. On a case-insensitive filesystem — macOS, Windows, i.e. both platforms in the issue — `Package-lock.json` is discovered and went straight through. Now lowercased for the name match only, leaving the suffix/prefix rules as they were.

**4. The test asserted against a function that is not the filter point.** It handed `discoverIngestFilePaths` a literal list, which is not how discovery reaches it in production. With the filter removed from that function the test would have passed while the real discovery path still yielded the lockfile. Rewritten to build a temp git repo and assert on `tryGitLsFiles` — the listing `runIngest` actually discovers from — including `packages.lock.json`.

**Security:** no concerns. The filter only ever removes entries, so the symlink-containment guarantee in `discoverIngestFilePaths` is untouched; moving it out leaves that function doing exactly one job again.

KageBinary (approved):
Reviewed for quality and security. Filter moved back to the discovery chokepoints, packages.lock.json added, case-insensitive match, and the test now asserts through tryGitLsFiles (c546948). CI green.

PR #530, opened 2026-08-28
fix: reject severely incomplete architecture maps

### LINKED ISSUE #524: [bug] Mixed-language repositories can record a severely partial architecture map as complete
## Description

The empty-map guard only rejects a completed response when `file_count` is zero. A mixed-language repository can therefore ingest every source file, receive an architecture hierarchy for only a small supported subset, and still persist a completed baseline.

A minimal repository containing one PHP source file and one JSON manifest reproduces this. The source graph contains the PHP class and method, but the map reports one of two files and `ix status` records `mapCompleted: true`. Locating the PHP class reports `hasMapData: false`.

## Steps to reproduce
1. Create `composer.json` and one PHP class under `src/`.
2. Run `ix map --full --format json`.
3. Run `ix status --format json`.
4. Run `ix locate <ClassName> --kind class --format json`.

## Expected behavior

A completed baseline is not recorded when the architecture hierarchy covers only a severe minority of the cleanly ingested source files.

## Actual behavior

The map returns `full_local_completed` with `file_count: 1`, while the clean ingest discovered and committed two files. Status reports `mapCompleted: true`, but the PHP class has no map hierarchy.

## CLI output

```json
{
  "file_count": 1,
  "region_count": 0,
  "outcome": "full_local_completed"
}
```

```json
{
  "mapCompleted": true,
  "staleFiles": 0
}
```

## Environment
- OS: macOS
- Node: 22.x
- Ix version: 0.10.4
- Backend release: 1.0.28

Hiro-Chiba:
Fixes #524

## Summary
Prevent a non-empty but severely incomplete architecture hierarchy from being recorded as a completed local map.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Compare completed map coverage with the clean local ingest count.
- Reject coverage of half or fewer discovered source files while allowing ordinary small omissions.
- Clear the local completion baseline so status remains unverified and the next run reparses the workspace.
- Add regression coverage for severe, ordinary, and ingest-failure cases.

## Validation
- `npm test` (1,419 passed, 3 skipped)
- `npm run typecheck`
- Re-ran a generic mixed PHP and JSON repository. The map exited 1 with `mapped only 1 of 2`, and status kept `mapCompleted: false`.

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/map-auto-skip.test.ts
+++ b/ix-cli/src/cli/__tests__/map-auto-skip.test.ts
@@ -7,7 +7,9 @@ import { afterEach, describe, expect, it, vi } from 'vitest';
 import {
   applyRequestedMapCoalesceExitCode,
   describeEmptyCompletedMap,
+  describeSeverelyIncompleteCompletedMap,
   invalidateBaselineForEmptyCompletedMap,
+  invalidateBaselineForIncompleteCompletedMap,
   mapModeForIngest,
   registerMapCommand,
   requestedMapCoalesceExitCode,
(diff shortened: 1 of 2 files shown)


KageBinary:

The bug in #524 is real, but the rule chosen to catch it fires on healthy maps. Replaced with the signal #524's own output already contains; pushed as `d121f53`.

### The coverage ratio rejects normal repositories

`filesDiscovered` counts every extension in `SUPPORTED_EXTENSIONS` — `.md`, `.json`, `.yaml`, `.css`, `.html`, `.tex` — while the hierarchy only ever covers files the backend can couple. A healthy map therefore covers a minority of what discovery found, routinely.

Measured on backend 1.0.28 with a 15-file fixture (6 linked `.ts`, 8 `.md`, `package.json`):

```
$ ix map . --full --format json
{ "file_count": 7, "region_count": 4, "outcome": "full_local_completed", ... }   # 4 named regions
```

`file_count * 2 > filesDiscovered` → `14 > 15` → **false**, so this map is rejected: `mapped only 7 of 15`, `exit 1`, completion marker cleared. Nothing is wrong with it.

#534 — same stack, same author — records the extreme version and treats it as the normal shape:

> the clean first map discovered about 7,400 supported files but the hierarchy covered only 381

That is 5% coverage. And because a rejection *clears the marker*, such a workspace can never record one: every run rejects, every run clears, `ix status` reads `map_complete=false` for ever, and with #531 `ix doctor` is permanently unhealthy with no `ix map` that can fix it.

### The signal is the region count

From #524's own CLI output:

```json
{ "file_count": 1, "region_count": 0, "outcome": "full_local_completed" }
```

`region_count: 0`. **No regions is not a hierarchy**, whatever the file count — and `ix locate <ClassName>` reporting `hasMapData: false` is the same fact from the other side. That is exact where a ratio is a guess, and it needs no notion of which discovered files were ever mappable. `describeSeverelyIncompleteCompletedMap` → `describeRegionlessCompletedMap`.

Both cases verified:

```
# #524 repro — one PHP class + composer.json
Error: Backend reported full_local_completed, but produced 0 regions while mapping
1 of 2 supported source files ... The source ingest baseline was preserved.
exit=1

# 15-file mixed fixture above
exit=0, map_complete=true
```

### Also

- **`persistCompletedMapBaseline` now refuses a regionless result too.** Unreachable behind the guard on the normal path, but reachable when the guard abstains — fewer than 2 files discovered, or a parse error — and recording completion for a hierarchy we would otherwise reject is the contradiction this guard exists to prevent.
- **`invalidateBaselineForEmptyCompletedMap` became a wrapper nothing calls.** Folded into `invalidateBaselineForIncompleteCompletedMap` so the tests exercise the function `ix map` actually runs, rather than a helper beside it.
- **Rebased onto #535**, so a rejection clears only the architecture marker and the clean source ingest survives — which is what makes the message's "the next `ix map` can reuse unchanged files" true. Confirmed on the retry: `All 2 files mtime-clean — skipping hash lookup`.
- Tests rewritten around the new signal, with the 381-of-7,400 case pinned as an explicit "must be accepted".

**Security:** none — this is a local classification of a backend response, no new I/O or inputs.

KageBinary (approved):
Reviewed for quality and security. The coverage ratio rejected healthy maps (measured: file_count 7 / 15 discovered, 4 regions, on a normal mixed repo); replaced with the region_count==0 signal from #524's own output in d121f53. Both cases verified against backend 1.0.28. CI green.

PR #531, opened 2026-08-28
fix: check map completion in doctor

### LINKED ISSUE #525: [bug] ix doctor reports healthy when the workspace has no completed map
## Description

`ix status` knows whether the active workspace has a completed local map baseline, but `ix doctor` only checks reachability, workspace registration, graph counts, conflicts, image, and schema. A partially committed graph with nodes and edges therefore passes every hard check.

This makes `ix doctor --format json` return `healthy: true` even while `ix status --format json` returns `mapCompleted: false` for the same workspace.

## Steps to reproduce
1. Create a workspace in which one of several file patches fails to commit.
2. Run `ix status --format json` and observe `mapCompleted: false`.
3. Run `ix doctor --format json` in the same directory.

## Expected behavior

Doctor includes the workspace's completion baseline in its health decision and reports the partial graph as unhealthy.

## Actual behavior

Doctor reports `healthy: true` because the partial graph still has nodes and edges.

## CLI output

```json
{
  "mapCompleted": false
}
```

```json
{
  "healthy": true,
  "checks": [
    { "name": "Graph has nodes", "ok": true },
    { "name": "Graph has edges", "ok": true }
  ]
}
```

## Environment
- OS: macOS
- Node: 22.x
- Ix version: 0.10.4
- Backend release: 1.0.28

Hiro-Chiba:
Fixes #525

## Summary
Make `ix doctor` include the active workspace's completed-map baseline in its health decision.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Add a `Completed map for this workspace` doctor check.
- Fail health when a registered workspace has no completed local baseline.
- Report the recorded revision when the baseline exists.
- Add regression coverage for a partial graph that still has nodes and edges.

## Validation
- `npm run typecheck`
- Ran doctor against a generic partial workspace. The result changed from `healthy: true` to `healthy: false` with the new failed check.

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/doctor-stats-once.test.ts
+++ b/ix-cli/src/cli/__tests__/doctor-stats-once.test.ts
@@ -16,7 +16,9 @@ const scope = {
   matched: undefined as FakeWorkspace | undefined,
   fallback: undefined as FakeWorkspace | undefined,
   systemId: undefined as string | undefined,
+  sourceBaseline: true,
   mapCompleted: true,
+  cloudReady: false,
 };
 
 vi.mock("../../client/api.js", () => ({
@@ -53,14 +55,29 @@ vi.mock("../resolve.js", async (orig) => ({
   resolveReadSystemId: async () => scope.systemId,
 }));
 
-vi.mock("../stale.js", () => ({
-  detectStaleFiles: () => ({
-    mapCompleted: scope.mapCompleted,
-    currentRev: scope.mapCompleted ? 42 : 0,
-    lastIngestAt: scope.mapCompleted ? "2026-01-01T00:00:00.000Z" : null,
-    staleFiles: 0,
-    sampleChangedFiles: [],
-  }),
+// The two completion markers, mocked separately because the check has to tell
+// them apart: a missing source baseline and a missing map marker are different
+// states with different verdicts.
+vi.mock("../stale.js", async (orig) => ({
+  ...(await orig<typeof import("../stale.js")>()),
+  hasCompletedMapBaseline: () => scope.mapCompleted,
+}));
+
+vi.mock("../ingest-baseline.js", async (orig) => ({
+  ...(await orig<typeof import("../ingest-baseline.js")>()),
+  loadIngestBaseline: () => scope.sourceBaseline
+    ? {
+      files: new Map<string, number>(),
+      deletedFiles: new Map<string, string[]>(),
+      currentRev: 42,
+      lastIngestAt: "2026-01-01T00:00:00.000Z",
+    }
+    : null,
+}));
+
+vi.mock("../remote.js", async (orig) => ({
+  ...(await orig<typeof import("../remote.js")>()),
+  isCloudReady: async () => scope.cloudReady,
 }));
 
 // Doctor also inspects the live container and probes the schema. Both are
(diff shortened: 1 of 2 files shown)


KageBinary:

Right gap to close — a partially committed graph passing every check is exactly the failure #506 was about. Three changes before merge, pushed as `430f699`.

**1. `detectStaleFiles` is the wrong tool for this question, and it is not cheap.**

It walks the entire workspace and `statSync`s every file, then `existsSync`s every path in the baseline — all to answer "which files changed". This check uses neither the walk nor that answer; it reads `mapCompleted` and `currentRev`, which are two small JSON reads. So on a large workspace this adds a full tree traversal to every `ix doctor`, in a command where the two `client.stats()` calls were deliberately collapsed into one because `/v1/stats` cost 3–4s. Reads `loadIngestBaseline` + `hasCompletedMapBaseline` directly now.

**2. Two different absences, one verdict.**

"No local baseline at all" and "clean source graph with no map for its revision" were both `no completed map baseline — the graph may be partial`. Post-#535 the second is the common one and the detail line did not describe it. Split, and the second names the revision:

```
check name="Completed map for this workspace" status=ok
  detail="recorded at revision 2466"
check name="Completed map for this workspace" status=fail
  detail="source graph at revision 2467 has no completed architecture map — run `ix map`"
```

**3. A cloud-ingested workspace would have been permanently unhealthy.**

The no-baseline case has three causes, and `hasCompletedMapBaseline`'s own docblock lists them: never mapped, an initial map that committed patches but never finished, or **a workspace ingested through the cloud runner, which writes no local baseline**. Only the last is healthy — and no `ix map` fixes it, because there is nothing to fix. That one now warns; everything else still hard-fails. `isCloudReady()` is a pure local check with no network I/O and returns false whenever no runner is registered, so OSS installs and #525's partial-commit repro are unaffected and still report `healthy=false`.

**Also:** resolve `matchedWorkspace.root_path` once. Both markers are keyed by the exact root string and `hasCompletedMapBaseline` resolves internally, so an unresolved `root_path` could have made the two lookups disagree — one finding a marker the other could not.

**Interaction with #535** (worth stating, not a defect): `mapCompleted` now means "an architecture map exists *for this source revision*", so the first `ix doctor` after upgrading reports this check failed until the next `ix map`. That is accurate — nothing has recorded a hierarchy for that revision — and one `ix map` clears it.

**Security:** the catch surfaces `e.message`, which can carry a local path. Consistent with the checks around it and it is the user's own path, so no change.

KageBinary:
### E2E caught a real defect — fixed in #536, now merged

The `E2E (real backend)` failure on this branch was not flaky. This PR is the first thing that reads #535's rev-tied map marker in anger, and it found that a **single-file workspace could never record map completion**:

```
map: 1 files · 0s/0ss/0m regions · 95ms
...
{ "name": "Completed map for this workspace", "ok": false,
  "detail": "source graph at revision 1 has no completed architecture map — run `ix map`" }
doctor not healthy
```

`ix-cli/test/fixtures/sample_project` is one Python file. There is no hierarchy to build over one file and no later `ix map` will produce one, so a tightening I made in #530 (`persistCompletedMapBaseline` refusing any regionless result) left it recording a source baseline and no map marker — `ix status` reading `map_complete=false` and doctor unhealthy for ever, with the suggested fix being the command that just ran. My regression, and the check added here is exactly what surfaced it.

Fixed in #536 (merged): the refusal is back to a *wholly* empty result, with a regression test on the one-file shape. The regionless-over-real-source case that #530 was aiming at is still rejected by `describeRegionlessCompletedMap`, which returns before persist is reached and which knows the discovered-file count that tells the two apart.

Branch rebased on that; full suite green locally (1444 passed).

KageBinary (approved):
Reviewed for quality and security. Swapped the full workspace walk for the two marker reads, split the two absences, and made the cloud-runner case warn rather than fail (430f699). The E2E failure was a real defect this check surfaced in #530's persist tightening — fixed and merged as #536. CI green.

PR #532, opened 2026-08-28
fix: fail unresolved machine-readable commands

### LINKED ISSUE #526: [bug] JSON query commands exit 0 with no output for missing targets
## Description

When entity resolution returns no match, `ix context`, `ix explain`, and `ix read` return from their handlers without setting a failure status or writing a JSON result. The resolver writes prose to stderr, but stdout remains empty and the process exits 0.

Machine callers cannot distinguish this semantic failure from a successful command that intentionally returned no records.

## Steps to reproduce
1. Map any small repository.
2. Run the following commands with a name that does not exist.
   - `ix context DefinitelyMissing --kind class --format json`
   - `ix explain DefinitelyMissing --kind class --format json`
   - `ix read DefinitelyMissing --format json`
3. Inspect stdout and the process status.

## Expected behavior

Each command emits a valid JSON error object and exits non-zero.

## Actual behavior

Each command emits no stdout and exits 0. Human-readable prose is written only to stderr.

## CLI output

```text
command  status  stdout
context  0       <empty>
explain  0       <empty>
read     0       <empty>
```

## Environment
- OS: macOS
- Node: 22.x
- Ix version: 0.10.4
- Backend release: 1.0.28

Hiro-Chiba:
Fixes #526

## Summary
Return a structured machine-readable failure when `context`, `explain`, or `read` cannot resolve the requested target.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Add a shared unresolved-target reporter for JSON and LLM formats.
- Emit a JSON error object from the three affected commands.
- Set the process status to 1 instead of returning a false success.
- Add command-level regression tests for all three commands.

## Validation
- `npm test` (1,418 passed, 3 skipped)
- `npm run typecheck`
- Re-ran all three generic missing-target commands. Each emitted valid JSON and exited 1.

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/unresolved-machine-output.test.ts
+++ b/ix-cli/src/cli/__tests__/unresolved-machine-output.test.ts
@@ -1,0 +1,55 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { Command } from "commander";
+
+vi.mock("../../client/api.js", () => ({
+  IxClient: class {
+    async workspaceSystem() { return { systemId: null }; }
+    async search() { return []; }
+  },
+}));
+
+type Register = (program: Command) => void;
+
+async function run(register: Register, args: string[]): Promise<{ stdout: string; stderr: string }> {
+  const program = new Command();
+  program.name("ix").exitOverride();
+  register(program);
+  const stdout: string[] = [];
+  const stderr: string[] = [];
+  const log = vi.spyOn(console, "log").mockImplementation((...parts) => stdout.push(parts.join(" ")));
+  const error = vi.spyOn(console, "error").mockImplementation((...parts) => stderr.push(parts.join(" ")));
+  try {
+    await program.parseAsync(args, { from: "user" });
+  } finally {
+    log.mockRestore();
+    error.mockRestore();
+  }
+  return { stdout: stdout.join("\n"), stderr: stderr.join("\n") };
+}
+
+describe("unresolved targets in machine formats", () => {
+  let savedExitCode: number | string | undefined;
+
+  beforeEach(() => {
+    savedExitCode = process.exitCode;
+    process.exitCode = undefined;
+  });
+
+  afterEach(() => {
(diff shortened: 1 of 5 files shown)


KageBinary:

Verified against a real backend on the merged branch; no code changes needed.

```
$ ix read DefinitelyMissing --format json
{
  "error": "unresolved_target",
  "message": "No entity found matching \"DefinitelyMissing\"."
}
exit=1

$ ix context DefinitelyMissing --format llm
error code=unresolved_target message="No entity found matching \"DefinitelyMissing\"."
exit=1
```

Three things I checked specifically:

- **`ix read` text mode.** The PR replaces `stderr("Could not resolve ... as a file or symbol.")` with a helper that prints nothing for `text`, which looked like it would leave a silent `exit 1`. It does not: `resolveEntityFull` already writes `No entity found matching "X".` to stderr on every not-found path, so text mode now prints one message where it used to print two. Confirmed on a live workspace — the line is still there and the status is now 1.
- **stdout/stderr separation.** The resolver's prose goes to stderr and the structured record to stdout, so `--format json | jq` never sees prose. Correct.
- **llm-format injection.** `target` is interpolated into an llm record, but `llmLine` escapes `\`, `"`, newlines and tabs, so a target containing a quote cannot forge a field. No issue.

**Windows CI failure is unrelated.** `mcp-install.test.ts > registers a host whose name is free` timed out at 5000ms. That test does a real PATH probe (`bin: "node"`) and spawns a process; the repo already carries a comment in `doctor-stats-once.test.ts` about this exact 5s-on-Windows-only pattern. Nothing in this diff is on that path. Re-ran the job.

KageBinary (approved):
Reviewed for quality and security. Verified the JSON/llm error records and exit 1 against a real backend; text mode still prints the resolver's line. The Windows CI failure was an unrelated 5s timeout in mcp-install.test.ts and passed on re-run.

PR #533, opened 2026-08-28
fix: surface cross-workspace stitch failures

### LINKED ISSUE #528: [bug] Cross-workspace stitch failures are silent without verbose output
## Description

Local ingest treats cross-workspace stitch as best-effort and catches every error. The catch branch writes a diagnostic only when debug or verbose output is enabled. In normal and silent map modes, a failed stitch produces no warning and leaves the process status at zero.

The source graph is valid, but cross-repository relationships may be incomplete. Machine callers are told the map succeeded with no indication that this secondary graph step failed.

A previously observed large-graph run returned HTTP 504 after a long stitch request. The deterministic reproduction below uses a local proxy that returns 504 only for `/v1/stitch` and forwards every other endpoint. This issue is about the hidden failure, not a claim that graph size alone always causes a timeout.

## Steps to reproduce
1. Create a minimal package with a package name and one exported TypeScript function.
2. Place a local proxy in front of the backend. Forward every endpoint normally, but return HTTP 504 from `POST /v1/stitch`.
3. Point Ix at that proxy.
4. Run `ix map --full --silent` without `--verbose`.

## Expected behavior

The command reports that cross-workspace stitching failed and exits non-zero, while clarifying that source patches were committed.

## Actual behavior

The command prints only the normal map summary and exits 0.

## CLI output

```text
map: 2 files · 0s/0ss/0m regions · 289ms
exit: 0
```

## Environment
- OS: macOS
- Node: 22.x
- Ix version: 0.10.4
- Backend release: 1.0.28

Hiro-Chiba:
Fixes #528

## Summary
Surface cross-workspace stitch failures even when ingest output is suppressed by normal or silent map mode.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Track stitch failures separately from source parse and commit failures.
- Print a bounded warning that distinguishes committed source patches from incomplete cross-repository relationships.
- Exit non-zero so machine callers can detect the degraded graph.
- Include `stitchErrors` in JSON ingest output and add regression coverage for bounded error rendering.

## Validation
- `npm run typecheck`
- Routed a generic map through a proxy that returns HTTP 504 only from `/v1/stitch`. The command printed the bounded warning and exited 1.

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/ingest-commit-outcome.test.ts
+++ b/ix-cli/src/cli/__tests__/ingest-commit-outcome.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from "vitest";
 
-import { describeCommitOutcome, describeStitchFailure, ingestCompletedCleanly } from "../commands/ingest.js";
+import { describeCommitOutcome, describeStitchFailure, ingestCompletedCleanly, isStitchUnsupported } from "../commands/ingest.js";
 
 describe("describeCommitOutcome", () => {
   it("says nothing when every patch committed", () => {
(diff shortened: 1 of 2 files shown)


KageBinary:

Surfacing a silent stitch failure is right, and not dumping the proxy's HTML body into the map summary is the right instinct. Two defects in how the failure is classified; pushed as `79e5669`.

**1. "Backend has no `/v1/stitch`" is now reported as a degraded graph.**

The catch this PR instruments has always covered two different things, and the comment directly above the block says so:

> Best-effort: a failure (**or an older backend with no `/v1/stitch`**) never fails the ingest.

Counting both as `stitchErrors` means a user on a backend without the endpoint gets, on **every full `ix map`**:

```
Warning: Cross-workspace stitch failed (HTTP 404). Source patches were committed,
but cross-repository relationships may be incomplete.
```
and `exit 1` — for a step that was never going to run. `ix map --silent` is documented as the hook/agent entry point and those callers read the exit code, so this turns every map into a failure for them. #528 is about a stitch that *failed*, not about one that was never implemented.

Added `isStitchUnsupported`: 404/501 skip silently exactly as before; 5xx, timeouts and transport errors surface and exit non-zero, which is what the issue asked for.

**2. The status regex was not anchored.**

```js
/(?:^Error:\s*)?(\d{3}):/
```
The `^` is *inside* the optional group, so the whole pattern still matches anywhere in the string. Any message ending `...1234: foo` reports `HTTP 234`. `IxClient` only ever puts the status at the front (`${resp.status}: ${text}`), so the pattern is now `/^(?:Error:\s*)?(\d{3}):/` and anything else reads `backend request failed`. Tests cover 404/501 vs 5xx vs a transport error with no status.

**Security:** good as-is, and worth noting explicitly — `describeStitchFailure` deliberately reports only the status and never the response body. The observed failure is a proxy 504 whose body is a full HTML error page, which can carry upstream hostnames and internal paths; keeping it out of a summary an agent may forward is the right default. The added test pins that.

KageBinary (approved):
Reviewed for quality and security. Two classification defects fixed in 79e5669 (unsupported endpoint no longer reported as a failure; status regex anchored). CI green.

PR #535, opened 2026-08-28
fix: preserve source graph after architecture map failure

### LINKED ISSUE #534: [bug] Preserve source graph baseline when architecture mapping fails
## Description

Ix currently uses the successful source-ingest cache as the architecture-map completion baseline. When a clean ingest succeeds but the backend returns an empty completed hierarchy, the CLI correctly rejects the hierarchy but deletes the successful ingest cache as well.

This conflates two independent states. Source-level queries can be complete and current even when architecture hierarchy generation is unsupported or incomplete. Deleting that baseline also forces every unchanged source file to be parsed and uploaded again on the next `ix map`.

## Steps to reproduce

1. Create a workspace containing supported PHP source files and no non-PHP manifest files.
2. Run `ix map . --silent` against backend release 1.0.28.
3. Observe that source ingestion succeeds and the backend returns `full_local_completed` with zero mapped files.
4. Run `ix status`, then run the same map again without changing any file.

## Expected behavior

The source graph baseline remains available after hierarchy generation fails. Status reports source graph completion separately from architecture-map completion, source-level reads remain current, and the next map reuses unchanged files.

## Actual behavior

The empty-map guard deletes the ingest baseline. Status can no longer distinguish a valid source graph from a completed hierarchy, context is classified as unverified, and the next map reparses every file.

On a larger anonymous PHP workspace, the clean first map discovered about 7,400 supported files but the hierarchy covered only 381. Clearing the shared baseline makes a retry repeat roughly 198 seconds of source work that can otherwise be reused in about one second.

## CLI output

```text
Backend reported full_local_completed, but mapped 0 files after local ingest found 2 supported source files (2 patches committed).
The source graph was ingested, but no architecture hierarchy was created.
```

## Environment

- OS: macOS
- Node: 22
- Ix CLI: 0.10.4 / current main
- Ix backend: 1.0.28

Hiro-Chiba:
Closes #534

## Summary

Separates source-graph completion from architecture-map completion so a failed hierarchy build no longer discards a clean source ingest or forces a full reparse on retry.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [x] Test
- [ ] CI

## Changes
- Add a workspace-scoped architecture-map marker tied to the current source revision.
- Preserve the source ingest baseline when an empty completed hierarchy is rejected.
- Report `graphCompleted` and `mapCompleted` independently in JSON and LLM status output.
- Base source-graph staleness and context freshness on the clean ingest baseline.
- Clear both baselines on reset or workspace migration, while hierarchy-only failures clear only the map marker.
- Record map completion only for non-empty successful outcomes, excluding backend guardrail refusals.

This also provides the hierarchy-only invalidation needed by #530 when a severely partial map is rejected.

## Validation

- `npm test`: 83 test files passed, 1421 tests passed, 3 skipped; parser smoke passed.
- `npm run typecheck`: passed.
- `npm run build`: passed.
- `npm run lint`: 0 errors; 41 existing warnings.
- PHP-only fixture against backend 1.0.28: first map ingested 2 files and correctly rejected an empty hierarchy; status reported `graph_complete=true map_complete=false`; unchanged retry skipped both files and completed the same hierarchy check in 0.34 seconds.
- Anonymous large PHP workspace: clean source baseline held about 7,400 files while the hierarchy covered 381; after hierarchy-only invalidation, PHP search and context remained current and an unchanged retry dropped from about 198 seconds to about 1 second.

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/context.test.ts
+++ b/ix-cli/src/cli/__tests__/context.test.ts
@@ -225,12 +225,8 @@ describe("ix context bundle", () => {
     ]);
   });
 
-  it("classifies a workspace with no completed map baseline as unverified", () => {
-    // Ix#506: the backend answers from graph patches an initial map committed
-    // before it failed. `ix status` already reported mapCompleted:false while
-    // the bundle said current, so an agent reading the bundle trusted a graph
-    // the CLI knew was half-built.
-    const bundle = buildBundle({ ...input(), mapCompleted: false });
+  it("classifies a workspace with no completed source graph as unverified", () => {
+    const bundle = buildBundle({ ...input(), graphCompleted: false });
 
     expect(bundle.freshness.classification).toBe("unverified");
   });
@@ -239,9 +235,9 @@ describe("ix context bundle", () => {
     // `stale` is a claim that a file changed since ingest. Without a baseline
     // nothing is known to have changed, so the two states stay distinguishable:
     // an agent can re-map on unverified without being told its files are dirty.
-    const unverified = buildBundle({ ...input(), mapCompleted: false });
+    const unverified = buildBundle({ ...input(), graphCompleted: false });
     const stale = buildBundle({
-      ...input(), mapCompleted: true, facts: makeFacts({ stale: true }),
+      ...input(), graphCompleted: true, facts: makeFacts({ stale: true }),
     });
 
     expect(unverified.freshness.classification).toBe("unverified");
--- a/ix-cli/src/cli/__tests__/llm-tier5.test.ts
+++ b/ix-cli/src/cli/__tests__/llm-tier5.test.ts
@@ -137,6 +137,7 @@ describe("read --format llm", () => {
 describe("status --format llm", () => {
   it("answers 'can I trust the graph' in one field", () => {
     const lines = renderStatusLlm("ok", "http://localhost:8090", {
+      graphCompleted: true,
       mapCompleted: true,
       currentRev: 11, lastIngestAt: "2026-08-09T12:00:00.000Z",
       staleFiles: 2, sampleChangedFiles: ["src/a.ts", "src/b.ts"],
@@ -151,6 +152,7 @@ describe("status --format llm", () => {
 
   it("says stale=false rather than omitting it when the graph is current", () => {
     const lines = renderStatusLlm("ok", "http://localhost:8090", {
+      graphCompleted: true,
       mapCompleted: true,
       currentRev: 11, lastIngestAt: null, staleFiles: 0, sampleChangedFiles: [],
     });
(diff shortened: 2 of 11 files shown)


KageBinary:

The strongest of the batch, and the one the other map PRs should sit on top of. Reviewed the design, the marker lifecycle and the interaction with the rest of the open stack; no changes needed beyond a `main` merge for the `context.ts` import conflict with #532 (`490d6e2`).

**Verified end to end** against backend 1.0.28 on a PHP + JSON fixture reproducing #534:

```
$ ix map . --full            # first run
Error: Backend reported full_local_completed, but produced 0 regions while mapping
1 of 2 supported source files ... The source ingest baseline was preserved, so the
next 'ix map' can reuse unchanged files.
exit=1

$ ix map . --full            # unchanged retry
  All 2 files mtime-clean — skipping hash lookup      <- baseline survived
exit=1

$ ix status --format llm
status backend=ok ... graph_complete=true map_complete=false rev=2467 stale=false
```

and on a healthy TypeScript fixture: `graph_complete=true map_complete=true`. The split is the right one — the two states really are independent, and conflating them is what made a failed hierarchy throw away a clean 7,400-file ingest.

**Lifecycle checked:**
- `clearIngestMtimeCache` clearing both is correct for its two callers (`reset`, migrated-workspace refresh) — both genuinely invalidate the source graph.
- `persistCompletedMapBaseline` runs after `bootstrap` on the **cloud** path too, but `loadIngestBaseline` returns null there, so a cloud-ingested workspace records neither marker rather than a stale one. Right outcome.
- Marker keying matches the ingest baseline exactly (`sha256(projectRoot).slice(0,12)` under `~/.ix`, `root` re-checked on load), so a mismatch reads as "absent" rather than as another workspace's answer.
- `hasCompletedMapBaseline` is now rev-tied, so a plain `ix ingest` after a map correctly flips `map_complete` to false. Intended, and `ix status` says so in words.

**Security:** the new marker stores only the project root and an integer rev, in the directory that already holds the ingest baseline, with the same permissions. No new attack surface.

One consequence worth stating for the release notes rather than fixing: an existing workspace has an ingest baseline but no map marker, so the first `ix status` after upgrading reports `map_complete=false` until the next `ix map`. That is accurate — nothing has recorded a hierarchy for that revision — and one `ix map` clears it.

KageBinary (approved):
Reviewed for quality and security; verified end to end against backend 1.0.28 on a #534 fixture. Marker lifecycle, cloud path and key derivation all check out. Merged main for the context.ts conflict with #532 (490d6e2). CI green.

PR #543, opened 2026-08-30
fix(text): enforce workspace-relative search paths

### LINKED ISSUE #542: [bug] ix text resolves --path outside --root and can search outside the workspace
## Description

`ix text --root <workspace> --path <relative-path>` resolves the path from the process working directory instead of from `--root`. An absolute `--path` outside the workspace is also accepted, so the command can search files outside the selected workspace.

## Steps to reproduce

1. Create `/tmp/ix-fixture/src/example.ts` containing a unique term.
2. Change to a different directory.
3. Run `ix text <term> --root /tmp/ix-fixture --path src --format json`.
4. Run `ix text localhost --root /tmp/ix-fixture --path /etc --format json`.

## Expected behavior

Relative paths are resolved from `--root`. Paths that escape the selected workspace, including symlink escapes, return a structured error and a non-zero exit status.

## Actual behavior

The relative path is resolved from the current directory. Absolute paths outside the workspace are searched successfully with exit status 0.

## CLI output

The first command searches the wrong `src` directory. The second can return matches from `/etc`.

## Environment

- OS: macOS
- Node: 22.x

Hiro-Chiba:
Fixes #542

## Summary
Resolve `ix text --path` from the selected workspace root and reject paths that escape it.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Resolve relative search paths against `--root` instead of the process working directory.
- Apply lexical and real-path containment checks before invoking ripgrep.
- Return structured JSON and LLM errors with exit status 1 for path escapes.
- Emit successful result paths relative to the selected workspace with POSIX separators on every platform. This changes the existing JSON path field from absolute to workspace-relative, including the default `--path .` case.
- Document that `--path` is workspace-root-relative, including when the command is launched from a subdirectory.
- Add command-level regression coverage for an accepted in-workspace path without depending on a system ripgrep binary.

## Plugin compatibility
Audited the current `ix-claude-plugin`, `ix-codex-plugin`, `ix-cursor-plugin`, `ix-gemini-plugin`, `ix-openclaw-plugin`, and `ix-opencode-plugin` consumers. Their text-search consumers display, forward, or reduce the path to a basename; none requires an absolute path. Several fixtures already use workspace-relative paths.

## Known limitation
A non-existent path below a symlinked workspace root can still be reported as `path_outside_workspace` because the existing containment helper falls back to a lexical candidate when `realpath` fails. This fails closed and is not a traversal bypass. It is intentionally left for the separate follow-up offered in review.

## Validation
- `npm test` (85 files, 1449 passed, 3 skipped; parser smoke passed)
- `npm run typecheck`
- `npm run build`
- `npm run knip`

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/text-workspace-boundary.test.ts
+++ b/ix-cli/src/cli/__tests__/text-workspace-boundary.test.ts
@@ -3,7 +3,7 @@ import { Command } from "commander";
 import { mkdtempSync, mkdirSync, rmSync, symlinkSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
-import { registerTextCommand } from "../commands/text.js";
+import { registerTextCommand, resolveTextSearchPath } from "../commands/text.js";
 
 describe("text workspace boundary", () => {
   let fixture: string;
@@ -44,13 +44,8 @@ describe("text workspace boundary", () => {
     await program.parseAsync(["node", "ix", ...args]);
   }
 
-  it("resolves --path relative to --root instead of the process cwd", async () => {
-    await run(["text", "boundaryNeedle", "--root", workspace, "--path", "src", "--format", "json"]);
-
-    const result = JSON.parse(logs.join("\n"));
-    expect(result).toHaveLength(1);
-    expect(result[0].path).toBe("src/inside.ts");
-    expect(process.exitCode).toBeUndefined();
+  it("resolves --path relative to --root instead of the process cwd", () => {
+    expect(resolveTextSearchPath(workspace, "src")).toBe(join(workspace, "src"));
   });
 
   it("rejects an absolute search path outside the explicit workspace", async () => {
--- a/ix-cli/src/cli/__tests__/text-workspace-boundary.test.ts
+++ b/ix-cli/src/cli/__tests__/text-workspace-boundary.test.ts
@@ -12,6 +12,7 @@ describe("text workspace boundary", () => {
   let originalCwd: string;
   let originalExitCode: number | string | undefined;
   let logs: string[];
+  let runRipgrep: ReturnType<typeof vi.fn<(args: string[]) => Promise<{ stdout: string }>>>;
 
   beforeEach(() => {
     fixture = mkdtempSync(join(tmpdir(), "ix-text-boundary-"));
@@ -29,6 +30,16 @@ describe("text workspace boundary", () => {
     vi.spyOn(console, "log").mockImplementation((...args: unknown[]) => {
       logs.push(args.join(" "));
     });
+    runRipgrep = vi.fn(async () => ({
+      stdout: JSON.stringify({
+        type: "match",
+        data: {
+          path: { text: join(workspace, "src", "inside.ts") },
+          line_number: 1,
+          lines: { text: "export const boundaryNeedle = true;\n" },
+        },
+      }),
+    }));
   });
 
   afterEach(() => {
@@ -40,12 +51,18 @@ describe("text workspace boundary", () => {
 
   async function run(args: string[]): Promise<void> {
     const program = new Command();
-    registerTextCommand(program);
+    registerTextCommand(program, runRipgrep);
     await program.parseAsync(["node", "ix", ...args]);
   }
 
-  it("resolves --path relative to --root instead of the process cwd", () => {
+  it("resolves --path relative to --root instead of the process cwd", async () => {
     expect(resolveTextSearchPath(workspace, "src")).toBe(join(workspace, "src"));
+
+    await run(["text", "boundaryNeedle", "--root", workspace, "--path", "src", "--format", "json"]);
+
+    expect(runRipgrep).toHaveBeenCalledWith(expect.arrayContaining(["boundaryNeedle", join(workspace, "src")]));
+    expect(JSON.parse(logs.join("\n"))).toMatchObject([{ path: "src/inside.ts" }]);
+    expect(process.exitCode).toBeUndefined();
   });
 
   it("rejects an absolute search path outside the explicit workspace", async () => {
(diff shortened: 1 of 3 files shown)


KageBinary:
## Review — sound fix, three things to weigh before merge

The containment logic is right and `isPathInsideResolvedRoot` is a good extraction: `isReadablePath` was doing exactly this inline, and hoisting it keeps one definition of "inside, after symlinks" rather than two that can drift. The symlink-escape test is the one that matters and it is there.

### 1. Machine output changed shape, and the PR body understates it

```ts
path: path.relative(root, absoluteFilePath),
```

Before this change the default `--path .` case already passed an absolute `searchPath` to rg, so `--format json` emitted **absolute** paths. They are now workspace-relative. That is a better contract and I would keep it — but it is a wire change for every existing `ix text --format json` consumer, not just a normalisation of the newly-resolved case. Worth a line in the body, and worth checking against the plugins that call `ix text` before this ships.

### 2. `--path` is now root-relative, which breaks running from a subdirectory

This is what #542 asked for, so I am not arguing against it — but the consequence is not written down anywhere. From `/repo/src`:

```
ix text TODO --path components      # was /repo/src/components, now /repo/components
```

and when `/repo/components` does not exist the user gets rg's failure surfaced as `Error: ripgrep failed: ... No such file or directory`, which does not explain what happened. Two options: mention the new resolution rule in the `--path` help text, or check existence and emit a structured error the way the escape path does. The second is more consistent with the rest of the PR.

### 3. A non-existent path under a symlinked root reports the wrong reason

`real()` falls back to a lexical resolve when `realpathSync` throws, which it does for any path that does not exist yet:

```ts
const real = (p: string) => { try { return realpathSync(p); } catch { return resolvePath(p); } };
```

On macOS, where `/tmp` → `/private/tmp` and `/var` → `/private/var`, that mixes a realpath'd root with a lexical candidate:

- `root = /tmp/ws` → `real(root) = /private/tmp/ws`
- `searchPath = /tmp/ws/typo` → throws → falls back to `/tmp/ws/typo`
- `isPathInside("/private/tmp/ws", "/tmp/ws/typo")` → **false** → `path_outside_workspace`

So a typo'd subdirectory under a symlinked root is reported as an escape attempt. It fails *closed*, so it is not a security hole — but it is a confusing error, and `ix text` is much more likely to be handed a non-existent path than `read` is. It is pre-existing in `isReadablePath`; this PR is what exposes it to a new command. The standard fix is to realpath the longest existing ancestor and re-append the remainder, which stays closed against symlink escapes because every symlink in the existing prefix is still resolved. Happy to do it as a follow-up if you want it separate.

## Smaller notes

- The three-way `json` / `llm` / stderr switch here is hand-rolled. `ui.ts` already has `reportFailure` (llm + stderr + exit code) and `reportUnresolvedTarget` (json + llm + exit code) — neither covers all three, so this is not a straight reuse miss, but #545 and #553 in this same stack each hand-roll it again. Four copies is the point at which one `reportError(code, message, format)` in `ui.ts` pays for itself.
- No positive test that a legitimate in-workspace `--path` is *accepted*. Both new containment tests assert rejection; the only acceptance check is the pure `resolveTextSearchPath` unit. Worth one case, especially given point 3 above.

## Separately: the other untrusted input to this command

While reading this I found that `term` reaches ripgrep's argv unseparated — `rgArgs.push(term, searchPath)` — so a term beginning with `-` is parsed by rg as a flag. Reachable through `ix mcp`, where the term is a string a model chose. `--files` runs rg's file listing; `--pre=<cmd>` takes the path argument away and leaves rg reading stdin for ever, which blocks the serialised MCP runner and leaks the child.

That is independent of this PR and reproduces on `main`, so it is **#564** rather than a request here. It touches only the `rgArgs.push` line, which is untouched context in this diff — the two merge cleanly in either order.

Hiro-Chiba:
Thanks! I addressed the items on my side and left the symlink follow-up untouched as suggested. The updated CI is passing.

KageBinary:
Reviewed and verified locally. Merging.

**What it does, restated:** `--path` becomes workspace-relative rather than cwd-relative, an out-of-workspace search path is refused with `path_outside_workspace` + exit 1, and result paths come back relative to the root.

**Why the `--path` semantics change is the right one.** Every other command that takes `--path` (`locate`, `inventory`, `overview`, `rank`, `search`, `explain`, …) treats it as a filter against graph source paths, which are workspace-relative. `ix text` was the only one resolving it against the process cwd. This aligns it. The default case is unchanged — `--path .` already meant the workspace root on main.

**The boundary is a real surface, not a hypothetical.** `ix text` returns file *snippets*, and the MCP servers in the plugin repos expose `ix_text` with `path` taken straight from the model. Before this PR a model could ask for `--path ~/.ssh` and get content back. Same family as `isReadablePath` on `ix read`.

**Plugin exit-code check (CONTRIBUTING → CLI Standards → Exit codes).** Ran it across all six repos. `ix text` is called programmatically in five of them, and this PR's new non-zero exit is reachable in all five:

| caller | behaviour on the refusal |
|---|---|
| cursor `mcp/tools/text.ts` | `Array.isArray(parsed) ? … : []` → clean "no matches" |
| opencode `tools/ix-locate.ts` | `parsed.hits ?? []` → clean "no matches" |
| openclaw `tools/ix-locate.ts` | `parsed.hits ?? []` → clean "no matches" |
| claude `hooks/ix-lib.sh` | `wait $PID || …` is guarded, stdout still read — but see below |
| gemini | prose only, no runner |

Note `ix text` already exits 1 today (rg missing, rg usage error), so non-zero is not new to these callers — what is new is a non-array on stdout.

**One finding, plugin-side and not a blocker.** `ix-claude-plugin/hooks/ix-lib.sh:329` summarises with `jq 'length'`, which on `{error, message}` returns **2**, so the intercept hook would inject `2 text hits` with no filenames when ix actually refused. Reproduced directly. It is the plugin's shape assumption — the three TS consumers all discriminate correctly — and it only fires when someone greps outside their ix workspace, so it degrades to one vague context line rather than anything acting on wrong data. Filed as a follow-up; it needs a `type == \"array\"` guard there, and that repo's merges are still stuck on the ruleset.

**Verified locally:** typecheck clean, ix-cli 1450 passed / 2 skipped. Built the branch and ran it against a scratch `HOME` — default search returns `src/inside.ts` (root-relative, as intended), an absolute out-of-workspace `--path` returns the `path_outside_workspace` record, and a cwd outside any workspace still searches normally.

PR #545, opened 2026-08-30
fix(map): canonicalize and validate workspace roots

### LINKED ISSUE #544: [bug] ix map can register duplicate or invalid workspace roots
## Description

`ix map` derives workspace identity from the invocation path without first selecting and canonicalizing the repository root. Running it from a nested directory can register that directory as a second workspace, while a symlink to the same repository can produce another workspace ID. A nonexistent explicit path is registered before the command fails.

## Steps to reproduce

1. Initialize a Git repository with a `src` subdirectory.
2. Run `ix map` from the repository root.
3. Change to `src` and run `ix map` without a path.
4. Run `ix map` through a symlink to the same repository.
5. Run `ix map /tmp/path-that-does-not-exist`.

## Expected behavior

Without an explicit path, map uses the nearest registered workspace root or Git root. Workspace paths and lock keys use the canonical real path. Missing and non-directory paths fail before workspace registration.

## Actual behavior

Nested and symlink paths can become separate workspaces for the same repository. A missing path can remain in configuration after map exits with an error.

## CLI output

The workspace configuration contains multiple roots for one physical repository, or a root that does not exist.

## Environment

- OS: macOS
- Node: 22.x

Hiro-Chiba:
Fixes #544

## Summary
Select, validate, and canonicalize the workspace root before `ix map` registers a workspace or acquires its lock.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Reuse the same root-resolution cascade as the other CLI commands: explicit path, nearest registered workspace, named workspace, default workspace, Git root, then cwd.
- Use `execFileSync` for Git root discovery instead of a shell command.
- Reject missing and non-directory paths before bootstrap can persist them.
- Canonicalize workspace roots at registration and lookup, preventing symlink aliases or Windows path aliases from creating an unfindable or duplicate workspace entry.
- Canonicalize single-flight lock keys so equivalent paths share one map lock.
- Emit structured JSON and LLM errors for invalid map paths, with a command-level JSON regression test.
- Cover named/default workspaces, nested Git roots, symlinks, registration migration, missing paths, files, and lock identity.

## Behavior change
Running `ix map` without a path from a subdirectory now maps the selected workspace or Git root. This also applies to `IX_AUTO_MAP=1` callers. `ix map .` remains the explicit way to map only the current directory.

An existing registry entry stored under a non-canonical alias is normalized instead of duplicated. If its path-derived workspace ID changes, the existing migration path forces a re-ingest and cleans up the previous workspace ID.

## Validation
- `npm test` (85 files, 1454 passed, 3 skipped; parser smoke passed)
- `npm run typecheck`
- `npm run build`
- `npm run knip`

## Checklist
- [x] Tests pass
- [x] Smoke tests pass


KageBinary:
## Review — correct, with two consistency risks worth a decision

Validating the path *before* bootstrap can persist it is the right shape, and `findGitRoot` using `execFileSync` with `stdio: ["ignore","pipe","ignore"]` is strictly better than the `execSync` (shell) that `resolveWorkspaceRoot` still uses for the same question. The lock-key canonicalisation is a real fix — a symlinked root and its target used to take different locks and could map concurrently.

### 1. `ix map` and every other command now use *different* cascades

`resolveMapRoot` is:

```
pathArg -> registered workspace -> git root -> cwd
```

`resolveWorkspaceRoot`, which `text`, `read`, `context`, `doctor` and the rest use, is:

```
--root -> registered workspace -> named workspace (ix config set workspace) -> default workspace -> git root -> cwd
```

The two middle steps are missing. So for a user who has set a named or default workspace and runs from a directory no workspace covers but a git repo does, `ix map` now maps the **git root** while `ix text` answers for the **named workspace**. Before this PR `ix map` used `process.cwd()` — obviously different, so nobody could mistake them for the same rule. Half-aligning is the state where they look identical and are not.

Reusing `resolveWorkspaceRoot` and wrapping it in `canonicalMapRoot` would give one cascade and keep everything else in the PR.

### 2. `realpathSync.native` can register a root string that nothing else matches

`canonicalMapRoot` returns `realpathSync.native(resolved)`, and that value becomes the registered `root_path`. But `selectWorkspaceForCwd` matches **lexically**:

```ts
workspaces.filter(w => isPathInside(w.root_path, cwd))
```

For the no-arg case this is fine — POSIX `process.cwd()` is already canonical, so the old code was effectively doing the same thing. For `ix map <path>` it is new: `ix map /tmp/ws` used to register `/tmp/ws` and now registers `/private/tmp/ws` on macOS.

Two consequences:

- **Windows.** `realpathSync.native` expands 8.3 short names, and `process.cwd()` may report the short form. `C:\Users\RUNNER~1\proj` registered as `C:\Users\runneradmin\proj` is not lexically inside itself, so other commands would not find the workspace. CI is green, and real users do not type 8.3 paths, but CI does — we have been bitten by exactly this `RUNNER~1` / `runneradmin` split before.
- **Existing registries.** `bootstrap.ts` matches on `w.root_path === rootPath`, so a workspace previously registered under a non-canonical path gets a **second** entry rather than being updated. Harmless (the longest match wins and both point at the same directory) but it leaves duplicates behind, and it is invisible until someone reads `~/.ix/config.json`.

Neither is a blocker; both argue for canonicalising `root_path` in one place — at registration — rather than at one caller.

### 3. Behaviour change worth putting in the body

`ix map` from a subdirectory now maps the whole repo instead of that subdirectory. That is the fix, and `ix map .` still scopes to cwd, so there is an escape hatch. But `IX_AUTO_MAP=1` editor hooks fire `ix map` from wherever the editor is, so a hook that used to ingest one subtree now ingests the git root on every change. Please confirm that is intended before this ships — the cost lands on people who never read the changelog.

### Smaller

- Third hand-rolled `json`/`llm`/stderr switch in this stack (see my note on #543). One `reportError` in `ui.ts` would cover all of them.
- No test that `ix map` actually *emits* `invalid_map_path` in json or llm — the new tests all exercise the resolver directly. One end-to-end case would pin the wire format the same way the resolver cases pin the logic.
- `lockPathFor` now canonicalises too, which is redundant once `resolveMapRoot` has already done it — harmless, and it does protect the other caller. Worth a one-line comment saying which one it is for.

Hiro-Chiba:
Addressed in `546b312`.

- `resolveMapRoot` now reuses `resolveWorkspaceRoot`, so named and default workspaces are selected before the Git root instead of maintaining a second cascade.
- Git root discovery in the shared resolver now uses `execFileSync` too.
- Workspace roots are canonicalized at registration and lookup, so symlink and Windows aliases do not make an existing entry unfindable or create a duplicate. The map working root and lock key remain canonical as well.
- The body now states that no-argument map and `IX_AUTO_MAP=1` select the workspace/Git root, while `ix map .` is the explicit subdirectory-scoped form.
- A command-level test pins the `invalid_map_path` JSON response.

`npm test`: 85 files, 1454 passed, 3 skipped; parser smoke passed.

Hiro-Chiba:
Thanks! I aligned the workspace root resolution and updated the PR body and tests. CI is passing.

KageBinary:
Reviewed, and I pushed one fix to the branch (`44a611b`) plus a merge of `main`.

## The defect I fixed

`resolveMapRoot` delegated the no-argument case to `resolveWorkspaceRoot`, whose cascade is:

1. explicit `--root` → 2. registered workspace containing cwd → 3. **named workspace** → 4. **default workspace** → 5. git root → 6. cwd

Steps 3 and 4 sit *ahead* of the git root. That ordering is right for `resolveWorkspaceRoot`'s actual question — "which graph am I querying?" — where a configured workspace outranking the current directory is the whole point.

`ix map` writes. With a default workspace configured, a bare `ix map` inside repo A re-ingested **repo B**, and rewrote B's map baseline on the way, with nothing on screen naming B. Reproduced on the branch before the change:

```
cwd    = /tmp/ix-repoA-c1gxTl/src   (a git repo)
mapped = /tmp/ix-repoB-DsmJPx      (the default workspace)
```

The PR's own tests asserted this as intended ("uses the default workspace before the current git root"), which is why CI was green. This is the open question from my 08-31 review — I've answered it in the branch rather than leaving it to block the merge, and the reasoning is in the commit message.

**New order** — the local answer wins whenever there is one:

1. explicit path argument
2. registered workspace containing cwd
3. **cwd's own git root** ← moved ahead of named/default
4. named/default workspace, for a cwd with no local context at all
5. cwd

Step 4 is why the fallback is still `resolveWorkspaceRoot`: `ix map` from a directory in no workspace and no repository keeps mapping the configured default, which is the useful answer there.

The git probe moved to an exported `gitRootFor`, keeping your `stdio` change — worth noting that change fixes a visible bug on its own. On main today, `ix text` outside a repo leaks `fatal: not a git repository` to the user's terminal; I hit it while testing #543.

## Tests

Two of yours asserted the old precedence by name and now assert the fix — both fail against the previous `resolveMapRoot` (mutation-checked, 2 failed / 7 passed). Two new ones pin what the reordering must *not* disturb: a registered workspace nested inside a repo still beats that repo's root, and a cwd with no local context still falls through to the default.

## The rest of the PR

The canonicalisation is sound and I have no changes to it. Two things I checked and am deliberately leaving:

- **`getOrCreateWorkspace` rewrites `root_path` in the user's config.yaml.** Both branches persist it (the migration branch already called `saveConfig`), so no half-written state. A user who registered a symlinked root will see one forced re-ingest as the workspace_id converges — correct, and the point of the change.
- **`canonicalWorkspacePath` per workspace per lookup** adds a `realpathSync.native` syscall per registered workspace. `selectWorkspaceForCwd` isn't called in a loop, so it's a handful of syscalls per command. Not worth complicating.

**Verified locally after merging main:** typecheck clean, ix-cli **1477 passed / 2 skipped**, eslint and knip clean.

Issue #539, opened 2026-08-29
[bug] ix locate exits 0 with a null result for a target that does not exist


KageBinary:
Found while unifying the error slugs in #538. Same failure mode as #526, in a command #526 did not cover.

## Description

`ix locate <missing>` reports the miss in every format but always exits **0**, and the JSON payload is shaped like a successful result:

```text
$ ix locate DefinitelyMissing --format json
{
  "resolvedTarget": null,
  "resolutionMode": "none",
  "systemPath": null,
  "diagnostics": ["No graph entity found."]
}
exit=0

$ ix locate DefinitelyMissing --format llm
error code=unresolved_target message="No graph entity found for \"DefinitelyMissing\"."
exit=0

$ ix locate DefinitelyMissing
No graph entity found for "DefinitelyMissing".
No matches found.
exit=0
```

This is #526 exactly — "machine callers cannot distinguish this semantic failure from a successful command that intentionally returned no records" — and `locate` is now the odd one out: `context`, `explain`, `read` (#532), `trace` and `subsystems` all exit 1 for the same condition. Note the `llm` output even emits an `error` record while exiting 0.

## Why this is not a one-line fix

Flipping the exit code would degrade at least two shipped plugins today, so it needs to land with them rather than before them.

`ix-opencode-plugin` and `ix-openclaw-plugin` both call `ix locate <target> --format json` through a `safeRun` helper built on Bun's `$`, whose `.text()` **throws on a non-zero exit**:

```ts
async function safeRun(cmd: ReturnType<typeof $>): Promise<string | null> {
  try { return await cmd.text(); } catch { return null; }
}
```

Today a miss returns the JSON body and the plugin reads `diagnostics` to say "No graph entity found — try `ix map` to refresh, or `ix locate` to check the exact name." With a non-zero exit, `safeRun` returns `null` and that specific guidance is lost. `ix-claude-plugin`'s `ix-bash.sh` hook also front-runs bash with `ix locate`.

## Suggested sequence

1. Teach the plugin `safeRun` helpers to keep stdout on a non-zero exit (they already tolerate a null result, so this is additive).
2. Then make `locate` exit 1 on an unresolved target, matching #526.

Doing 2 first is what breaks things; doing 1 first is harmless on its own.

## Environment

- Ix CLI: current main (post-#529–#536)
- Backend release: 1.0.28

KageBinary:
Worked the sequence in the order this issue lays out. All four PRs are green.

[...]

PR #547, opened 2026-08-30
fix(cli): fail unresolved graph commands

### LINKED ISSUE #546: [bug] graph commands exit 0 without JSON for unresolved targets
## Description

Several graph commands treat an unresolved target as a successful empty result. With `--format json`, they write no JSON to stdout and exit with status 0.

Affected commands are `overview`, `impact`, `contains`, `callers`, `callees`, `imports`, `imported-by`, `depends`, `trace`, and `history`. `locate` has the same exit-status problem but is already tracked separately in #539 because its plugin consumers require coordinated changes.

## Steps to reproduce

1. Run `ix overview DefinitelyMissing --format json > output.json`.
2. Inspect `output.json` and the process exit status.
3. Repeat with the other commands listed above.

## Expected behavior

Each command emits a JSON `unresolved_target` error and exits non-zero, consistently with `context`, `explain`, and `read`.

## Actual behavior

Stdout is empty and the process exits with status 0.

## CLI output

```text
$ ix overview DefinitelyMissing --format json > output.json
$ echo $?
0
$ wc -c < output.json
0
```

## Environment

- OS: macOS
- Node: 22.x

Hiro-Chiba:
Fixes #546

## Summary
Make unresolved graph targets a structured command failure instead of an empty successful result, while keeping ambiguous targets distinguishable and actionable.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [x] Docs
- [x] Test
- [ ] CI

## Changes
- Route target resolution through one result-aware helper for `context`, `explain`, `overview`, `impact`, `contains`, `callers`, `callees`, `imports`, `imported-by`, `depends`, `trace`, `history`, and `diff`.
- Emit `unresolved_target` JSON and LLM records and set exit status 1 only when no target matched.
- Emit `ambiguous_target` with candidates when several targets match. An out-of-range `--pick` remains accurately classified and includes a `pick_out_of_range` diagnostic.
- Exit non-zero when ambiguity prevents the requested graph operation from choosing a target, including an out-of-range `--pick`.
- Report both missing endpoints from `trace --to` in one machine-readable result.
- Preserve structured LLM or JSON error output in the built-in MCP server when the CLI exits non-zero, even when stderr also contains resolver guidance.
- Cover all 14 command names in JSON, LLM, and text formats without writing stitch-scope cache files into the real Ix home.
- Include the `diff` exit-status change that #566 explicitly left queued behind this compatibility work.

## Compatibility and merge order
The in-repo MCP blocker is fixed in this branch. External consumers still need to retain stdout from non-zero CLI runs before this PR merges.

Ready consumer changes:
- `ix-cursor-plugin#26` is merged.
- `ix-gemini-plugin#29` is merged.
- `ix-opencode-plugin#21` is merged and covers every OpenCode tool path.

Remaining gates:
- `ix-claude-plugin#38` is open and covers the affected `impact`, `overview`, and `inventory` hook paths. `ix-claude-plugin#37` is the separate `locate` compatibility change for #559.
- `ix-openclaw-plugin#33` is open, but its current change does not cover the `hooks/ix-read` path that discards stdout after a non-zero `overview` or `impact` run. That path still needs coverage before this PR is safe to merge.

`ix locate` remains outside this PR and is tracked by #539 and draft PR #559.

## Output behavior
- Command-specific miss messages are unified as `No entity found matching`.
- A genuine miss exits 1 and reports `unresolved_target`.
- Ambiguity exits 1 and reports `ambiguous_target` with candidates because the requested graph operation could not choose a target.
- An out-of-range `--pick` exits 1 while retaining the ambiguity candidates and adding `pick_out_of_range`.
- Text mode keeps stdout empty on resolution failure and writes guidance to stderr.

## Validation
- `npm test` (92 files, 1690 passed, 3 skipped; parser smoke passed)
- `npm run typecheck`
- `npm run lint` (0 errors, 41 warnings)
- Targeted MCP and resolution contract tests (78 passed)
- Plugin compatibility status rechecked across Cursor, Gemini, OpenCode, OpenClaw, and Claude

## Checklist
- [x] Tests pass
- [x] Smoke tests pass


KageBinary:
## Review — good consolidation, one wrong-code case to decide

Replacing ten hand-rolled llm-only handlers with `reportUnresolvedTarget` is the right move: it deletes two copies of `llmUnresolved` (callers.ts, imports.ts) and eight inline duplicates, and it gives `--format json` an actual payload where it previously got **empty stdout and exit 0**. The parameterised test covering all ten commands is the right way to keep them aligned.

Confirmed the text path is unchanged: `reportUnresolvedTarget` prints nothing for `text`, and the resolver has already written its human guidance to stderr, so the only difference for an interactive user is the exit code — which is the point.

### The one thing worth deciding first

`resolveFileOrEntity` returns `null` for **two** conditions:

```ts
if (result.ambiguous) {
  printAmbiguous(symbol, result.result, opts);   // stderr: 'Ambiguous symbol "X": ...'
}
return null;
```

So an ambiguous symbol now emits, on stdout:

```json
{ "error": "unresolved_target", "message": "No entity found matching \"X\"." }
```

while stderr says the opposite — it resolved to several candidates and the user should pass `--pick`. A machine consumer is told nothing was found when in fact too much was.

This is **not new to your PR** — `context`, `explain` and `read` have done it since they adopted the helper, and you are propagating an existing behaviour rather than inventing one. But this PR takes it from 3 commands to 13, which is the point at which it stops being an oddity.

`resolve.ts` already has the right slug: `diagnostics: [{ code: "ambiguous_resolution", ... }]`. And #559 in this same stack draws exactly this distinction in prose — *"Ambiguity is a different condition … it resolved to several candidates, which is an answer, not a miss."* Two PRs open at once disagreeing about it is worth resolving now.

Two ways to go, both fine by me:

1. Merge as-is and fix `ambiguous_resolution` across all 13 in a follow-up — the exit code is right either way, only the slug is wrong.
2. Have `resolveFileOrEntity` signal which condition it hit, and give `reportUnresolvedTarget` a sibling. Bigger, but it is the actual fix.

### Smaller notes

- **The message text changes**, from `No entity resolved for "X".` to `No entity found matching "X".` (and `trace`'s `Could not resolve X.` to the same). Anything grepping the old strings breaks. Uniformity is worth it; just worth knowing it is a string change and not only a routing change.
- `callees` and `imported-by` are registered from `callers.ts` / `imports.ts`, so the test list has the same loader twice under two names. That is correct, and not obvious — a one-line comment would save the next reader the double-take.

KageBinary:
## Blocking, on evidence — this has #559's problem, at ten times the surface

I was clearing the batch and went to merge this. It should not merge yet, and the reason is the one that already made **#559** a draft.

**#559 is held back because making `ix locate` exit 1 degrades shipped plugins until they are updated.** This PR does the same thing to ten commands, and the plugins call them far more than they call `locate`.

### The plugins discard stdout on a non-zero exit

Counted across the six plugin repos, in code — not prompt prose:

| call site | shape |
|---|---|
| `ix-claude-plugin/hooks/ix-pre-edit.sh:54` | `RAW=$(ix impact "$REL_PATH" --format json …) \|\| { … ix_log "FAILED ix impact exit=$_exit" … }` |
| `ix-claude-plugin/hooks/ix-read.sh:99–103` | `wait $_OV_PID \|\| ix_capture_async … "overview failed"` |
| `ix-openclaw-plugin/hooks/ix-read/handler.ts:49,53` | `runIx(["overview", …]).catch(…)`, `runIx(["impact", …]).catch(…)` |
| `ix-opencode-plugin/tools/ix-impact.ts:49` | `await $\`ix impact …\`` in `try/catch` → `return unavailable(…)` |
| `ix-opencode-plugin/tools/ix-neighbors.ts:124` | `await $\`ix depends …\`` in `try/catch` |
| `ix-opencode-plugin/tools/ix-decide.ts:123` | `await $\`ix impact …\`` in `try/catch` |

Bun's `$` throws on non-zero; the shell hooks run under `set -euo pipefail` and branch on `||`. So after this PR, an unresolved target on `ix impact` — the most-called of these — turns into `FAILED ix impact exit=1` in the hook log, and **the structured `{error:"unresolved_target"}` body this PR just added is the thing that gets thrown away**. The user sees a generic "unavailable" where they used to see an empty-but-successful result. `ix-pre-edit` fires before every edit.

### And the MCP layer does the same, worse

`mcp/server.ts`'s `runCommand` prefers **stderr** over stdout when the exit is non-zero:

```ts
if (!result.ok) {
  const detail = result.stderr.trim() || result.stdout.trim() || `… failed without output`;
  return textResult(JSON.stringify({ error: detail, tool }), true);
}
```

`reportUnresolvedTarget` writes its record to **stdout** and nothing to stderr — but the resolver has already printed human guidance to stderr by then. So `stderr.trim()` wins, and `ix_impact` / `ix_overview` / `ix_callers` / `ix_trace` return `{"error": "<resolver prose>"}` instead of the clean `error code=unresolved_target message="…"` this PR emits. The structured record is built and then discarded, for all ten.

(`ix_context`, `ix_explain` and `ix_read` already behave this way, so this is not a regression *in kind* — but going from 3 commands to 13 is what makes it worth stopping for.)

### What I would do

Put this in the same queue as #559 — **after** ix-opencode-plugin #20, ix-openclaw-plugin #33 and ix-claude-plugin #37, which are the PRs that teach the plugins to read a payload from a failed run. The change itself is right and I am not asking for edits to it.

Worth noting the two PRs also still disagree on the ambiguity slug (my earlier comment, and the comment in #559's diff). If the plugin work is landing anyway, fixing `ambiguous_resolution` in `reportUnresolvedTarget` before this merges would fix it once for all thirteen commands rather than thirteen times later.

### Merged from this batch, for context

#549, #553, #555, #556, #558, #561, #562, #563, #564. This one, #543, #545, #551 and #559 stayed open — reasons on each.

KageBinary:
Correction to my comment above: I listed three plugin repos with runners. Running the check systematically, it is **five of the six**.

`ix impact` alone, excluding `.md` and agent prompt strings:

```
ix-cursor-plugin/mcp/hooks/pre-edit.ts        runIx(["impact", …
ix-cursor-plugin/mcp/dist/hooks/pre-edit.js   runIx(["impact", …
ix-gemini-plugin/mcp/tools/ix_impact.ts       runIx(["impact", …
ix-gemini-plugin/mcp/tools/ix_query.ts        runIx(["impact", …
ix-claude-plugin/hooks/ix-pre-edit.sh         $(ix impact …
ix-opencode-plugin/tools/ix-impact.ts         $`ix impact …
ix-opencode-plugin/tools/ix-docs-tool.ts      $`ix impact …
ix-opencode-plugin/tools/ix-decide.ts         $`ix impact …
```

plus `ix-openclaw-plugin/hooks/ix-read/handler.ts` for `overview` and `impact`. So `ix-cursor-plugin` and `ix-gemini-plugin` are affected as well, and the three step-1 plugin PRs (#20 opencode, #33 openclaw, #37 claude) do not cover them.

That does not change the conclusion, it widens it: the plugin-side work this needs is five repos, not three.

The check itself is now written down in **#565**, along with why it is invisible from inside this repo.

Hiro-Chiba:
I left the code unchanged, following the review's request that this CLI change itself does not need edits. I re-audited the current heads and call sites in all six shipped plugin repositories and recorded the merge blocker in the PR body.

At present, `ix-cursor-plugin#26` and `ix-gemini-plugin#29` cover their affected fallback paths. `ix-opencode-plugin#20`, `ix-openclaw-plugin#33`, and `ix-claude-plugin#37` are primarily locate/docs fixes; some `impact`, `depends`, and `overview` call sites expanded by #547 still discard stdout from a non-zero run. Keeping #547 open is therefore still the correct state.

The body also records the existing shared-helper behavior where ambiguity is reported as `unresolved_target`. CLI `npm test`: 84 files, 1456 passed, 3 skipped; parser smoke passed.

Hiro-Chiba:
Thanks! I left the CLI code unchanged and only documented the plugin compatibility findings. I’ll leave the plugin-side follow-ups with you.

KageBinary:
Re-ran the plugin exit-code check today (2026-09-01). **Still blocked, and the blocker has not moved.**

The change itself is right — `reportUnresolvedTarget` is the correct shape and these ten commands should join the contract. The problem is entirely one of ordering.

`reportUnresolvedTarget` sets `process.exitCode = 1`. Five of the six plugin repos discard stdout when ix exits non-zero, so the structured `{error: "unresolved_target"}` record this PR *adds* is exactly what gets thrown away. Re-verified line by line against each repo's current `main`:

| repo | site | what happens on a miss |
|---|---|---|
| ix-claude-plugin | `hooks/ix-pre-edit.sh:54` | `RAW=$(ix impact …) \|\| { … exit 0 }` — stdout dropped, hook bails |
| ix-openclaw-plugin | `hooks/ix-read/handler.ts:49,53` | `.catch(() => "")` — stdout dropped |
| ix-opencode-plugin | `tools/ix-impact.ts:49` | bun `$` throws → `unavailable(...)` |
| ix-cursor-plugin | `mcp/hooks/*` | gated on `result.ok` |
| ix-gemini-plugin | 6 MCP tools | reports the runtime's `IX_UPSTREAM_UNAVAILABLE`, not the CLI's record |

The reason this matters more than it looks: `ix impact <file>` on the file being edited is the highest-traffic ix call in the fleet, and an unresolved target there is a **normal, frequent** outcome — a new file, a file not yet ingested. This PR turns that routine miss into "ix unavailable" everywhere.

The five PRs that fix it are all open and all still `BLOCKED`: ix-opencode-plugin #20, ix-openclaw-plugin #33, ix-claude-plugin #37, ix-gemini-plugin #29, ix-cursor-plugin #26. Every one is green with the ruleset rules satisfied; `PUT /pulls/N/merge` still 405s because `bypass_actors` is empty. That is a permissions fix on the plugin repos, not something that can be resolved from here or from this PR.

**Nothing to change on this branch** — I'm not pushing a commit, because the fix isn't on this side. It merges the moment those five land. Also still open from 08-31: this PR and #559 disagree about whether an *ambiguous* target should report `unresolved_target`; worth settling before both go in.

Rule this came from: `CONTRIBUTING.md` → CLI Standards → Exit codes (#565).

KageBinary:
Second pass, this time on the code rather than just the plugin-compatibility question. **Re-verified against today's `main`** (`793c56b`, after #543/#545/#551 landed): merges with no conflicts, typecheck clean, **ix-cli 1515 passed / 2 skipped**.

## Correction to what I said on 08-31

I flagged that this PR and #559 disagree about whether an **ambiguous** target should report `unresolved_target`. **They don't — I was wrong, and there is nothing to settle.**

All ten commands here resolve through `resolveFileOrEntity`, which never returns null for ambiguity. `resolve.ts:730` onward sorts the matches, applies a path-based tie-break, and returns `matches[0]`; the only `null` returns are a failed raw-UUID lookup, a `Class::method` phase-B miss, and `matches.length === 0`. So `!target` in these ten commands means *nothing matched*, full stop, and exiting 1 is unambiguously the right reading.

Ambiguity is only reachable through `resolveWithAmbiguity`, which is used by exactly one command — `locate` — and has its own `ambiguous_target` slug that exits 0. #559 leaves that path alone deliberately. So both PRs already agree: **ambiguity is an answer, zero matches is a failure.**

## Test quality — good

The suite isn't stubbing the thing under test. `vi.mock` replaces `IxClient` with `search() => []`, so the real `resolveFileOrEntity` runs and genuinely fails to resolve. Mutation-checked by reverting `overview.ts` to the old inline pattern: **1 failed / 12 passed**, and it failed on `ix overview` specifically.

## One gap: `ix diff` is left behind

The PR converts ten commands and `#559` takes `locate`, which accounts for eleven of the twelve commands that resolve a target. **`diff.ts` is the twelfth and isn't touched**, so after this lands it is the only command still emitting the old `No entity resolved for "…"` wording while exiting 0.

Worse, it has *two* handlers that disagree with each other and with the rest of the CLI:

- `diff.ts:520` and `:526` (the `--format json` branch) emit `{"error": "No entity found matching \"X\"."}` — **the human sentence in the `error` field**, no `message`, no slug. Every other command emits `{error: "unresolved_target", message: "…"}`. A consumer routing on `error === "unresolved_target"` gets an English sentence instead.
- `diff.ts:527` (the text/llm branch) is the old inline llm-only record.

The slug half is a pre-existing bug on `main` — `docs/llm-format.md:29` says a target that does not exist is *always* `unresolved_target`, and #538 unified everything except this file. It is **separable from the exit code**, so it can be fixed now without widening this PR's plugin blast radius. I'd suggest doing that separately and leaving `diff`'s exit code to join this PR's queue, so `diff` doesn't become a lone straggler.

Worth knowing for the eventual exit-code change: `ix-cursor-plugin/mcp/tools/map.ts:198` consumes `ix diff --format json` and, on an unresolved target today, reads `raw.summary`/`raw.total` off the error object and reports a **successful "0 changes"** — a miss rendered as an empty answer. Only the exit code fixes that, which is the same reason this PR is queued behind the plugin work.

**Still blocked on the same thing**, unchanged: opencode #20, openclaw #33, claude #37, gemini #29, cursor #26.

KageBinary (commented):
Re-reviewed against current `main`. The CLI change itself is mechanically sound — all 12 substitutions are faithful, no call site is missed among the commands the PR claims to cover, and `history.ts` correctly passes the positional `target` string rather than the resolved entity. The problems are scope and consistency, and one of them is a blocker the PR body does not mention.

## Blocker: this breaks Ix's own MCP server

The PR body names six external plugin repos as the merge gate. There is a seventh consumer in this repo, and no linked PR fixes it.

`mcp/server.ts:625`:

```ts
const detail =
  result.stderr.trim() ||
  result.stdout.trim() ||
  `${args.slice(0, 2).join(" ")} failed without output`;
```

On a non-zero exit, **stderr wins and stdout is discarded**. And stderr is never empty for an unresolved target — `resolve.ts:251` and `:308` both call `stderr(\`No entity found matching "${symbol}".\`)` unconditionally.

So for the nine MCP tools that route through `runFormatted` → `runCommand` (`ix_impact`, `ix_overview`, `ix_callers`, `ix_callees`, `ix_imports`, `ix_imported_by`, `ix_depends`, `ix_trace`, `ix_history`):

- **before**: exit 0, `runFormatted`'s `semanticError` branch fires, the agent receives `error code=unresolved_target …` — the record is preserved
- **after**: exit 1 → `!result.ok` → the agent receives `{"error":"No entity found matching \"X\".","tool":"ix_callers"}` — the structured record this PR adds is thrown away

`mcp.test.ts:115` is titled *"marks a structured LLM error as an MCP error when the CLI exits successfully"*, and its fixture is `ok: true` with the record on stdout and `No entity found matching "Missing".` on stderr. That test encodes precisely the contract this PR flips.

Fixing `runCommand` to prefer stdout when it parses as an llm/json error record would resolve it, and would be worth doing in this PR rather than after.

## Ambiguity is reported as "not found"

`resolveFileOrEntity` returns `null` for three different conditions, and all three now produce the same record:

1. **Ambiguity** — `resolveEntity` prints the candidate list to stderr, then returns null. `ix impact parse --format json` reports `No entity found matching "parse"` even though several `parse` entities exist.
2. **`--pick` out of range** (`resolve.ts:425-427`) — reported as a missing symbol rather than an out-of-range pick.
3. A genuine miss.

The resolver already distinguishes these: `ResolveResult.ambiguous` exists, `AmbiguousResult.diagnostics` already carries `ambiguous_resolution`, and `locate.ts` already ships a distinct slug with a `candidates` field. This is filed under "Known behavior" in the body, but it is a wrong answer emitted on stdout as machine-readable truth across 12 new call sites — an agent told "not found" gives up, where "ambiguous" would make it disambiguate.

## `diff.ts` is left on the old contract

`ix diff <rev> <rev> <target>` still exits 0 and still emits the removed wording:

- `ix callers Missing --format llm` → `error code=unresolved_target message="No entity found matching \"Missing\"."`, exit 1
- `ix diff 1 5 Missing --format llm` → `... message="No entity resolved for \"Missing\"."`, exit 0

Its json branches (`diff.ts:514,520`) emit a third shape — message-in-`error` rather than code-in-`error` plus `message` — so a consumer switching on `parsed.error === "unresolved_target"` misreads it. `diff.ts` is the only remaining in-repo `resolveFileOrEntity` call site with a bespoke handler.

## `locate` and the exit-code rule

`docs/llm-format.md:27-28` says an error line is emitted and *"the process still exits non-zero"*. `renderLocateLlm` (`locate.ts:222`) returns `llmError("unknown_target", ...)` while `locate` exits 0. After this PR the inconsistency gets sharper, since `ix locate → ix callers` is the chain SKILL.md recommends and the two halves would disagree on exit status for the same missing symbol.

Note this is a *different slug* — `unknown_target`, not `unresolved_target` — which is itself worth reconciling if the goal is one uniform contract.

## Test coverage is on the wrong format

All 13 parameterized cases in `unresolved-machine-output.test.ts` pass `["--format", "json"]`. But `llm` is the format whose wording actually changed for these ten commands, and the one every plugin parses. A regression in `llmError` quoting, or a revert to the old wording, would ship green. Nothing asserts `--format text` exits 1 with empty stdout either, which is the other half of the fix.

## Altitude

The fix is applied at 12 call sites rather than in `resolveFileOrEntity`, which is the one function that knows the target failed and already owns the human stderr line. There is no type, lint rule, or test enforcing that a caller remembers — and `diff.ts` is the proof the pattern does not hold, having been missed by this very PR. A `resolveOrReport(...)` wrapper would cover `diff`, `locate`, and every future command for free, and because `ResolveResult.ambiguous` is in scope there, it would fix the ambiguity problem at the same depth instead of deferring it.

## Minor

`trace --to` names only the `from` endpoint when both fail, so a caller fixes one and immediately hits the other — two round trips for one report. Both names are in scope at `trace.ts:466`.

`unresolved-machine-output.test.ts` mocks `../../client/api.js` but not `../config.js`, so `ensureReadScope` reaches the real registry and `writeStitchScope(ws, null)` writes `~/.ix/stitch_scope_<key>.json`. On a machine where Ix is stitched into a System, running the suite caches `null` as authoritative until a map or ingest clears it. Pre-existing for the three original cases, but this PR expands the file.

---

Net: the direction is right and the mechanical work is clean. I would hold the merge on the MCP regression, and would rather see the ambiguity split and the `resolveFileOrEntity` consolidation land here than as follow-ups, since both get cheaper at that altitude.

Alot1z:
Coordination note — this overlaps #559 on the same surface, and the two should be sequenced deliberately.

Both touch `ix-cli/src/cli/__tests__/unresolved-machine-output.test.ts` (the only shared file; whichever merges second needs a rebase there), and both flip the same behavior class: a non-zero exit on an unresolved target where the command previously exited 0 with a null result (#546 here, #539 there).

#559 is a draft on purpose and names the reason: its exit-code flip is a breaking change for machine consumers, and it is sequenced as step 2 after three plugin PRs land (ix-opencode-plugin#20, ix-openclaw-plugin#33, ix-claude-plugin#37) precisely so those consumers are ready for the non-zero before it ships. This PR's 8-command flip is the same breaking-change class — if it merges ahead of those consumers, the degradation #559 is carefully sequencing around can arrive through a different door.

Suggestions: reference #559 in the body so reviewers see the overlap; rebase the shared test file against whichever lands first; and if any of the three plugin consumers call these graph commands, land this PR in the same order #559 uses (plugins first).

KageBinary:
Agreed on the sequencing, and worth adding: sequencing is not the only thing
outstanding here. The blocker from my earlier review is still live at `d112b78`,
and it is not mentioned in the PR body.

**This PR breaks Ix's own MCP server.** `runCommand` in
`ix-cli/src/mcp/server.ts:618` branches on the exit code:

```ts
const result = await runIx(args, timeoutMs);
if (!result.ok) {
  const detail = result.stderr.trim() || result.stdout.trim() || …;
  return textResult(JSON.stringify({ error: detail, tool }), true);
}
return textResult(result.stdout.trim() || "{}");
```

The eight commands this PR flips (`callers`, `contains`, `depends`, `history`,
`impact`, `imports`, `overview`, `trace`) still print a well-formed
"unresolved" JSON record on stdout — that is the whole point of the change —
but they now exit non-zero, so every one of them takes the `!result.ok` branch.
`detail` prefers `stderr`, so the structured record on stdout is discarded, and
the MCP client gets `{ error, tool }` with `isError: true` instead of the answer
the CLI just produced.

So the same consumer class #559 is sequencing around includes a consumer that
ships **in this repo**, and it is not covered by landing the three plugin PRs
first. It needs `runCommand` taught that a machine-format record on stdout is
an answer, not a failure — either before this merges or in it.

Everything else in my earlier review still stands as written.

Hiro-Chiba:
Thanks. I addressed the latest review by preserving structured stdout in the built-in MCP server, distinguishing ambiguity and out-of-range picks, bringing `diff` onto the same exit contract, reporting both missing `trace --to` endpoints, isolating the tests, and covering JSON/LLM/text. `npm test` now passes with 1690 tests and 3 skipped; typecheck and lint also pass. I refreshed the PR body and left #559 and the plugin-owned branches untouched. Claude #38 and the uncovered OpenClaw hook path remain the external merge gates.

KageBinary:
The MCP blocker is genuinely closed. Verified on `4768197`, not read.

## `runCommand` now keeps the structured record

```ts
const machineError = machineErrorOutput(result.stdout);
if (machineError) return textResult(machineError, true);
```

Both shapes it sniffs are the ones the CLI actually emits — I built the branch and sampled them rather than assuming:

```
llm : error code=ambiguous_target message="Ambiguous symbol \"foo\"." candidates=1:foo,2:foo
json: { "error": "ambiguous_target", "message": ..., "candidates": [...] }
```

`startsWith("error code=")` and the `error`-is-a-string check cover both, and the MCP server always passes `--format=llm` or `--format=json` (`toArgv`), so the text-format gap does not arise. Mutation-checked — deleting the two lines:

```
× preserves a structured LLM error when the CLI exits non-zero
× preserves a structured JSON error when the CLI exits non-zero
× preserves structured CLI errors written to stdout
  Tests  3 failed | 22 passed (25)
```

Suite on this head: **1690 passed / 2 skipped**, typecheck clean. The one failure I saw is `watch-dedup`'s `check-ignore` assertion, which is my worktree's symlinked `node_modules` (Ix#585, fixed by #592) and not yours.

## One deliberate decision I would push back on

`reportAmbiguousTarget` prints a machine-readable error record to **stdout** and leaves the exit code at 0, while `reportUnresolvedTarget` — the other arm of the same `reportResolutionFailure` helper this PR introduces — sets it to 1. Measured directly against the built branch:

```
format=json   process.exitCode=0   {"error": "ambiguous_target", ...}
format=llm    process.exitCode=0   error code=ambiguous_target message=...
format=text   process.exitCode=0   (stdout empty)
```

`unresolved-machine-output.test.ts:115` and `:137` pin `expect(process.exitCode).toBeUndefined()`, so this is a choice rather than an oversight. Two reasons I think it is the wrong one:

**It is the shape of #546.** That issue is "exits 0 without JSON for unresolved targets". This PR fixes the missing-JSON half for ambiguity — it adds the record, which `main` never printed at all — and leaves the exit-0 half. A consumer running `ix impact Foo --format json || handle` gets a success status and an object whose only key is `error`. The JSON contract is that an `error` key means the command did not answer; the exit code should not disagree with the payload printed alongside it.

**The out-of-range `--pick` case is harder to defend.** `--pick 3` against two candidates is not "the user needs to disambiguate", it is an argument that cannot be satisfied — and `#551` just landed on `main` making invalid option *values* exit non-zero (`error: option '--format <fmt>' argument 'yaml' is invalid`). `--pick 3 is out of range (1-2)` exiting 0 sits directly against a contract merged three days ago. Classifying it as ambiguity for the *message* is right; carrying exit 0 along with it is what I would separate.

If ambiguity is meant to stay 0 as a "needs more input" state, that is arguable — but then it deserves a sentence in the PR body and in CLI Standards saying so explicitly, because it is a second exit-code contract landing in the PR that defines the first one. At minimum I would flip the out-of-range pick.

## On sequencing

The MCP consumer was the part not covered by "land the three plugin PRs first", and it is now handled in-repo, so that specific objection is gone. The `#559` ordering question is still Joey's call.

Nothing else open from me.

Hiro-Chiba:
Thanks. Agreed. I updated `reportResolutionFailure` so an ambiguous graph operation, including an out-of-range `--pick`, exits 1 while preserving the `ambiguous_target` candidates and `pick_out_of_range` diagnostic. `locate` remains unchanged under #559. The docs and PR body now state this contract, and all 1690 tests still pass.

KageBinary:
The ambiguity fix is right — `5280ec6` puts `process.exitCode = 1` in `reportResolutionFailure` so both arms agree, and the two tests now pin `toBe(1)` where they pinned `toBeUndefined()`. That closes my finding, including the out-of-range `--pick` case, which was the one I thought hardest to defend at exit 0.

**I have merged every other open PR on this repo today and deliberately left this one open.** Not because of anything in the code — because your own merge gate is still open upstream, and I checked rather than assumed:

| consumer | state |
|---|---|
| `ix-opencode-plugin` #20, #21 | **MERGED** |
| `ix-openclaw-plugin` #33 | **OPEN** |
| `ix-claude-plugin` #37, #38 | **OPEN** |

Two of the three still treat a non-zero exit as a failure and discard the body. Landing this now would ship the exit-code flip for eight graph commands ahead of them — which is the same shape as the bug I reported inside this PR (the in-repo MCP server discarding a well-formed record because `!result.ok`), just in consumers we do not control from here. You fixed the in-repo one; the external ones are what is left.

So this is ready on the merits and waiting on those three PRs, exactly as you and @Alot1z sequenced it. Once `ix-openclaw-plugin#33` and `ix-claude-plugin#37`/`#38` land, this should go straight in — I do not have any open findings against it.

One residual, small and out of scope for this PR, recorded so it is not lost: `resolveFileOrEntity` (`resolve.ts:692`) still calls `reportAmbiguousTarget` directly with a hardcoded `"text"` and sets no exit code, so `ix locate` keeps exiting 0 on an ambiguous target. That is #559's surface rather than this one's, and #559 is a draft behind the same three plugin PRs — worth folding into it there.

## For context on what changed underneath you

`main` moved a long way today: #570, #577, #578, #582, #583, #584, #587, #589, #590, #591, #592, #593–#596, #597, #598, #601 and #602 all landed. This branch is still `MERGEABLE`, but expect to want a rebase before it goes in — `ci-success` now requires six jobs (`api-parity`, `doc-parity`, `links` and two harness smoke jobs joined the original three), and `docs/` gained a committed flag reference plus a dead-link gate that will check any URL you add.

KageBinary:
**The consumer gate is closed — merging this.**

I landed the remaining plugin PRs today, reviewing each rather than merging on green:

| consumer | PR | state |
|---|---|---|
| `ix-opencode-plugin` | #20, #21 | already MERGED |
| `ix-claude-plugin` | #37 (`ix locate`), **#38** (`ix impact`/`inventory`/`read`) | **MERGED** |
| `ix-openclaw-plugin` | **#33** (`safeRun` keeps stdout) | **MERGED** |

`#38` was the one that actually mattered for this PR: `#37` only covered `ix locate` inside `ix_run_text_locate`, while the hooks calling `ix impact`, `ix overview` and `ix inventory` **directly** — the calls this PR changes — still did `RAW=$(ix impact …) || { …; exit 0; }` and discarded the body. `ix impact` on the file being edited is the most frequent ix call that plugin makes, and a miss there is routine.

Each was mutation-checked on its merged tree, not taken on the PR's word:

- `ix-claude-plugin#38` — reverting `hooks/` and keeping the tests turns exactly two red (`pre-edit/non-zero exit with a body still warns`, `pre-edit/logs a miss rather than a failure`); 91 passed / 0 failed with the fix.
- `ix-openclaw-plugin#33` — reverting only `safeRun`'s catch while keeping `IxCommandError` fails exactly `ix_docs keeps the JSON body when ix exits non-zero`, and nothing else; 13/13 with the fix.

I also closed `ix-claude-plugin#35` rather than merging it: `git diff main <head>` had become deletions-only, so merging it would have reverted #36 and #37.

## Verified against today's `main`, not the CI that ran days ago

`main` moved a long way since this branch was last built — six required jobs now, and three committed docs gates that did not exist when this PR's checks last ran. Merged `main` in locally and ran everything:

```
test:unit          1749 passed | 2 skipped   (95 files)
typecheck          clean
lint               0 errors (41 pre-existing warnings)
check-api-parity   42 documented, 42 called, 13 schemas, 13 refs
check-doc-parity   54 commands, 196 flags
check-links        87 references, 0 broken
```

## One residual, out of scope here

`resolveFileOrEntity` (`resolve.ts:692`) still calls `reportAmbiguousTarget` directly with a hardcoded `"text"` and sets no exit code, so `ix locate` keeps exiting 0 on an ambiguous target. That is #559's surface, and I am landing it there rather than widening this PR.

Thanks for the rounds on this one — the MCP fix in particular was the right shape.

PR #549, opened 2026-08-30
fix(doctor): exit non-zero when checks fail

### LINKED ISSUE #548: [bug] ix doctor exits 0 when health checks fail
## Description

`ix doctor` can report `healthy: false` while exiting with status 0. This also occurs when the backend is unavailable or a completed-map check fails, so automation cannot rely on the process status.

## Steps to reproduce

1. Stop the Ix backend or use a workspace without a completed map baseline.
2. Run `ix doctor --format json`.
3. Inspect the JSON result and process exit status.

## Expected behavior

If any non-warning check fails, doctor reports an unhealthy result and exits non-zero. Warning-only runs remain successful.

## Actual behavior

The JSON result contains `"healthy": false`, but the process exits with status 0.

## CLI output

```text
healthy=false
exit=0
```

## Environment

- OS: macOS
- Node: 22.x

Hiro-Chiba:
Fixes #548

## Summary
Set a non-zero process status whenever `ix doctor` has a failing health check.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Set exit status 1 when any non-warning doctor check fails.
- Keep warning-only doctor runs successful.
- Verify the exit status for incomplete maps, unmatched workspaces, and warning-only cloud workspaces.

## Validation
- `npm test` (84 files, 1446 passed, 3 skipped)
- `npm run typecheck`

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -265,7 +265,13 @@ jobs:
           # The converse, and the regression guard for #518: in a directory no
           # workspace covers, doctor must say so instead of passing every check
           # against whichever workspace happens to be the default.
-          echo "::group::doctor (unmapped cwd)"; (cd "$RUNNER_TEMP" && $IX doctor --format json) | tee doctor-unmapped.json; echo "::endgroup::"
+          echo "::group::doctor (unmapped cwd)"
+          set +e
+          (cd "$RUNNER_TEMP" && $IX doctor --format json) | tee doctor-unmapped.json
+          doctor_exit=${PIPESTATUS[0]}
+          set -e
+          if [ "$doctor_exit" -ne 1 ]; then echo "expected doctor to exit 1 for an unmapped cwd, got $doctor_exit"; exit 1; fi
+          echo "::endgroup::"
           node -e "const d=require('./doctor-unmapped.json');const c=(d.checks||[]).find(c=>c.name==='Workspace for this directory');if(!c){console.error('no workspace check in doctor output');process.exit(1)}if(c.ok||d.healthy){console.error('doctor reported healthy in a directory with no workspace');process.exit(1)}console.log('unmapped cwd reported:',c.detail)"
 
       - name: Backend logs on failure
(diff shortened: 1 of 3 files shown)


KageBinary:
## Review — clean, smallest correct change. No objections.

One line of behaviour, guarded three ways:

```ts
if (hasFailure) process.exitCode = 1;
```

`hasFailure` is already computed and already the thing `healthy` is derived from, so the exit code and the payload cannot drift. Warnings deliberately stay at 0, which matches the issue and matches the existing `warn` semantics — verified by the cloud-ingested case in the tests, which stays `healthy=true` / `exitCode` undefined.

Checked the blast radius before agreeing this is safe:

- **Not exposed over MCP.** No `ix_doctor` tool, so no agent-visible behaviour changes.
- **Not in the installers.** `install.sh`, `install.ps1` and both `skills/ix/scripts/bootstrap.*` reference `ix doctor` only in guidance text — nothing invokes it under `set -e`. A fresh install with no map baseline would otherwise have started failing.
- **The one CI caller you patched** is the only one that expects failure. The `PIPESTATUS[0]` handling is right: `set +e` suspends errexit but leaves `pipefail` on, so the pipeline still reports doctor's status and `${PIPESTATUS[0]}` reads it rather than `tee`'s 0.

Tests save and restore `process.exitCode` in `beforeEach`/`afterEach`, which matters in a shared worker — without it a later test inherits the 1.

### One thing to be aware of, not a change request

`ci.yml:263` — the *healthy* doctor call — is unguarded and the step runs `set -euo pipefail`:

```bash
echo "::group::doctor"; (cd "$FIXTURE" && $IX doctor --format json) | tee doctor.json
node -e "...if(!d.healthy){console.error('doctor not healthy');process.exit(1)}"
```

That assertion on the next line now cannot be reached: if doctor is ever unhealthy the step dies at the pipeline, and the log shows a bare non-zero exit instead of the check that failed. Failing is correct; losing the diagnostic is the cost. Same `set +e` / `PIPESTATUS` treatment there would keep the explanation, and costs four lines.

PR #553, opened 2026-08-30
fix(read): report invalid and denied reads

### LINKED ISSUE #552: [bug] ix read exits 0 without machine errors for denied paths and invalid ranges
## Description

`ix read` refuses a path outside the workspace, but in JSON mode it emits no JSON and exits with status 0. Invalid numeric line ranges such as `0-0` or a reversed range are also returned as successful empty reads.

## Steps to reproduce

1. Run `ix read /etc/hosts --root /tmp/ix-fixture --format json`.
2. Run `ix read src/example.ts:0-0 --root /tmp/ix-fixture --format json`.
3. Run `ix read src/example.ts:5-2 --root /tmp/ix-fixture --format json`.

## Expected behavior

Denied paths emit a structured `path_outside_workspace` error. Invalid ranges emit `invalid_line_range`. Both cases exit non-zero and never emit source content.

## Actual behavior

The denied read writes no JSON and exits 0. Invalid ranges return an empty content payload and exit 0.

## CLI output

The first command has empty stdout. The range commands return a successful JSON result with empty `content`.

## Environment

- OS: macOS
- Node: 22.x

Hiro-Chiba:
Fixes #552

## Summary
Return structured failures for denied reads and invalid numeric line ranges.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Emit `path_outside_workspace` in JSON and LLM formats when containment checks reject a file.
- Set exit status 1 for every denied read while preserving detailed text guidance.
- Reject zero-based and reversed numeric line ranges as `invalid_line_range`.
- Ensure failure paths never emit source content.
- Add regression coverage for absolute escapes, symlink escapes, and invalid ranges.

## Validation
- `npm test` (84 files, 1448 passed, 3 skipped)
- `npm run typecheck`

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/__tests__/read-containment.test.ts
+++ b/ix-cli/src/cli/__tests__/read-containment.test.ts
@@ -17,6 +17,7 @@ let workspace: string;
 let outside: string;
 let savedHome: string | undefined;
 let savedProfile: string | undefined;
+let savedExitCode: number | string | undefined;
 
 beforeEach(() => {
   home = mkdtempSync(join(tmpdir(), "ix-read-home-"));
@@ -24,6 +25,8 @@ beforeEach(() => {
   outside = mkdtempSync(join(tmpdir(), "ix-read-outside-"));
   savedHome = process.env.HOME;
   savedProfile = process.env.USERPROFILE;
+  savedExitCode = process.exitCode;
+  process.exitCode = undefined;
   process.env.HOME = home;
   process.env.USERPROFILE = home;
 
@@ -40,6 +43,7 @@ beforeEach(() => {
 afterEach(() => {
   if (savedHome === undefined) delete process.env.HOME; else process.env.HOME = savedHome;
   if (savedProfile === undefined) delete process.env.USERPROFILE; else process.env.USERPROFILE = savedProfile;
+  process.exitCode = savedExitCode;
   for (const dir of [home, workspace, outside]) rmSync(dir, { recursive: true, force: true });
 });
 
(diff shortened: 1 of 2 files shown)


KageBinary:
## Review — both fixes are real. One coverage loss to restore.

### The line-range fix catches a genuine silent-success bug

`readFileRange` does no bounds checking:

```ts
const content = lines.slice(lineStart - 1, lineEnd).join("\n");
```

So `ix read src/main.ts:0-0` was `slice(-1, 0)` → `[]`, and `:5-2` was `slice(4, 2)` → `[]`. Both returned **empty content with exit 0 and a `lineStart`/`lineEnd` the caller had asked for**, which is indistinguishable from a file that legitimately has nothing there. Rejecting up front is right, and the two test cases are exactly the two shapes.

### Routing the refusal to the requested format is right too

Structured `path_outside_workspace` on stdout for `json`/`llm` beats prose on stderr that a machine caller cannot parse. And dropping the `allowed root:` enumeration for machine formats is a quiet improvement nobody claimed: that hint lists **every registered workspace path on the machine**, and through `ix mcp` the recipient is a model. Keeping it for the human text path and not for agents is the right split — worth saying so in the docblock, since it currently reads as a formatting decision.

### The one thing to fix: the text branch is now untested

`runRead` passes `--format json` on every call, and this PR deleted the only two assertions that exercised the other branch:

```diff
-    expect(err).toContain("Refusing to read file outside the workspace");
-    expect(err).toContain("allowed root:");
+    expect(err).toBe("");
```

So the `stderr` + `allowed root:` + "Use --root …" path — the security refusal a human sees, and the only branch that still prints the roots — has no coverage at all now. One case with `--format text` asserting the refusal reaches stderr and names a root would restore it. Cheap, and it is the branch you least want to regress silently.

### Smaller

- `reportInvalidLineRange` is the fourth hand-rolled `json`/`llm`/stderr switch in this stack (#543, #545, and `guardReadable` right above it). See my note on #543 — one `reportError(code, message, format)` in `ui.ts` would collapse all of them, and would stop the four copies drifting on exit-code or stream choice.
- The `rangeStart !== undefined && rangeEnd !== undefined` guard is unreachable-safe rather than needed: the regex `/^(.+?):(\d+)-(\d+)$/` sets both or neither. No harm, and it does keep the condition honest if the regex ever gains an optional end.

Issue #554, opened 2026-08-30
[bug] ix map exits 0 and reports full_local_completed while silently dropping files to patch build errors


KageBinary:
## Summary

`ix map` can finish with `outcome: "full_local_completed"` and **exit 0** while
silently dropping files that failed to build a patch. On a plain
`npm ci`'d TypeScript project, 48 files were dropped and nothing in the
machine-readable output said so.

There are two separate defects here; they can be split if you prefer.

---

### Defect 1 — `Conflicting UpsertEdge identity` on ordinary dependency code

`deduplicateUpsertEdges` throws when one deterministic edge id resolves to two
different `(src, dst, predicate)` triples:

https://github.com/ix-infrastructure/Ix/blob/main/core-ingestion/src/patch-builder.ts#L120

The comment above it is explicit that this is meant to be an identity bug:

> A repeated id with different endpoints is an identity bug, not a duplicate;
> fail before commit rather than silently disconnecting part of the graph.

It fires on unremarkable, widely-used packages — 48 files across
`@modelcontextprotocol/sdk`, `ajv`, `zod` and `tsx`:

```
[patch build error] node_modules/@modelcontextprotocol/sdk/dist/cjs/client/index.js: Error: Conflicting UpsertEdge identity for deterministic id 9f4615e4-0a56-5ac1-e58c-35ecf0ee1d9c
[patch build error] node_modules/@modelcontextprotocol/sdk/dist/cjs/client/sse.js: Error: Conflicting UpsertEdge identity for deterministic id 6974f776-cde5-c941-89e0-eb956454a2ab
```

**Fully deterministic:** 48 errors on two consecutive runs, 48 distinct ids
(one per file), and the two runs' id sets are byte-identical.

Not a cross-workspace content collision — I checked. Copying three of the
offending files into two separate workspaces and mapping both produced zero
errors, so the collision is *within* a single file's op set, not between
workspaces.

### Defect 2 — the failure is invisible to `--format json`

`ingest.ts` catches the throw into a counter and writes only to stderr:

https://github.com/ix-infrastructure/Ix/blob/main/ix-cli/src/cli/commands/ingest.ts#L1519

`parseErrors` is tracked properly — it gates `ingestCompletedCleanly`, feeds
`skipReasons.parseError`, and is printed in text mode as `parse errors: N`.
But `ix map --format json` emits only:

```json
[...]

PR #555, opened 2026-08-30
fix(map): report files dropped by patch build errors

KageBinary:
Closes half of #554 (the reporting half; the id collision itself is a separate change).

## Problem

`ix map` can finish with `outcome: "full_local_completed"` and **exit 0** while silently omitting every file whose patch failed to build. Reproduced on a plain `npm ci`'d TypeScript project — 48 files dropped, nothing in the machine-readable output said so:

```json
{"file_count": 96, "region_count": 15, "levels": 2, "map_rev": 2483, "outcome": "full_local_completed"}
```

The count was already tracked — `ingestFiles` returns it as `parseErrors`, which gates `ingestCompletedCleanly` and feeds `skipReasons.parseError` — but nothing read it here. Both `describeEmptyCompletedMap` and `describeRegionlessCompletedMap` deliberately bail out when `parseErrors > 0`, so a partial map was the one failure mode `ix map` never mentioned in *any* format.

## Change

- `parse_errors` and `commit_errors` in the JSON payload, always present so a consumer can branch on them without a key check.
- A yellow stderr warning in every format, so the text path surfaces it too. stderr keeps the JSON/llm payload on stdout uncontaminated.
- `describeDroppedFiles` is exported and unit-tested rather than asserted through captured stderr.

## The exit code deliberately stays 0

Tempting to fail here, but the opencode and openclaw plugins read this command's JSON through a Bun `$` runner whose `.text()` throws on a non-zero exit — they would lose the very diagnostics this PR adds. That is the trap recorded on #539: fix the consumers first, then the exit code. Happy to follow up in that order if you want the stronger signal.

## Verification

Against the real reproducer (`ix-gemini-plugin/mcp` after `npm ci`), local backend 1.0.27:

```
{'file_count': 95, 'region_count': 14, 'levels': 2, 'map_rev': 2493,
 'outcome': 'full_local_completed', 'parse_errors': 48, 'commit_errors': 0}

  Map is incomplete: 48 files failed to build a patch. Those files are absent
  from the graph; re-run with 'ix ingest' to see the per-file errors.
```

Full gate: **1453 tests passed** (1447 + 6 new), 2 skipped; lint 0 errors; typecheck clean; knip clean.

### DIFF
--- a/ix-cli/src/cli/__tests__/llm-renderers.test.ts
+++ b/ix-cli/src/cli/__tests__/llm-renderers.test.ts
@@ -64,6 +64,22 @@ describe("renderMapLlm", () => {
     const noCross = captureLog(() => renderMapLlm(result, [region({ id: "y", label: "Y", crosscut_score: 0.005 })]));
     expect(noCross[1]).not.toContain("crosscut=");
   });
+
+  it("carries dropped-file counts, and only when non-zero (#554)", () => {
+    const result: MapResult = { file_count: 1, region_count: 1, levels: 1, map_rev: 1, outcome: "ok", regions: [], hierarchy: [] };
+
+    const dropped = captureLog(() => renderMapLlm(result, [], { parseErrors: 48, commitErrors: 3 }));
+    expect(dropped[0]).toBe("map files=1 regions=0 levels=1 rev=1 outcome=ok parse_errors=48 commit_errors=3");
+
+    // A clean ingest carries no signal, so the fields are dropped -- as is
+    // every other zero in this format.
+    const clean = captureLog(() => renderMapLlm(result, [], { parseErrors: 0, commitErrors: 0 }));
+    expect(clean[0]).toBe("map files=1 regions=0 levels=1 rev=1 outcome=ok");
+
+    // `ix subsystems` renders the same record with no local ingest to report.
+    const noIngest = captureLog(() => renderMapLlm(result, []));
+    expect(noIngest[0]).toBe("map files=1 regions=0 levels=1 rev=1 outcome=ok");
+  });
 });
 
 describe("renderSubsystemScoreLlm", () => {
--- a/ix-cli/src/cli/__tests__/map-auto-skip.test.ts
+++ b/ix-cli/src/cli/__tests__/map-auto-skip.test.ts
@@ -6,6 +6,7 @@ import { Command } from 'commander';
 import { afterEach, describe, expect, it, vi } from 'vitest';
 import {
   applyRequestedMapCoalesceExitCode,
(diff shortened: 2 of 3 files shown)


KageBinary:
## Self-review — the fix missed its own audience, fixed on the branch

The counts reached `--format json` and stderr, but **not the `map` llm record** — and `docs/llm-format.md` tells agents to pass `--format llm` unconditionally:

> It is accepted on every command that accepts `--format` … so consumers can pass the flag unconditionally without a per-command lookup.

So every plugin-driven `ix map` still saw a completed outcome with no hint that files were missing. That is the exact consumer #554 is about, and the one the PR left out.

Fixed in 5341c8d — `renderMapLlm` now takes the ingest summary and emits:

```
map files=1 regions=0 levels=1 rev=1 outcome=ok parse_errors=48 commit_errors=3
```

Emitted **only when non-zero**, which is this format's documented rule ("Zeros and other defaults are dropped where they carry no signal") and matches how `crosscut`, `children`, `parent` and `signals` already behave. A clean ingest says nothing. `ix subsystems` renders the same record with no local ingest and is unchanged — covered by a test.

## Also reworded the warning, and this one is a judgement call

It opened `Map is incomplete:` — while the same run persists a completed map baseline three lines earlier, so `ix doctor` answers **"Completed map for this workspace"** immediately afterwards. A user reading both is entitled to be confused.

The baseline is the part that is right. Gating completion on `parseErrors > 0` is what turned doctor red for ever in #530/#536: a file the parser cannot handle is normal, and a workspace containing one must still be able to record a completed map. So the warning is what had to give. It now states what is missing instead of contradicting the health signal beside it:

```
48 files failed to build a patch — those files are absent from the graph. Re-run with 'ix ingest' to see the per-file errors.
```

The `describeDroppedFiles` docblock now says why it does not claim incompleteness, so the next person does not "fix" it back. Revert just that hunk if you prefer the louder wording — it is independent of the llm change.

## Two things I did not change

- **`--silent` still emits this line.** `IX_AUTO_MAP=1` hooks run silent, so a repo with permanently unparseable files gets the warning on every save. Arguably right (it is actionable and true), arguably noise at that frequency. Worth a deliberate answer rather than my default.
- **`commitErrors` may deserve different treatment from `parseErrors`.** A parse error is permanent — that file will never build a patch. A commit error is transient, and a re-run would very likely fix it. Both are reported identically here and neither affects the baseline. If a transient commit failure should keep the workspace from recording completion so the next `ix map` retries, that is a separate change and I would rather you chose it than have me infer it.

`npm run test:unit` clean, `typecheck` clean.

PR #556, opened 2026-08-30
fix(ingest): keep both edges when one source key resolves two ways

KageBinary:
Closes the other half of #554 — the `Conflicting UpsertEdge identity` error itself. (#555 is the reporting half; the two are independent and can land in either order.)

## Root cause

An edge id folds the **unresolved** destination key:

```ts
id: edgeId(idPath, srcKey, edgeDstKey, r.predicate)
```

but an edge's identity is its **resolved** `(src, dst, predicate)`, which is what the dedup guard keys on:

```ts
const edgeIdentity = `${nodeId(idPath, srcKey)}\x00${dstNodeId}\x00${r.predicate}`;
```

`resolveKey` returns the bare name whenever a name is unknown or maps to several qualified keys. So two distinct caller spellings (`run` and `A.run`) can collapse to one `srcKey` — while `edgeResolution` is keyed on the *raw* name, so each caller still resolves its callee independently.

Both edges are real and distinct, but they mint the same id. `deduplicateUpsertEdges` sees one id with two endpoints and throws — correctly, by its own comment — and `ingest.ts` catches it, counts a `parseError`, and drops the whole file.

Instrumenting the throw on the real reproducer shows exactly that shape — same `src`, same predicate, different `dst`:

```
Conflicting UpsertEdge identity for deterministic id 6519c6cc-…
  | A src=89116ede-… dst=33f845bc-… pred=CALLS
  | B src=89116ede-… dst=dbfb22ad-… pred=CALLS
```

## Approach, and the one judgement call

The obviously "correct" fix is to fold `dstNodeId` into every edge id. I did **not** do that: it rewrites the id of every edge in every existing graph and forces a full re-ingest to avoid orphans.

Instead a pre-pass finds the id tuples that genuinely resolve to more than one destination, and salts **only those**. Every other edge keeps a byte-identical id, so no migration is needed. Verified directly by building the same file against both revisions — 46 edges, identical ids — and asserted in `leaves ids of non-colliding edges unsalted`.

Happy to switch to the wholesale derivation if you would rather take the churn once and have the id match the identity by construction.

The destination resolution moves into a pure `resolveDst` helper so the pre-pass can ask the same question the emission loop does. External-node minting stays exactly where it was, so op order is unchanged.

## Verification

Real reproducer (`ix-gemini-plugin/mcp` after `npm ci`), local backend 1.0.27:

| | before | after |
|---|---|---|
| `[patch build error]` lines | **48** | **0** |

The three new tests fail on `main` with the exact production error (`Conflicting UpsertEdge identity for deterministic id 0ca69949-…`) and pass here.

Suites: core-ingestion **354 passed** (351 + 3 new); ix-cli **1447 passed**, 2 skipped; lint 0 errors; typecheck clean.

> Note: 6 core-ingestion tests fail on `main` in my environment (Scala/Rust grammar snapshots) — unchanged by this PR, and they look like CI does not run this package's suite. Flagging separately.

### DIFF
--- a/core-ingestion/src/__tests__/patchBuilder.resolution.test.ts
+++ b/core-ingestion/src/__tests__/patchBuilder.resolution.test.ts
@@ -473,3 +473,80 @@ function handler($obj): void { handle(); $obj->handle(); }
     expect(claims).toHaveLength(1);
   });
 });
+
+describe('colliding edge ids (#554)', () => {
+  // `resolveKey` collapses an ambiguous or unknown name to the bare name, so two
+  // distinct caller names can share one `srcKey` — while `edgeResolution` is
+  // keyed on the RAW name, so each caller still resolves its callee separately.
+  // The edge id folds the unresolved `dstKey`, so both edges mint the same id
+  // while carrying different `dst`. Before the fix `deduplicateUpsertEdges`
+  // threw on that and the caller dropped the whole file.
+  const sourceFile = '/repo/app.ts';
+
+  function buildCollidingPatch() {
+    const result = fileResult(
+      sourceFile,
+      SupportedLanguages.TypeScript,
+      [entity('run', SupportedLanguages.TypeScript, 'function', 'A')],
+      [
+        { srcName: 'run', dstName: 'helper', predicate: 'CALLS' },
+        { srcName: 'A.run', dstName: 'helper', predicate: 'CALLS' },
+      ],
+    );
+
+    // Only the 'run' spelling resolves cross-file; 'A.run' falls through to the
+    // local node. Same src, same predicate, two different destinations.
+    const resolvedEdges: ResolvedEdge[] = [
+      {
+        srcFilePath: sourceFile,
+        srcName: 'run',
+        dstFilePath: '/repo/helpers.ts',
+        dstName: 'helper',
+        dstQualifiedKey: 'helper',
+        predicate: 'CALLS',
+        confidence: 0.9,
+      },
+    ];
+
+    return buildPatchWithResolution(result, 'test-hash', '', resolvedEdges);
+  }
+
+  it('does not throw when one source key carries two resolved destinations', () => {
+    expect(() => buildCollidingPatch()).not.toThrow();
+  });
+
+  it('commits both edges under distinct ids', () => {
+    const callEdges = buildCollidingPatch().ops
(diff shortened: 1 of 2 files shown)


KageBinary:
## Self-review — re-read the fix against the failure modes it has to survive. It holds.

Walked the collision cases rather than trusting the tests:

- **Two relationships, same tuple, same resolved `dstNodeId`** — `emittedEdgeIdentities` skips the second before the salt is consulted. Unchanged.
- **Two relationships, same tuple, different `dstNodeId`** — both salted, two distinct ids, both committed. The case #554 dropped 48 files on.
- **Three, two of which share a destination** — the duplicate is skipped by identity, the remaining two are salted independently. No pair collides.
- **Across files** — `edgeId` folds `idPath`, so nothing from another file can land in the same tuple. The pre-pass being per-file is correct, not a scoping bug.

`resolveDst` is genuinely pure — the external-node `UpsertNode` push stayed in the emission loop and is reached through the returned `externalPkg`, so op order is byte-identical to before. That was the thing most likely to go wrong in this refactor and it did not.

### Two things I want on the record before this merges

**1. The `edgeDstKey` computation now exists twice**, once in the pre-pass and once in the loop:

```ts
r.phpCallKind && (relationshipShapeCounts.get(baseKey) ?? 0) > 1 ? `${dstKey}:${r.phpCallKind}` : dstKey
```

They are identical today. If they ever drift, the pre-pass computes a tuple the loop never looks up, and the salt silently stops applying — which surfaces as `deduplicateUpsertEdges` throwing again, a long way from the edit. Worth hoisting to one helper beside `edgeIdTuple`.

**2. A salted id is stable only while the collision is.** If a file changes so a tuple stops colliding, that edge's id reverts to the unsalted form. Per-file re-ingest replaces the file's edges, so it resolves itself — but it does mean edge ids are content-dependent in a way they were not before, and it is the one property of this design that is not obvious from reading it.

### Still yours to decide

The trade-off flagged in the body stands, and I have not moved on it: this **salts only colliding tuples**, so every non-colliding edge keeps a byte-identical id and no graph needs re-ingesting. Folding `dstNodeId` into every edge id is more obviously correct — an edge's identity really is `(src, dst, predicate)` — but rewrites every edge id in every existing graph. The third test pins the current guarantee (`deterministicId('/repo/plain.ts:alpha:beta:CALLS')`), so if you prefer the wholesale fix it will fail loudly rather than quietly, which is what you want from that assertion.

PR #558, opened 2026-08-30
test(core-ingestion): fix the 6 failing tests and gate the suite in CI

### LINKED ISSUE #557: core-ingestion's 358-test suite is not run by CI, and 6 tests fail on main
## Summary

`core-ingestion` has a 358-test vitest suite that **no CI job runs**, and 6 of
those tests fail on `main` today. Three of them have been failing since the
Scala backend was extracted in April.

Found while verifying an unrelated change did not regress the package.

## CI does not run this suite

Every step in `.github/workflows/ci.yml` is `working-directory: ix-cli`:

```
106:  working-directory: ix-cli   → npm ci
109:  working-directory: ix-cli   → npm run lint
112:  working-directory: ix-cli   → npm run typecheck
115:  working-directory: ix-cli   → npm run knip
196:  working-directory: ix-cli   → npm run test:unit
201:  working-directory: ix-cli   → npm run test:coverage
205:  working-directory: ix-cli   → npm run smoke
```

`core-ingestion` appears only in `release.yml`, which builds it and does a
single smoke `import` of `dist/index.js`. Nothing executes
`core-ingestion`'s own tests.

Note `ix-cli`'s `npm test` *builds* core-ingestion
(`node scripts/build-core-ingestion.mjs`) before running vitest — so the
package is compiled on every CI run, which makes it easy to assume it is also
tested.

## Reproduce

Clean worktree at `3baa7dc` (current `main`), fresh `npm ci`:

```bash
cd core-ingestion && npm ci && npx vitest run
```

```
Test Files  4 failed | 30 passed (34)
     Tests  6 failed | 351 passed | 1 skipped (358)
```

Not environment-specific: reproduced on a fresh checkout with a clean install,
and the tree-sitter grammars are lockfile-pinned.

### 1. Three tests read a directory this repo no longer has

```
ENOENT: no such file or directory, open '<repo>/memory-layer/src/main/scala/ix/memory/model/Node.scala'
ENOENT: no such file or directory, open '<repo>/memory-layer/src/main/scala/ix/memory/model/Identifiers.scala'
ENOENT: no such file or directory, scandir '<repo>/memory-layer/src/main/scala'
```

- `parseFile.test.ts:24` — `snapshots Node.scala`
- `parseFile.test.ts:28` — `snapshots Identifiers.scala`
- `resolveEdges.integration.test.ts:9` — `resolves GraphQueryApi NodeKind references back to Node.scala`

`memory-layer/` was removed by **bc62830 — "Extract Scala backend to private
ix-memory-layer repo (#120)", 2026-04-02**. These three have failed
unconditionally for everyone ever since — **318 commits**.

### 2. One snapshot tracks a live, actively-edited source file

`snapshots doctor.ts` mismatches. The test snapshots
`ix-cli/src/cli/commands/doctor.ts` — a real source file under active
development — so *any* edit to it breaks the snapshot:

- snapshot recorded **2026-06-02**
- `doctor.ts` last edited **2026-08-29**

Worth a frozen fixture rather than a live file, otherwise this fails again on
the next doctor change.

### 3. Two tests assert a coarser symbol attribution than the parser now produces

**`queries.rust.test.ts` — "captures REFERENCES for impl Trait parameters (Bug 1 — abstract_type)"**

```
- Expected: { srcName: "impl_trait.rs", dstName: "Sink", predicate: "REFERENCES" }
+ Received: { srcName: "run",           dstName: "Sink", predicate: "REFERENCES" }
```

The test's own comment says *"Parameters are not inside a struct/class range,
so the src falls back to the file"*. The parser no longer falls back — it
attributes the reference to the enclosing function `run`.

**`queries.scala.test.ts` — "captures class-level val and var definitions but not method-local vals"**

```
- Expected: { srcName: "Foo",     dstName: "NodeKind", predicate: "REFERENCES" }
+ Received: { srcName: "Foo.baz", dstName: "NodeKind", predicate: "REFERENCES" }
```

Same shape: attributed to the specific val `Foo.baz` rather than the enclosing
class.

**These read as stale tests, not regressions** — in both cases the current
output is strictly more precise, which looks like an intentional improvement
that predates nobody running these tests. The files were last touched
2026-03-16 and 2026-03-24, with **64 commits to `core-ingestion/src` since**.

Flagging rather than asserting: if the coarser attribution was deliberate,
these two are real regressions and matter more than the rest of this issue.
Someone who knows the intended attribution should confirm which way it goes
before the expectations are updated.

## Suggested fix

1. Repoint or delete the three `memory-layer/` tests — the fixtures live in
   another repo now.
2. Give the `doctor.ts` snapshot a frozen fixture.
3. Confirm the intended attribution, then update the two expectations (or file
   the regression).
4. Add a CI job running `core-ingestion`'s suite, so the package that does the
   parsing is gated like `ix-cli` is.

Step 4 is the one that matters — without it these come back.

## Environment

- `main` @ `3baa7dc`, Node v22.22.2, Linux, fresh `npm ci`

KageBinary:
Closes #557.

`core-ingestion` has a **358-test suite that no CI job ran**, and 6 of those tests failed on `main`. It had a `test` script all along (`vitest run --pool threads`) — nothing ever invoked it, which is probably how it drifted this far.

> **Correction:** an earlier version of this description said the package had no `test` script, and this branch added one on that basis. It did not — main already had one, so the added line was a **duplicate key** in the same JSON object and was silently dropped by every parser. Removed in 5c663bb; the surviving script is unchanged, so the CI step below runs exactly what it has been running.

```
before:  Test Files  4 failed | 30 passed (34)
              Tests  6 failed | 351 passed | 1 skipped (358)

after:   Test Files  34 passed (34)
              Tests  357 passed | 1 skipped (358)
```

## Three groups, three causes

**1. Fixtures in a repo that no longer exists** (3 tests)

They read `memory-layer/`, which **#120 extracted to a separate private repo on 2026-04-02**. ENOENT for everyone since — **318 commits**. Because the backend is not public, they *cannot* simply be repointed at it, so I wrote synthetic Scala fixtures under `test-fixtures/`, following the `mini-repo` pattern already in that directory.

Checked the integration test still does real work rather than passing vacuously:

```
files parsed: 3
NodeKind refs from GraphQueryApi: 2
  GraphQueryApi.kindOf     -> NodeKind @ .../model/Node.scala (conf 0.9)
  GraphQueryApi.defaultKind -> NodeKind @ .../model/Node.scala (conf 0.9)
```

**2. A snapshot of a live source file** (1 test)

`snapshots doctor.ts` snapshotted `ix-cli/src/cli/commands/doctor.ts` — a file under active development, so *any* edit to it broke the test. Snapshot recorded **2026-06-02**; `doctor.ts` last edited **2026-08-29**. A snapshot needs an input that changes only when someone means it to, so it now uses a frozen fixture covering the same constructs.

**3. Two stale attribution expectations** (2 tests)

I flagged in #557 that these might be real regressions and shouldn't be "fixed" by rewriting expectations to match whatever the code does. They are not regressions — three independent confirmations:

- `index.ts` documents the behaviour: *"Prefer the enclosing function/method so parameter and return-type references point back to the actual caller. Fall back to class, then file."*
- `queries.go.test.ts` asserts the same rule (`srcName: 'use'` for refs inside a function) and **passes**.
- The Scala test contradicted *itself* — expecting `Foo` for REFERENCES two lines above `Foo.baz` for CALLS.

So both encoded pre-fix behaviour. Updated to `run` and `Foo.baz`.

## Keeping it fixed

A step in the **existing** test job — which already installs and builds this package, so it costs one step instead of a whole job and inherits the matrix's OS/node spread (worth having: the tree-sitter grammars are native).

**Verified the gate is real, not decorative.** Reverting the attribution to the old file-level fallback:

```
Test Files  5 failed | 29 passed (34)
     Tests  15 failed | 342 passed | 1 skipped (358)
```

Mutation reverted, tree clean.

## Checks run

core-ingestion 357 passed · ix-cli **1447 passed** / 2 skipped · lint 0 errors · typecheck clean · knip clean · build clean (tsconfig `include: [src]`, so fixtures never reach `dist`).

Linux only locally — this PR's own CI run is the first proof of the Windows/macOS matrix legs.


KageBinary:
## Self-review — one defect, fixed on the branch

The PR body says core-ingestion "had no `test` script". It did:

```
main:  "test": "vitest run --pool threads"
```

so the line this branch added was a **second `"test"` key in the same object**:

```json
"scripts": {
  "test": "vitest run",                    <- added here
  "build": "...",
  "test": "vitest run --pool threads",     <- already on main
  "test:watch": "vitest --pool threads"
}
```

Every JSON parser keeps the last one — which is why CI was green. `npm test` has been running the pre-existing pool-threads script all along, and the added line was dead.

That is worth removing rather than shrugging at. `npm pkg`, publish-time normalisation and any jq-based tooling silently keep one key and drop the other; and if the dead line ever ends up *last* — a merge, a formatter, `npm pkg fix` — CI switches vitest pools with nobody touching CI. Not cosmetic in this package: the suite loads native tree-sitter grammars, and the pool decides worker isolation and how the addons are loaded.

Removed in 5c663bb. The surviving script is unchanged, so the CI step this branch adds runs exactly what it has been running. Re-ran locally:

```
Test Files  34 passed (34)
     Tests  357 passed | 1 skipped (358)
```

Please correct the "had no `test` script" line in the body before merging — the diff contradicts it, which costs a reviewer time.

## The rest holds up

- **CI placement.** `core-ingestion/node_modules` is cached and restored in the same job, and `ix-cli`'s build step runs `build-core-ingestion.mjs`, so `npm test` has its deps. Cache key includes the lockfile hash, so a devDependency change busts it. Costs one step rather than a job, and picks up the 5-way OS/node matrix for free — which matters for native grammars.
- **The two attribution expectations.** The argument that these encoded pre-fix behaviour rather than a regression is well supported — `index.ts` documents the enclosing-function rule, `queries.go.test.ts` asserts it and passes, and the Scala test contradicted itself two lines apart. Agreed.
- **Snapshotting a live source file.** Replacing `doctor.ts` with a frozen fixture is the right call; a snapshot whose input is under active development tests nothing but the last edit.

PR #569, opened 2026-09-02
fix(test): stop watch-dedup's child build failing on unrelated type errors, and keep it out of the working tree

### LINKED ISSUE #567: watch-dedup: "starts both the tsx source CLI and a normal built CLI child" fails intermittently, and leaves a full build in the package root when interrupted
`ix-cli/src/cli/__tests__/watch-dedup.test.ts:99`.

## What was seen

Five failures on 2026-09-01, all of `canonical watch refresh > starts both the tsx source CLI and a normal built CLI child`:

- once inside a full `vitest run` (1 failed / 1508 passed)
- then **four consecutive times** running that file alone, in the same worktree, against clean `main` (`793c56b`) with local changes stashed

## What could not be reproduced

In a **fresh worktree at the same commit**, the test passed **15/15 on every attempt** — roughly twelve runs, including six while a full `vitest run` was executing concurrently to load the machine.

I could not capture the assertion message before the failing worktree was removed, which is the main gap here. **If anyone hits this, please paste the failure output** — the two `expect(status, stderr)` assertions carry the child's stderr and would name the cause immediately.

## Hypotheses tested and ruled out

| hypothesis | result |
|---|---|
| Disk pressure (machine was at 97%, 1.6G free) | **No.** The build the test performs is only **1.3M / 89 files**. |
| Leftover `.watch-child-runtime-*` debris from an earlier interrupted run | **No.** Recreated the debris (including a full `dist` inside it) and the test still passed 15/15. |
| The update notice leaking onto stdout and breaking the `--version` assertion | **No.** With a fresh `IX_HOME` seeded exactly as the test seeds it, stdout is exactly `0.10.5` and nothing else. |

## Why this test is more exposed than most

The test body does real work in the environment rather than asserting on pure functions like the rest of the file:

1. spawns `src/cli/main.ts` under tsx and asserts `--version` on stdout
2. **shells out to a full `tsc -p tsconfig.build.json` build** and asserts `build.status === 0` (`:130-143`)
3. spawns the built CLI and asserts on `--help`

Any environmental problem in step 2 or 3 fails the test for a reason unrelated to what it is checking, which is whether the child process *starts*.

## A separate, definite defect worth fixing either way

`:109` creates its temp dir **inside the package root**:

```js
const tempRoot = fs.mkdtempSync(path.join(packageRoot, ".watch-child-runtime-"));
```

Cleanup is in a `finally`, which does not run if the process is killed. An interrupted run therefore leaves a complete `dist` build sitting in the working tree — I hit exactly this and found `.watch-child-runtime-ZLHEp7/` as an untracked directory afterwards. It is also not in `.gitignore`, so it shows up in `git status` and can be committed by accident.

Moving `tempRoot` to the OS tmpdir fixes both, and costs nothing — nothing in the test requires it to be under `packageRoot` (`packageRoot` is already passed explicitly as `cwd` to each `spawnSync`).

Found while verifying #566; not caused by it.

josephismikhail:
Fixes #567.

Three findings. The issue reports one intermittent failure and one definite defect; both are here, plus the repo-hygiene defect that turned out to be the worktree-to-worktree difference behind the "fresh worktree passes 15/15" result.

---

## 1. The intermittent failure — reproduced, with the assertion message

`canonical watch refresh > starts both the tsx source CLI and a normal built CLI child` shells out to a full `tsc -p tsconfig.build.json` and asserts `build.status === 0`. `tsconfig.json` has `"include": ["src"]`, so that build type-checks **every** `.ts` file in the working tree — untracked ones included.

`git stash` without `-u` does not stash untracked files. A WIP module left in `src` therefore fails this test, in that worktree, every time, and passes in a fresh worktree at the same commit, which has no such file. It is also the **only** test that fails, because nothing else in the suite shells out to tsc — vitest transforms TypeScript with esbuild, which does not type-check.

That is the reported shape exactly: `1 failed / 1508 passed` in a full run, four consecutive failures of the file alone in the same worktree, 15/15 in a fresh one.

Reproduced on `main` (`a2cab21`):

```console
$ cat > ix-cli/src/cli/wip-scratch.ts <<'TS'
// Untracked WIP file, of the kind `git stash` (without -u) leaves behind.
export const answer: number = "not a number";
TS
$ npx vitest run src/cli/__tests__/watch-dedup.test.ts
 ❯ src/cli/__tests__/watch-dedup.test.ts (15 tests | 1 failed)
   × starts both the tsx source CLI and a normal built CLI child

AssertionError: src/cli/wip-scratch.ts(2,14): error TS2322:
  Type 'string' is not assignable to type 'number'.
: expected 2 to be +0 // Object.is equality
 ❯ src/cli/__tests__/watch-dedup.test.ts:143:58
    143|       expect(build.status, build.stderr || build.stdout).toBe(0);

 Tests  1 failed | 14 passed (15)
```

That is the `expect(status, stderr)` message #567 asks for and could not capture.

**Fix (two parts — see the review round below for why one was not enough):**

1. Compile the CLI entry's **own module graph** rather than `src`: a generated tsconfig with `files: ["<packageRoot>/src/cli/main.ts"]` and `include: []`. tsc follows the imports, so the program is exactly what the child needs to run. A file the child never loads is no longer part of the build at all.
2. `--noCheck`, for the remaining case of a *semantic* error in a file that **is** in that graph (one being edited right now). Syntax errors still fail the emit, correctly — those genuinely break the child.

The package as a whole is still type-checked by `npm run build`, `npm run typecheck` and CI (`.github/workflows/ci.yml:113`); none of that is this test's job.

**Re-reproduced after the fix:** with the untracked type-error file present, **16/16 pass**. With a *syntactically incomplete* untracked file present (`export const half = {`), also **16/16**.

This does not claim to be the only way that test can fail environmentally — just that it is a mechanism that produces the reported signature exactly, and that after this change an untracked file the CLI does not import cannot fail it.

---

## 2. The definite defect — a build left in the working tree

`:109` created its temp dir inside the package root, and cleanup is a `finally` a killed runner never reaches:

```js
const tempRoot = fs.mkdtempSync(path.join(packageRoot, ".watch-child-runtime-"));
```

Confirmed it is invisible to `.gitignore` and committable:

```console
$ mkdir -p .watch-child-runtime-ZLHEp7/dist/cli && touch .watch-child-runtime-ZLHEp7/dist/cli/main.js
$ git check-ignore -v .watch-child-runtime-ZLHEp7; echo "exit=$?"
exit=1
$ git status --short
?? .watch-child-runtime-ZLHEp7/
```

### The suggested fix does not work, and the reason is worth recording

#567 proposes moving `tempRoot` to the OS tmpdir, on the grounds that "nothing in the test requires it to be under `packageRoot` (`packageRoot` is already passed explicitly as `cwd` to each `spawnSync`)". I tried that first. `cwd` has no bearing on ESM bare-specifier resolution, which walks up from the **module's own** path:

```console
$ npx vitest run src/cli/__tests__/watch-dedup.test.ts
AssertionError: Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'commander'
  imported from C:\...\Temp\ix-watch-child-runtime-zouOJ9\dist\cli\main.js
 ❯ src/cli/__tests__/watch-dedup.test.ts:169:42   <- the built-child assertion
```

Being under `packageRoot` is exactly why the directory was there.

**Fix:** build into `ix-cli/node_modules/.cache/ix-watch-child-runtime-*`, which satisfies both constraints. Node walks `.cache/<tmp>/dist/cli` up to `packageRoot` and finds `packageRoot/node_modules`; git never sees anything under `node_modules`.

**Re-reproduced after the fix** — debris from a simulated killed run:

```console
$ mkdir -p ix-cli/node_modules/.cache/ix-watch-child-runtime-KILLED/dist/cli
$ touch ix-cli/node_modules/.cache/ix-watch-child-runtime-KILLED/dist/cli/main.js
$ git status --short          # nothing; the tree is clean of it
$ git check-ignore -v ix-cli/node_modules/.cache/ix-watch-child-runtime-KILLED
.gitignore:23:node_modules	ix-cli/node_modules/.cache/ix-watch-child-runtime-KILLED
```

### Regression test

A new test pins the property by **asking git**, not by matching `.gitignore` text: `git check-ignore --quiet <buildRoot>` must exit 0. Outside a work tree (installed tarball, source archive) there is no git answer to get, so it falls back to the structural half of the same claim — the build lives inside `node_modules`. Neither branch is vacuous.

Validated by restoring the bug: pointing `childBuildCacheRoot` back at `packageRoot` fails it with `git does not ignore <packageRoot>`, and it passes again once reverted.

`ix-cli/.watch-child-runtime-*` is also added to `.gitignore` — not for the new location, which `node_modules` already covers, but so a directory an interrupted run left on a pre-fix checkout does not surface as untracked work.

---

## 3. A third defect found on the way: `ix-cli/node_modules` is a tracked symlink

This is the per-worktree environment difference the issue was hunting.

```console
$ git ls-files -s ix-cli/node_modules
120000 ce7a785d0928f2c44baea81d28272794f89c1427 0	node_modules
$ git cat-file -p HEAD:ix-cli/node_modules
/home/ianhock/ix-work/oss/Ix/ix-cli/node_modules
```

A public repo should not ship an absolute path into a contributor's home directory. It has been tracked since #545 (`3ca56a6`), and it has real costs:

* `npm install` replaces it with a real directory, so every checkout reports a spurious `D ix-cli/node_modules` in `git status` forever.
* What `ix-cli/node_modules` *means* depends on whether a given checkout has been installed into — a fresh worktree gets the symlink, a worked-in one gets a real directory. That is precisely the "fresh worktree vs. the one it failed in" axis #567 could not isolate.

**Why the ignore rules missed it:** `node_modules/` matches directories only, and `ix-cli/node_modules/*` matches a directory's *contents* — neither covers a **symlink** named `node_modules`. Verified in a scratch repo:

```console
$ printf 'node_modules/\nix-cli/node_modules/*\n' > .gitignore
$ ln -s /some/abs/path ix-cli/node_modules
$ git check-ignore -v ix-cli/node_modules; echo "exit=$?"
exit=1                                   # not ignored

$ printf 'node_modules\n' > .gitignore   # no trailing slash
$ git check-ignore -v ix-cli/node_modules
.gitignore:1:node_modules	ix-cli/node_modules
```

**Fix:** `git rm --cached ix-cli/node_modules`, and drop the trailing slash so the rule matches the entry whatever its type, at any depth. That subsumes the two package-specific lines, which are removed.

> [!WARNING]
> After merging, anyone whose checkout relied on that symlink must run `npm install` in `ix-cli/`.

---

## Verified against the real backend

This PR is test infrastructure only — it changes no shipped code path — but the same backend the other two issues were reproduced against was used to smoke it, so the claim that nothing here affects the CLI is measured rather than assumed.

**Harness:** `docker-compose.standalone.yml` — `ghcr.io/ix-infrastructure/ix-memory-layer:latest` against **ArangoDB 3.12.10-1 enterprise**.

```console
$ ix map $REPO --verbose
[ingest-save] files=1 resolveEdgesMs=2 buildPatchMs=24 commitMs=17 totalSaveMs=45 nodes=2 edges=3
Architectural Map — 21 files · 3 regions
```

Identical to the other branches and to `main` against the same backend.

---

## Review round 1

`/code-review 569` raised three findings; all three are fixed in `8dc8b34`.

1. **`--noCheck` closed only half the class (medium).** It suppresses semantic diagnostics only, so a *syntactically incomplete* untracked file still failed the emit — and a mid-edit WIP module usually is exactly that. Confirmed on the branch:

   ```console
   $ printf 'export const half = {
' > src/cli/zz-wip-midedit.ts
   $ npx vitest run src/cli/__tests__/watch-dedup.test.ts
    Tests  1 failed | 15 passed (16)
     ❯ src/cli/__tests__/watch-dedup.test.ts:214   <- the build assertion
   ```

   Fixed by scoping the program to `files: [src/cli/main.ts]` with `include: []`, as described above. The review was also right that the original PR body overclaimed; it is corrected.

2. **`--noCheck` needs TypeScript >= 5.6, the range said `^5.4.0` (low).** The lockfile pins 5.9.3, so `npm ci` was safe, but any install resolving to 5.4.x would fail the emit with `error TS5023: Unknown compiler option '--noCheck'` — blaming this test again. Range bumped to `^5.6.0`, lockfile spec refreshed.

3. **Relocating under `node_modules/.cache` removed the only signal that debris exists (low).** `git status` cannot show it and nothing prunes it until the next `npm ci`, and `release.yml:284` copies `ix-cli/node_modules` wholesale into the release tarball. The test now sweeps leftover `ix-watch-child-runtime-*` siblings before `mkdtempSync`, so at most one interrupted run's build is ever on disk.

---

## Review round 2

`/code-review 569` on the fix round raised two findings. Both were real; both are fixed in `8877ea2`.

### 1. The sweep could delete a concurrent run's build (the bug this PR exists to remove)

The sweep added in round 1 deleted every `ix-watch-child-runtime-*` sibling unconditionally — including one a *live* instance of this same test was building into. Two runners in one checkout is ordinary: vitest in watch mode in one terminal, `npm test` in another. Run B deletes run A's tree between A's emit and A's child spawn and A fails with `ERR_MODULE_NOT_FOUND`; or, on Windows, B throws `EPERM` against A's live tree, since `force` swallows only `ENOENT`. Either way it manufactures exactly the intermittent single-test failure #567 is about.

**Fix:** the build directory carries its owner's pid — the convention `ix upgrade`'s temp dirs already use — and the sweep skips any whose process is still alive.

Verified with one directory owned by a live `node` process and one by a dead pid:

```console
before:
  ix-watch-child-runtime-52308-LIVE
  ix-watch-child-runtime-999999-DEAD
Tests  16 passed (16)
after:
  ix-watch-child-runtime-52308-LIVE
```

### 2. The regression test did not pin the `.gitignore` change it ships with

Correct, and a good catch: `node_modules/.cache` is a real directory, so the *old* directory-only `node_modules/` rule matched it too. The test was green on `main`, and restoring that rule would have left it green while `ix-cli/node_modules` became committable again — #545 all over.

**Fix:** it now also checks a path that does not exist. Git treats such a path as a non-directory, so only a rule *without* the trailing slash matches it — which is exactly the property that changed:

```console
# rules: node_modules/ (old)
  ix-cli/__probe__/node_modules -> exit 1     # not ignored
# rules: node_modules (new)
  ix-cli/__probe__/node_modules -> exit 0
```

Validated by restoring `node_modules/` in `.gitignore`:

```
AssertionError: the node_modules ignore rule is directory-only again;
  a symlink or file named node_modules is committable (Ix#545)
Tests  1 failed | 15 passed (16)
```

### Review round 3

`/code-review 569` on the round-2 fixes raised three findings. All three were real; all three are fixed in `a777f5f`.

### 1. The ignore probe could still be false-green (medium/low)

`git check-ignore`'s exit code is 0 for a rule from the user's **global** `core.excludesFile` or `.git/info/exclude`, not only from the repo's `.gitignore`. So a contributor whose personal ignore file says `node_modules` (no slash — a common spelling) could restore the directory-only rule here, make `ix-cli/node_modules` committable again, and still see 16/16. The same false green round 2 set out to close, relocated.

**Fix:** `-v` instead of `--quiet`, asserting the *source* is this repo's own `.gitignore`.

Verified — with `.gitignore` regressed **and** a global excludes file saying `node_modules`:

```console
$ git check-ignore -q --no-index ix-cli/__ix_ignore_probe__/node_modules; echo $?
0                                          # the old assertion passes

$ git check-ignore -v --no-index ix-cli/__ix_ignore_probe__/node_modules
/tmp/global-excludes:1:node_modules	ix-cli/__ix_ignore_probe__/node_modules

$ npx vitest run src/cli/__tests__/watch-dedup.test.ts
AssertionError: the node_modules ignore rule is directory-only again; a symlink or
  file named node_modules is committable (Ix#545):
  expected 'C:/Users/…/global-excludes' to be '.gitignore'
```

The review's mirror case is fixed too: `--is-inside-work-tree` is `true` for an **enclosing** repo, which is exactly the unpacked-source-archive case the fallback branch exists for — so the probe would have run against *that* repo's conventional `node_modules/` rule and failed, blaming Ix. The branch now asks whether the toplevel *is* this repo.

(While fixing this I found my own `-v` parser split the source on the first `:` and so returned `C` for a Windows path. It still failed the assertion, but by luck rather than by meaning. It anchors on `:<line>:` now.)

### 2. The pid parsing was a worse copy of a convention already in this codebase (low)

`isLiveScratch` in `upgrade.ts:1078` documents the part that matters — *"Anything unparseable is treated as dead: the pre-existing naming had no pid, and those must stay collectable."* — and my copy got it backwards. Two immortal-debris cases followed:

* a name with no pid (debris from this branch's own intermediate commit `8dc8b34`) was skipped **forever**;
* `parseInt("4abcde")` is `4` — a live kernel process on Windows *and* Linux — so a mkdtemp suffix starting with a digit was also immortal.

**Fix:** `isLiveScratch` is exported and used. One convention, one definition.

### 3. The `.gitignore` entry silenced #567's symptom without removing it (low/medium)

`ix-cli/.watch-child-runtime-*` hid the 1.3 MB build in the package root — the debris the issue actually reports finding — while nothing swept that location. Before this PR `git status` surfaced it, which is how it was noticed at all. The review is right that round 1's own argument (losing the debris signal is a real cost) applies here.

**Fix:** the sweep reclaims the legacy path too.

### Verified with one directory of each kind present

```console
before (cache):        ix-watch-child-runtime-35156-LIVE   ← live node process
                       ix-watch-child-runtime-999999-DEAD
                       ix-watch-child-runtime-AbC123       ← unparseable
                       ix-watch-child-runtime-4abcde       ← parsed as pid 4
before (packageRoot):  .watch-child-runtime-LEGACY

Tests  16 passed (16)

after (cache):         ix-watch-child-runtime-35156-LIVE
after (packageRoot):   (empty)
```

Only the live run's tree survives; the two previously-immortal cases and the legacy location are now collected.

### Verification after the fixes

| check | result |
|---|---|
| `watch-dedup.test.ts` | 16/16 |
| `npx vitest run` (full) | **1488 passed, 3 failed, 21 skipped** |

The 3 failures remain the pre-existing Windows-only `EPERM` symlink failures in `read-containment.test.ts`.

---

## Verification after the fixes

| check | result |
|---|---|
| `npx eslint` (touched file) | 0 errors |
| `watch-dedup.test.ts` | 16/16 |
| `npx vitest run` (full) | **1488 passed, 3 failed, 21 skipped** |

The 3 failures remain the pre-existing Windows-only `EPERM` symlink failures in `read-containment.test.ts`.

---

## Verification

| check | result |
|---|---|
| `npx eslint src/cli/__tests__/watch-dedup.test.ts` | clean |
| `npx vitest run` (full) | **1488 passed, 3 failed, 21 skipped** |
| `watch-dedup.test.ts` alone | **16/16** |

The 3 failures are the pre-existing Windows-only `EPERM: operation not permitted, symlink` failures in `read-containment.test.ts`; they fail on `main` on this machine too and are untouched by this branch.

<details>
<summary>A fourth, unrelated failure seen while verifying (not fixed here)</summary>

`stitch-scope-cache.test.ts > asks the backend on a miss and records the answer for next time` fails on **every run after the first**, on clean `main`, on Windows. The test isolates itself with `process.env.HOME = <tmpdir>`, but Windows `os.homedir()` reads `USERPROFILE`, so the override does nothing and the test writes its fixtures into the developer's real `~/.ix`:

```console
$ rm -f ~/.ix/stitch_scope_*.json
$ npx vitest run src/cli/__tests__/stitch-scope-cache.test.ts   # 10 passed
$ npx vitest run src/cli/__tests__/stitch-scope-cache.test.ts   # 1 failed | 9 passed

$ cat ~/.ix/stitch_scope_265bbafd247e.json
{"workspaceId":"ws-fresh","systemId":"sys-fresh"}   # left by the previous run
$ cat ~/.ix/stitch_scope_1c053616ad8e.json
{not json
```

The second run finds `ws-fresh` already cached, so `workspaceSystem` is never called. Out of scope for #567; flagged so it is not mistaken for fallout from this branch.

## Not addressed

The issue notes the test "does real work in the environment rather than asserting on pure functions like the rest of the file" — steps 1 and 3 still spawn real children, deliberately. That *is* the subject of the test: whether the two child shapes start. This PR removes the exposure that was not the subject (a whole-package type-check) and leaves the part that is.


KageBinary:
Reviewed at `08e7df6`. The diagnosis in part 1 is the best thing in this PR — `tsconfig.json`'s `"include": ["src"]` type-checking untracked files, and the fact that this is the only test in the suite that shells out to `tsc` while vitest transforms with esbuild and never type-checks, together explain the exact reported signature (one failure in a full run, four in isolation in the *same* worktree, 15/15 in a fresh one). That is a real root cause, not a plausible story. The pid-owned build dir and the ask-git-rather-than-match-the-text regression test are both the right shapes.

Two findings, one of which is a collision you should decide rather than a defect.

---

### 1. The `.gitignore` fix stops one line short of the defect it describes

The PR states the rule precisely:

> `node_modules/` matches directories only, and `ix-cli/node_modules/*` matches a directory's *contents* — neither covers a **symlink** named `node_modules`.

That is exactly right, and it is fixed for `node_modules`. The three lines immediately below it keep the trailing slash:

```
core-ingestion/dist/
ix-cli/dist/
ix-cli/coverage/
```

so the identical hole stays open for them. On this branch:

```console
$ ln -s /tmp ix-cli/dist
$ git status --short -- ix-cli/dist
?? ix-cli/dist
```

That is not hypothetical here — a symlinked `dist` is how a worktree borrows a build from the main clone, the same practice that produced the `node_modules` symlink in #545. And the new regression test would stay green through it: `explain(nonDirectory)` probes `__ix_ignore_probe__/node_modules` only, so it pins the property for one of the four rules. Dropping the slash from all four costs three characters and makes the test's claim true of the file rather than of one line in it.

(`core-ingestion/dist` reads as ignored in a working tree only because it *exists* as a directory there. Probe it as a non-existent path and it is unignored, same as the other two.)

### 2. This collides with #580, which does the same untrack

#580 is standalone and does exactly parts of this PR's finding 3 — `git rm --cached ix-cli/node_modules` plus the `.gitignore` fix — with the slash dropped from all four rules. They conflict:

```console
$ git merge-tree --write-tree origin/fix/567-watch-dedup-temp-dir origin/fix/untrack-node-modules-symlink
CONFLICT (content): Merge conflict in .gitignore
```

Both are correct; carrying it twice is the only problem. Your call which way — either take #580 first and drop finding 3 from here (this PR keeps the `ix-cli/.watch-child-runtime-*` line, which #580 does not have and should), or extend this one to the other three rules and I will close #580. I have no preference beyond it landing once.

---

### Checked and not findings

- **The TypeScript `^5.4.0` → `^5.6.0` bump is justified.** `--noCheck` is still passed (`watch-dedup.test.ts:339`) alongside the scoped `files: [main.ts]` program, so round 1's finding 2 still applies. Worth saying because the round-1 write-up reads as if scoping *replaced* `--noCheck`.
- **The sweep cannot eat a live run's tree.** `mkdtempSync` bakes the pid into the name at creation, so there is no window between "directory exists" and "directory is attributable", and `isLiveScratch`'s `-(\d+)-[^-]*$` parses the `mkdtemp` suffix correctly (it is alphanumeric, never containing a dash).
- **`check-ignore --no-index` is the right call** for a path that may be tracked, and the `sourceOf(...) === ".gitignore"` assertion correctly rejects a match that came from a global excludes file instead of this repo's.

### One correction to the description

> This PR is test infrastructure only — it changes no shipped code path

It changes a shipped *file*: `upgrade.ts`'s `isLiveScratch` goes from module-private to exported, so the test can share the liveness rule rather than reimplement it. That is the right move and changes no behaviour, but "no shipped code path" reads as "no shipped file", and the next person diffing the branch will find one.

KageBinary:
Follow-up now that this has landed, since it came out of reviewing it:

**#580** takes the `.gitignore` fix the rest of the way — `core-ingestion/dist`, `ix-cli/dist` and `ix-cli/coverage` kept the trailing slash, so a symlink at any of them was still committable (`?? ix-cli/dist`). Merged as `59933d5`. It also extends this PR's regression test to probe all four rules instead of `node_modules` alone — it would have stayed green through every one of those lines.

**#585** records something the test cannot help: `git check-ignore` refuses any path beyond a symbolic link, so the `cacheRoot` assertion fails in a checkout whose `ix-cli/node_modules` is itself a symlink —

```console
```

— which is the worktree layout that produced the #545 symlink in the first place. Not reachable in CI (`npm ci` makes a real directory) and reproduced identically on `main` before #580, so nothing here regressed; it is just worth knowing before someone chases it as a flake.

Issue #527, opened 2026-08-28
[bug] Scoped workspace reset leaves patch deduplication state behind


Hiro-Chiba:
## Description

`POST /v1/reset/workspace` returns success and removes the workspace's live graph, but reapplying the exact same source patch is treated as an idempotent no-op. The commit response is successful and the CLI counts the patch as applied, while the workspace remains empty.

This indicates that workspace-scoped reset removes graph nodes and edges without removing the patch identity or deduplication record used by later commits.

## Steps to reproduce
1. Create a repository with one TypeScript class.
2. Run `ix map --full --silent` and confirm that workspace-scoped stats contain nodes and edges.
3. Call `POST /v1/reset/workspace` with that workspace ID.
4. Remove the workspace's local ingest and stitch-scope cache entries so the same file is submitted again.
5. Run the same map command without changing the source.
6. Inspect workspace-scoped stats.

## Expected behavior

The scoped reset removes all workspace-owned graph and patch-deduplication state. Reapplying the same patch recreates the graph.

## Actual behavior

The reset returns HTTP 200. The next ingest reports one committed patch, but map validation finds zero files and workspace-scoped stats remain at zero nodes and zero edges.

## CLI output

```text
before reset: 3 nodes, 4 edges
reset: HTTP 200 {"ok":true,"message":"Workspace data deleted."}
re-ingest: 1 patch committed
map: completed response contains 0 files
after re-ingest: 0 nodes, 0 edges
```

## Environment
- OS: macOS
- Node: 22.x
- Ix version: 0.10.4
- Backend release: 1.0.28

KageBinary:
Moved to ix-infrastructure/Ix-memory#174 — `POST /v1/reset/workspace` is served by the memory layer, so the fix belongs there and there was no counterpart issue on that repo.

Added one thing to the new issue that changed since this was written: as of 0.10.5 the empty-map guard turns this into a **non-zero exit** from `ix map`, and the message it prints blames the source language ("The active backend may not map this source language yet"). That is wrong for this bug — the language is fine, the write was swallowed by the dedup record. Same root cause, but the symptom went from a silently empty graph to a loud failure pointing the wrong way.

Keeping this open or closing it is your call; the backend issue is the actionable one.

KageBinary:
Both halves are now up:

[...]

PR #570, opened 2026-09-02
fix(map): stop blaming the language when the ingest was deduplicated

KageBinary:
Addresses the CLI half of #527. The write that gets swallowed is a backend bug — ix-infrastructure/Ix-memory#177 fixes that — but what `ix map` *says* about it is ours, and today it points at the wrong thing.

## Problem

`describeEmptyCompletedMap` has one hypothesis for a completed-but-empty map and states it as a conclusion:

> Backend reported full_local_completed, but mapped 0 files after local ingest found 1 supported source file (1 patch committed). **The source graph was ingested**, but no architecture hierarchy was created. **The active backend may not map this source language yet.**

For #527 both bolded claims are false. The backend answered the commit `Idempotent` — it still held the patch record for a graph that had been deleted underneath it — so the run wrote nothing. The TypeScript parsed fine; it never reached the graph to be mapped. A reader who trusts this message goes looking at language support, which is the one thing that was never involved.

## Why the CLI can tell the difference

It already has the answer and throws it away. `PatchCommitResult.status` is on the wire (`CommitStatus.toString` from `PatchCommitRoutes` / `BulkPatchCommitRoutes`), and both commit sites in `ingest.ts` read `result` — they just count every answer as applied:

```ts
latestRev = advanceRev(latestRev, result.rev);
patchesApplied++;                      // Ok and Idempotent, indistinguishable
```

So `ingestFiles` now also counts `idempotentPatches`, and `describeEmptyCompletedMap` uses it. One status covers a whole bulk chunk — the backend commits it as a unit — so `Idempotent` there means every patch in the chunk was a replay.

## The new message

```
Error: Backend reported full_local_completed, but mapped 0 files after local ingest
found 1 supported source file. The backend answered every one of the 1 committed
patch with 'already applied', so this run wrote nothing — the graph is empty because
the commits were deduplicated against patch records the backend still holds, not
because the source could not be parsed. That is what a workspace whose graph was
deleted without its patch records looks like; the records expire within 24 hours, or
an upgraded backend clears them with the reset. The source ingest baseline was
preserved, so the next 'ix map' can reuse unchanged files.
```

The 24h line is a real workaround, not filler: `idempotency_keys` carries a TTL index on `created_at`, so the same command works the next day with no intervention.

## What does not change

- Exit code stays 1 and the architecture baseline is still cleared — it is still a failure, just a different one.
- A **partial** replay (`idempotentPatches < patchesApplied`) keeps the existing wording; only "every commit was a replay" is unambiguous enough to assert.
- `patchesApplied === 0` — the hash-skip path — keeps it too, and is guarded explicitly because `idempotentPatches >= patchesApplied` is trivially true at zero.
- A `--full` re-map of an intact graph deduplicates every commit as well, but it maps files, so `file_count !== 0` returns before any of this. Checked live, not just reasoned about (below).

`idempotentPatches` also joins `ix ingest --format json`, beside `commitErrors`.

## Verification

`ix` built from this branch, memory layer run from source at `Ix-memory@main` (i.e. **without** the backend fix), following #527's steps.

```
run 1                                    map: 1 files
POST /v1/reset/workspace                 {"ok":true,"message":"Workspace data deleted."}
db after reset                           {nodes: 0, keys: 1}      ← the stale record
re-map                                   Error: ...answered every one of the 1 committed
                                         patch with 'already applied'...
ix ingest --format json                  patchesApplied 1, idempotentPatches 1, commitErrors 0
```

Regression case — full re-map of an **intact** graph, baseline cleared so every file is re-submitted and every commit deduplicates:

```json
{ "file_count": 1, "region_count": 0, "outcome": "full_local_completed",
  "parse_errors": 0, "commit_errors": 0 }        exit 0
```

Gate: **1512 passed / 2 skipped** (89 files), parser smoke passed, `typecheck` clean, `lint` 0 errors (41 pre-existing warnings).

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

### DIFF
--- a/ix-cli/src/cli/commands/ingest.ts
+++ b/ix-cli/src/cli/commands/ingest.ts
@@ -912,7 +912,6 @@ export type CommitOutcome =
   | { kind: "warn"; message: string }
   | { kind: "fatal"; message: string };
 
-/** Minimal local-ingest facts needed by commands that continue after ingestion. */
 /**
  * The wire value the backend sends for a commit it short-circuited because it
  * had already recorded the patch id — `CommitStatus.Idempotent.toString` in
@@ -920,6 +919,15 @@ export type CommitOutcome =
  */
 const COMMIT_STATUS_IDEMPOTENT = 'Idempotent';
 
+/**
+ * The wire value for a commit the backend refused because the optimistic
+ * base-rev check lost a race — `CommitStatus.BaseRevMismatch`. It writes
+ * nothing, exactly like `Idempotent`, but unlike `Idempotent` it means the
+ * work still needs doing.
+ */
+const COMMIT_STATUS_BASE_REV_MISMATCH = 'BaseRevMismatch';
+
+/** Minimal local-ingest facts needed by commands that continue after ingestion. */
 export interface IngestFilesSummary {
   filesDiscovered: number;
   patchesApplied: number;
@@ -927,13 +935,40 @@ export interface IngestFilesSummary {
    * How many of `patchesApplied` the backend answered `Idempotent` — it had
    * already recorded that patch id, so the commit wrote nothing.
    *
-   * Normally uninteresting: `--full` re-submits every file, so an unchanged
-   * repo legitimately deduplicates everything against a graph that is already
-   * correct. It matters only when the graph turns out to be EMPTY afterwards,
-   * because then the deduplication is the reason nothing was written rather
-   * than a sign that nothing needed to be (#527).
+   * Normally uninteresting: `ix ingest --force` re-submits every file (and a
+   * lost or blank local baseline has the same effect against a backend that
+   * still holds the patch records), so a repo whose graph is already correct
+   * legitimately deduplicates everything. Note `ix map --full` is NOT such a
+   * case — it forwards `full` to the backend map and never sets `force` on the
+   * ingest below it, so it re-submits nothing.
+   *
+   * It matters only when the graph turns out to be EMPTY afterwards, because
+   * then the deduplication is the reason nothing was written rather than a
+   * sign that nothing needed to be (#527).
    */
   idempotentPatches: number;
+  /**
+   * Files this run skipped because they were already unchanged — mtime-clean
+   * against the local baseline, or hash-clean against the backend.
+   *
+   * Load-bearing for the #527 diagnosis, not informational. `idempotentPatches`
+   * only describes the patches a run actually submitted; it says nothing about
+   * the files the run never submitted, and those are what separate a graph
+   * deleted under its own patch records from a graph that is simply fine. A run
+   * in the #527 state skips nothing here: the DB-reset guard clears the mtime
+   * cache and the hash lookup comes back empty, so every file is re-submitted
+   * and every one is answered `Idempotent`. A run that skipped even one file as
+   * unchanged has proof the backend still holds that file's source hash, and
+   * therefore its nodes.
+   *
+   * Counts only the unchanged skips — empty, oversized and minified files are
+   * skipped for reasons that carry no such proof. It is the same counter the
+   * stitch-completeness gate reads, which also means it inherits that counter's
+   * Path-B reset: a run with no baseline re-reads and re-hashes every file, so
+   * the stat loop's mtime skips did not actually happen and are subtracted
+   * back out. That is the shape a #527 run has, and it is exactly right here.
+   */
+  filesSkippedAsUnchanged: number;
   parseErrors: number;
   commitErrors: number;
   stitchErrors: number;
@@ -2124,6 +2159,12 @@ export async function ingestFiles(
                 recordMs(chunkMs);
                 timings.fallbackCommitMs += chunkMs;
                 latestRev = advanceRev(latestRev, result.rev);
+                // See onBulkCommitted: a lost base-rev race wrote nothing and
+                // must not be reported as applied.
+                if (result.status === COMMIT_STATUS_BASE_REV_MISMATCH) {
+                  commitErrors++;
+                  continue;
+                }
                 patchesApplied++;
                 patchesTheBackendTook++;
                 commitBreaker.recordSuccess();
@@ -2441,6 +2482,21 @@ export async function ingestFiles(
             },
             onBulkCommitted: (items, result) => {
               latestRev = advanceRev(latestRev, result.rev);
+              // `BaseRevMismatch` wrote nothing: the backend read the latest rev
+              // outside the transaction and it moved before the commit ran. It
+              // is a lost race, not a replay, so it must not count as applied —
+              // and it belongs in commitErrors, because that is what stops
+              // `persistIngestBaselineIfClean` caching these files as unchanged
+              // and leaving them missing from the graph until the next edit.
+              if (result.status === COMMIT_STATUS_BASE_REV_MISMATCH) {
+                commitErrors += items.length;
+                if (debug) {
+                  process.stderr.write(
+                    `\n  [commit lost the base-rev race] ${items.length} patches wrote nothing; they will be re-sent next run\n`
+                  );
+                }
+                return;
+              }
               patchesApplied += items.length;
               patchesTheBackendTook += items.length;
               commitBreaker.recordSuccess();
@@ -3359,6 +3415,7 @@ export async function ingestFiles(
     filesDiscovered,
     patchesApplied,
     idempotentPatches,
+    filesSkippedAsUnchanged,
     // `+ crashedParses()`, as the baseline and delete guards already do. Files
     // lost to a dead parse pool raise `filesSkippedUnparsed`, never
     // `parseErrors`, so without this everything downstream read the run as
--- a/ix-cli/src/cli/commands/ingest.ts
+++ b/ix-cli/src/cli/commands/ingest.ts
@@ -913,9 +913,27 @@ export type CommitOutcome =
   | { kind: "fatal"; message: string };
 
 /** Minimal local-ingest facts needed by commands that continue after ingestion. */
+/**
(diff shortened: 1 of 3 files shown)


KageBinary:
Merge-order note, since three PRs are in `ingest.ts` at once. I test-merged this branch against each:

| against | result |
|---|---|
| #571 `fix/568-stitch-single-flight` | clean |
| #569 `fix/567-watch-dedup-temp-dir` | clean |
| #572 `fix/560-commit-failure-breaker` | **conflict**, `ix-cli/src/cli/commands/ingest.ts` |

**The conflict is textual, not semantic.** Both hunks are "both sides added a line at the same spot" — #572 adds `commitBreaker.recordSuccess()` and this PR adds the `idempotentPatches` increment, at the per-file commit and at `onBulkCommitted`. The resolution is to keep both, in either order:

```ts
                patchesApplied++;
                commitBreaker.recordSuccess();
                if (result.status === COMMIT_STATUS_IDEMPOTENT) idempotentPatches++;
```

Resolved that way locally: **typecheck exit 0, 90 files, 1541 passed, 2 skipped**, parser smoke passed.

Worth stating explicitly, because the two do touch the same counters: #572 adds a path that raises `patchesApplied` **without** a commit status to observe — the breaker's replay arm, `patchesApplied += items.length - itemIndex`, counting patches the server has confirmed it already holds. Those are genuinely applied and genuinely not observed as `Idempotent`, so after both land `idempotentPatches` can only fall further below `patchesApplied`. This PR's new diagnosis is gated on `idempotentPatches >= patchesApplied`, so the composition can only make it *less* likely to fire — it cannot produce a false "the ingest was deduplicated" claim on a run that was actually cut off. The safe direction, but not obvious from either diff alone.

#572's `onAbandoned` raises `commitErrors`, which `describeEmptyCompletedMap` already bails on before reaching any of this.

Happy either way on order. If #572 lands first I will rebase this and re-run; if this lands first the same two-line resolution applies there.

One flake seen while checking: the first of three full runs had a single unrelated failure and took 242s against ~100s for the other two, which is the shape of the order-dependent failure in #567. The other two runs were clean.

ix-cli/src/cli/commands/map.ts
```diff
@@ -161,7 +166,23 @@ export function describeEmptyCompletedMap(
...
+
+  // The backend deduplicated every commit, so this run wrote nothing at all —
+  // and the two sentences below are then both false: nothing was ingested, and
+  // the language had no chance to be the reason (#527). The observed cause is a
+  // graph deleted out from under the backend's own patch records: a scoped
+  // reset used to remove a workspace's nodes and patches while leaving its
+  // idempotency keys, so the identical re-ingest was accepted as a replay and
+  // committed nothing (Ix-memory#174). Those keys expire after 24h, which is
+  // why waiting is a real workaround and worth saying out loud.
+  if (patches > 0 && ingest.idempotentPatches >= patches) {
```
josephismikhail: `idempotentPatches >= patches` only proves that every patch *this run submitted* was a replay — it says nothing about the files the run never submitted, and those are the ones that disprove the new diagnosis.

Concrete case: a workspace on a language the backend cannot build a hierarchy for (the exact case the old wording was written for). 99 files are mtime/hash-clean and are skipped — which is itself proof the backend still holds their source hashes, and so their nodes. The user then reverts one file with `git checkout` to content that was ingested earlier, so its `computePatchId(filePath, sourceHash, extractor)` matches a patch record that is still there and `/v1/patch` answers `Idempotent`. Now `patchesApplied === 1`, `idempotentPatches === 1`, `file_count === 0`, and `ix map` prints "the graph is empty because the commits were deduplicated ... the records expire within 24 hours". The graph is not empty, waiting 24 hours changes nothing, and the true cause is the sentence this branch removed.

The distinguishing signal exists but is not on `IngestFilesSummary`: a run whose graph really was deleted under its patch records skips nothing as unchanged (the DB-reset guard clears the mtime cache and the hash lookup comes back empty), so an unchanged-skip count of 0 separates #527 from the case above. Worth carrying that on the summary and into this condition before asserting the cause.

ix-cli/src/cli/commands/ingest.ts
```diff
@@ -1587,6 +1607,9 @@ export async function ingestFiles(
             onBulkCommitted: (items, result) => {
               latestRev = advanceRev(latestRev, result.rev);
               patchesApplied += items.length;
+              // One status covers the whole bulk: the backend commits the chunk
+              // as a unit, so `Idempotent` means every patch in it was a replay.
+              if (result.status === COMMIT_STATUS_IDEMPOTENT) idempotentPatches += items.length;
```
josephismikhail, line 2505: `Idempotent` is not the only status that is counted as applied while writing nothing. Both backend handlers return `BaseRevMismatch` and write nothing when the optimistic base-rev check loses a race (`BulkPatchCommitRoutes` reads `queryApi.getLatestRev` outside the transaction and `commitPlan` compares it against `loadLatestRev(txId)` inside it; `PatchCommitRoutes` has the same guard), and the CLI still does `patchesApplied += items.length` and `advanceRev(latestRev, result.rev)` for it.

So a run whose chunk lost that race gets `idempotentPatches === 0`, falls into the language branch, and is told the source graph was ingested when nothing was written. Worse than the message: the mtime cache is persisted at the end of the run, so the next `ix map` skips those files as unchanged and they stay missing from the graph until they change again or `--force` is passed.

Pre-existing, but this is the line that decides which answers count as a write — treating `Ok` as the only applied status (or counting non-`Ok` into `commitErrors`) would close it here rather than leaving one half fixed.

ix-cli/src/cli/commands/ingest.ts
```diff
@@ -806,9 +806,27 @@ export type CommitOutcome =
   | { kind: "fatal"; message: string };
 
 /** Minimal local-ingest facts needed by commands that continue after ingestion. */
+/**
+ * The wire value the backend sends for a commit it short-circuited because it
+ * had already recorded the patch id — `CommitStatus.Idempotent.toString` in
+ * `PatchCommitRoutes` / `BulkPatchCommitRoutes`.
+ */
+const COMMIT_STATUS_IDEMPOTENT = 'Idempotent';
```
josephismikhail, line 920: The const was inserted between `IngestFilesSummary` and its doc comment, so both hovers are now wrong: `/** Minimal local-ingest facts needed by commands that continue after ingestion. */` documents `COMMIT_STATUS_IDEMPOTENT`, and the wire-value comment documents the interface. Move the new const above that line (or below the interface).

ix-cli/src/cli/commands/ingest.ts
```diff
@@ -806,9 +806,27 @@ export type CommitOutcome =
...
+
 export interface IngestFilesSummary {
   filesDiscovered: number;
   patchesApplied: number;
+  /**
+   * How many of `patchesApplied` the backend answered `Idempotent` — it had
+   * already recorded that patch id, so the commit wrote nothing.
+   *
+   * Normally uninteresting: `--full` re-submits every file, so an unchanged
```
josephismikhail: "`--full` re-submits every file" does not describe any path in this tree. `ix map --full` only forwards `full` to `client.map(...)`; the `ingestFiles(cwd, ...)` call above it never passes `force`, so `--full` cannot re-submit anything. The flag that re-submits is `ix ingest --force` (`opts.force` clears the mtime cache and widens `hashLookupPaths`), and `ix map` never sets it.

Same claim is load-bearing in the PR description ("a `--full` re-map of an intact graph deduplicates every commit as well, but it maps files, so `file_count !== 0` returns before any of this") — the scenario it reasons about cannot occur, which means the `file_count` early return is not actually what protects the intact-graph case. Worth restating on the mechanism that is real (`ix ingest --force`, or a lost/blank local baseline against a backend that still has the patch records).

KageBinary:
Rebased onto \`main\` — #571 and #572 landed and both touched the commit loop this depends on. Three things changed in the resolution, all of them worth naming because two are not mechanical:

**1. The cutoff drain is a third commit site.** #571 added a bulk commit inside the held-patch drain (`ix-cli/src/cli/commands/ingest.ts`, the `[cutoff] bulk placed` path) that did not exist when `idempotentPatches` was written. It carries a status like every other bulk, so it now counts:

```ts
if (result.status === COMMIT_STATUS_IDEMPOTENT) idempotentPatches += chunk.length;
```

Without it a run whose held patches were all placed there and all deduplicated would report `idempotentPatches: 0` and fall straight back to the language hypothesis this PR exists to retire. The 409-replay sites next to it are deliberately **not** counted — those patches are confirmed-landed by the server's own error body, which is a different fact from the backend answering `Idempotent`, and conflating them would let a partial-bulk failure produce a false #527 diagnosis.

**2. `unchangedSkipped` is gone; the guard reads `filesSkippedAsUnchanged` instead.** #571 landed a counter incremented at exactly the three unchanged-skip sites mine used, so mine was a duplicate and is deleted. This is not a rename for tidiness — main's counter is strictly more correct, because it has a reset mine did not:

```ts
filesSkipped -= filesSkippedAsUnchanged;
filesSkippedAsUnchanged = 0;
```

on the no-baseline / `--force` path, where the stat loop's mtime skips never actually happen because every file is re-read and re-hashed anyway. That is the shape a real #527 run has — the DB-reset guard clears the mtime cache — so the guard now zeroes out phantom skips that would otherwise have suppressed the correct diagnosis. The summary field is renamed to match the counter.

**3. `parseErrors: parseErrors + crashedParses()`** from main is kept as-is; the summary just gains the extra field beside it.

Gates on the rebased head: **1650 passed / 2 skipped** (92 files), parser smoke passed, `typecheck` clean, `lint` 0 errors (41 pre-existing warnings), `knip` clean.

Issue #568, opened 2026-09-02
[bug] Stitch AQL stacks in Arango after HTTP timeout and pegs CPU


helgetan:
## Summary

`POST /v1/stitch` runs an unbounded cross-workspace AQL join. The HTTP call times out at ~60s (500), but the query keeps running. Hooks then retry (`GET /v1/stitch/system/…` → `POST /v1/map` → `POST /v1/stitch`) and start more copies.

Arango stays at 800–1200% CPU. Memory-layer stays idle (~1%). Restart only resets the pile-up.

Related: #560 looks like a downstream timeout once Arango is saturated. #290 bounded the client wait, but does not cancel or single-flight the server-side query.

## Environment

- macOS, Docker Desktop
- Ix CLI 0.10.5
- Memory Layer 1.0.28
- ArangoDB 3.12.11, db `ix_memory`
- Large TypeScript monorepo 1.6M LoC, graph already populated

## What we see

- 10–16 identical queries in `/_api/query/current`, each 2–38 min
- `POST /v1/stitch` → 500 after ~60s
- then a new map + stitch
- later `/_api/cursor` and `/_api/transaction/begin` also time out
- one finished query logged as slow query at 2229s

## The query
`FOR sc IN symbol_consumes LET providerWss = (sc.pkg != null ? (FOR p IN provides FILTER p.name == sc.pkg AND p.workspace_id != sc.workspace_id RETURN DISTINCT p.workspace_id) : (FOR c IN consumes FILTER c.workspace_id == sc.workspace_id FOR p IN provides FILTER p.name == c.name AND p.workspace_id != sc.workspace_id RETURN DISTINCT p.workspace_id)) LET matches = (FOR ex IN exports FILTER ex.name == sc.symbol AND ex.workspace_id IN providerWss RETURN ex) FILTER LENGTH(matches) == 1 RETURN { src: sc.caller_node_id, dst: matches[0].node_id, name: sc.symbol }`

Nested scan over `symbol_consumes × provides × consumes × exports`. CPU-bound, ~2MB peak memory.

## Repro

1. Index a large multi-workspace repo.
2. Leave auto-map/stitch hooks on, or call `POST /v1/stitch` a few times.
3. Check `GET /_db/ix_memory/_api/query/current` and `docker stats`.

First stitch fails at 60s, query stays `executing`, next hook starts another copy.

## Expected

- One stitch job at a time
- Client timeout cancels the AQL
- No retry while the previous job is still running
[...]

PR #571, opened 2026-09-02
fix(ingest): stop ix map piling stitch queries onto a backend that is still running the last one

josephismikhail:
Fixes the client-side half of #568.

`POST /v1/stitch` runs an unbounded cross-workspace AQL join. The HTTP call times out and returns 500 at ~60s; **ArangoDB keeps executing the query**. The CLI treated "that request failed" as "that work stopped" and started another copy on the next map — 10–16 identical queries, 2–38 minutes each, Arango at 800–1200% CPU.

---

## Reproduction, on the real backend

`docker-compose.standalone.yml` — `ghcr.io/ix-infrastructure/ix-memory-layer:latest` against **ArangoDB 3.12.10-1 enterprise**. A counting proxy in front of `:8090` times the client out at **60s**, matching the report, and deliberately does **not** abort upstream, so the backend keeps working on a request the client has been told failed.

**Sequential maps — this is `main`:**

```console
19:53:50 REQ POST /v1/stitch #1
19:54:50 TIMEOUT #1 -> 500 after 60015ms (upstream still working)
19:54:54 REQ POST /v1/stitch #2      ← 4s after the last one was 500'd
19:55:54 TIMEOUT #2 -> 500 after 60013ms
19:55:58 REQ POST /v1/stitch #3
19:56:58 TIMEOUT #3 -> 500 after 60011ms
```

Three maps, three queries, each started while the previous join was still running. No concurrency needed: a hook fires one `ix map` per change, each completes, releases its workspace lock, and the next starts a fresh stitch.

**Concurrent maps:** `ix map` single-flights per *workspace*; the stitch is *cross-workspace*, so two maps of two different workspaces hold two different locks and stitch together.

---

## The fix

Admission control keyed on the backend endpoint (`ix-cli/src/cli/stitch-guard.ts`):

| Rule | Behaviour |
|---|---|
| One at a time per endpoint | A second `ix map` — including one for a different workspace — waits up to `IX_STITCH_WAIT_MS` (30s) for the in-flight stitch, then skips. |
| A cooldown written when a stitch **starts** | Removed only on proof nothing is running. Until then no further stitch goes to that endpoint for `IX_STITCH_COOLDOWN_MS` (15 min), measured from when the attempt ended. |

### Why the cooldown is written up front

This is the part worth reviewing, because it is the opposite of the obvious design and it is what five earlier revisions got wrong.

The obvious design inspects the **failure** and writes a cooldown when it looks like a timeout. That asks "did the backend start the join?" of an error, and the evidence keeps not being there:

* the message is whatever proxy sits in front (a 500 here, a 504 with an HTML body in #528), so text matching is inert where it is needed;
* elapsed time cannot separate a slow join from a slow **upload**, and the stitch payload is megabytes on a large monorepo;
* "the client aborted" cannot separate a mid-flight hang-up from a run deadline that fired before the request was serialized;
* and none of it survives the process being killed, which is the ordinary end of a hook that timed out.

Writing the marker before the request needs none of those answers. Every ending without a definite refusal — timeout, abort, deadline, SIGTERM, power cut — leaves it where it was written, because leaving it is the default rather than something a dying process must still do.

It is removed on **proof**, of which there are four kinds:

* a success;
* a **4xx** except 408 — the backend refusing rather than executing. 408 is a proxy saying *it* gave up, which is this bug;
* **501**, which is how `isStitchUnsupported` already spells "no `/v1/stitch` here";
* **a connection that was never established**, so no bytes reached the backend.

The last is narrower than "a transport error", deliberately, and the codes are **derived rather than guessed** — driven through `IxClient.stitch` itself, which flattens HTTP errors into `"NNN: ..."` strings but rethrows transport failures intact:

| what happened | `cause.code` | proof? |
|---|---|---|
| closed port | `ECONNREFUSED` | yes |
| unresolvable host | `ENOTFOUND` | yes |
| socket dropped **after** the request went out | `UND_ERR_SOCKET` | **no** |
| port on the WHATWG blocked list | *(none)* | no |

The third row is why the rule is not simply "a transport error": an upstream that restarted killed its join, a proxy that hung up did not, and both spell the same code. It keeps the marker. The tests drive real errors out of a real closed port, a real bad hostname and a real server that resets after reading the request — a classifier keyed on a shape nobody verified is the failure mode this file has already hit twice.

The cost is that a stitch failing for an unclassified reason cools down when it need not have. That errs toward skipping one stitch rather than stacking joins on a database that is already the reason.

### Why the cooldown is re-stamped when the attempt ends

The marker goes down at the start and is **re-stamped to the end time** when the attempt reports back without proving anything stopped. Both halves matter. The start-time write is what survives a kill, and is the whole reason it goes down before the request. But leaving it *at* the start time made every `IX_STITCH_COOLDOWN_MS` shorter than the attempt itself already expired by the time the attempt ended — and `IxClient.post` caps a request at two minutes, so on the timeout path every setting under that silently bought nothing: the next map was admitted immediately and stacked a second join, which is the bug.

One residue remains, and is documented: a process that is *killed* never re-stamps, so a sub-two-minute cooldown is expired when the next map looks at it.

### Results, on the real backend

| | before | after |
|---|---|---|
| 3 sequential maps, 60s ingress | 3 stitches | **1** |
| 3 concurrent maps, slow backend | 3 stitches | **1** |
| 3 concurrent maps, **healthy** backend | 1 (2 registrations lost) | **3** |
| `ix map` killed mid-stitch, then re-run | 2 stitches | **1** |
| run deadline fires 18s into a stitch, then re-run | 2 stitches | **1** |
| backend down, connection refused | (round 8) 15-min cooldown | **no cooldown** |
| healthy backend | works | works, unchanged |

The last two are the cases the classifier could not reach. The deadline one lands at **18s** — under the 20s threshold the elapsed-based rule used, which is exactly why it missed it. The killed-process one is verified by killing the process for real; the lock goes, the marker stays, and the next map is refused with a distinct message: *"a stitch was started and never reported back"*.

---

## Review history

Twenty-three rounds. The findings are worth reading as a record of how the design got here — several rounds fixed a defect and introduced another, and the redesign in `12e0d33` was the response to that pattern rather than to any single finding.

<details>
<summary>Rounds 1–5: the classifier, and why it was replaced</summary>

**Round 1** (5): `IX_STITCH_COOLDOWN_MS=0` didn't release an active cooldown though the message said it did; an abort from an already-expired deadline wrote a cooldown for a request that never left; `settle` could throw and replace the caller's error; the lock was keyed on the raw endpoint string (three locks for one backend); the Note claimed the next map re-registers, which was never true.

**Round 2** (7): single-flight *shed* stitches rather than waiting, losing a workspace's registration on healthy backends; the elapsed rule cooled down on a slow 4xx; the skip was invisible to machine consumers; `IX_STITCH_SLOW_FAILURE_MS=0` meant the opposite of off; an abort at ~0 elapsed still cooled down; one remedy for two different rules; a comment resting on a false premise.

**Round 3** (7): `isAbortError`'s message fallback read a *backend* error saying "aborted" as a client abort; the contended remedy was wrong; the waiter ignored the run deadline; `stitch_skipped` landed where the hooks don't look; the 4xx short-circuit swallowed 408; two comments still carried a disproved claim; the code branched on prose.

**Round 4** (4): the cooldown was read before the lock, leaving a stale-read race; `stitch_skipped` still missed `--silent`; the 5ms floor was measured across `JSON.stringify` of a megabyte payload; `describeStitchSkipped` had no test.

**Round 5** (6, two MEDIUM): removing the abort arm reopened the deadline path; a SIGTERM released the lock and wrote no cooldown. Those two, plus the fact that the abort rule had been added, floored, removed and then needed restoring, are why the classifier was replaced rather than adjusted a fifth time.
</details>

<details>
<summary>Rounds 6–8: the redesign, and one retraction</summary>

**Round 6** (3): 501 wasn't accepted as proof; the refusal said "another ix map" when only `ix map` takes the workspace lock; `ix ingest --format json` dropped the field when undefined. A fourth — that default ports key differently — was a **false positive**: the WHATWG parser already strips them, mutation testing showed the "fix" was dead code, and it was reverted with the reason recorded so it is not re-fixed.

**Round 7** (3): the wait loop checked the deadline only *after* admission, so a holder releasing mid-wait got admitted and marked the backend for 15 minutes over a request that could not leave; a deadline firing mid-wait was reported as `in-flight`; the short Note promised a "next attempt" that never comes.

**Round 8** (4): **one of my fixes was wrong and is reverted in `d816a03`.** I held the mtime baseline back when a stitch was refused, claiming it let the next map re-register. It does not: `loadExistingHashes` still returns the previous run's hashes, the hash-clean branch increments `filesSkipped`, and the `filesSkipped === 0` gate is never entered — so the next map re-reads and re-hashes the whole repo, makes a full source-hash round trip to the already-degraded backend, and still skips the stitch. My verification was wrong, not just incomplete: I saw "Source hash lookup" instead of "mtime-clean" and read that as success, without checking the run reached the stitch block. It had also broken `persistCompletedMapBaseline`, so a first map during a cooldown left `ix status` reporting `map_complete=false`.

Also fixed: the default skip message told a *contended* run to force a full re-ingest, which the verbose form of the same function deliberately hedges against — and `--verbose` is off by default, so the wrong advice was the one almost everyone saw. And `IX_MAP_LOCK_MAX_MS`, which the stitch lock silently inherits, is now in the env table.

**Round 9** (4): a connection that was never established was not accepted as proof, so a backend that went down mid-run took a full 15-minute endpoint-wide cooldown for a request it never received; the cooldown was stamped at the attempt's start and never moved, so any value under the two-minute request cap was already expired the moment a timing-out stitch returned; and twelve inline `expect(admitStitch(x).admitted)` assertions dropped a granted admission without settling it, leaking its `exit`/`SIGINT`/`SIGTERM` listeners until the file printed `MaxListenersExceededWarning` three times. The fourth was the baseline-hold limitation below, re-raised and again not fixed — for the reasons the round-8 revert records.

Those three are mutation-validated: dropping the `neverConnected` arm, re-stamping back to the start time, and accepting `UND_ERR_SOCKET` each turn exactly one test red.

**Round 10** (1, and two counting bugs it exposed): the skip message tells the user to recover with `ix ingest <root> --force`, and that promise could not be kept. The stitch block is gated on `filesSkipped === 0`, but that total also counts zero-byte files, files that look minified, and files the parser returned nothing for — and under `--force` those are the *only* increments that can still fire, since every mtime and hash skip is gated on `!opts.force`. One empty `__init__.py` disqualified a repo from stitching permanently.

Reproduced against a stub that logs stitch arrivals, on a three-file repo whose third file is an empty `__init__.py`, with `--force`: **0** stitch requests before, **1** after (`exports=2 -> 3 edges`).

The gate now reads a narrower `filesSkippedAsUnchanged`, incremented only where we assumed a file had not changed. The distinction is not cosmetic: an empty, minified or unparseable file has no entities to contribute under *any* run, so the collected registration set is complete without it — whereas an mtime- or hash-clean skip means the file has content we would have indexed and did not look at.

Two pre-existing counting bugs surfaced while verifying it. `skipReasons.unchanged` in `--format json` reported the whole skip total, so a consumer could not tell an incremental no-op from a repo full of empty `__init__.py`; and `skipReasons.emptyFile` was hardcoded to `0`. With it reporting for the first time it came back as **2** for a repo with **one** empty file — the unconditional stat loop and the `--force` ingest path both walk `filePaths`, so every empty or oversized file was counted twice, and `filesSkipped` has always been wrong that way on that path. Both fixed; the fstat check itself stays, since it is what makes the size cap TOCTOU-free.
</details>

<details>
<summary>Rounds 11–12: a regression I introduced, and every remedy made true</summary>

**Round 11** (3): round 10's narrowed gate was **too** narrow. It let a file the parser returned nothing for through, and `ParsePool` resolves `null` both for a reported parse failure and for the in-flight task of a worker that **crashed** — neither of which raises `parseErrors`. So a `--force` run where one worker OOMs would have stitched a registration missing that file's exports over the complete one, shrinking the cross-repo graph on a transient failure. Verified by forcing one file's parse result to null on the `--force` path: **0** stitch requests with a crashed worker, **1** without, and the empty-`__init__.py` case still stitches. The difference is whether the skip is a property of the *file*: empty and minified are decided by content under a rule that does not vary between runs; a parse-null is not.

Also: with `unchanged` narrowed, the pool's nulls landed in no `skipReasons` bucket, so the breakdown stopped reconciling with `filesSkipped`. `parseError` now carries both.

And documented, not fixed: the lock and cooldown are keyed on the **endpoint**, so one transient failure while mapping repo A refuses the stitch for every other repo on that backend for the cooldown. That blast radius is inherent — the query being bounded is cross-workspace, so a per-workspace key bounds nothing — but the docs named only the non-re-attempt, not the scope.

**Round 12** (7): the short remedies for `in-flight` and `deadline` were the same false promise round 10 removed from the cooling branch, and `--verbose` is off by default so they were the advice almost everyone saw. All three branches now name `--force`; the plain re-run is still offered first on contention, because the other run may well be registering this same workspace. `ix reset` cleared the mtime and stitch-scope caches but not the cooldown, so a reset during one wiped the registration and then had the only run that could restore it refused. Round 10's double-count fix left a file that changes between the two stats in no counter at all. Three stale or backwards comments.
</details>

<details>
<summary>Rounds 13–23: the connect classifier, the reporting contract, and a hang I shipped</summary>

**Rounds 13–15** (7). `connectionNeverEstablished` was an errno list, so a VPN drop or an unreachable host still cost a fifteen-minute endpoint-wide cooldown for a request that never left the machine — the exact harm round 9 added the arm to prevent, for a neighbouring code. It is keyed on Node's `syscall` now (`connect` / `getaddrinfo`), which names the *phase* rather than enumerating outcomes, with the errno set kept alongside and a walk into `AggregateError.errors` — because `localhost`, the default endpoint, is multi-address and reports a refusal that way, which a syscall-only rule missed. Every shape in the table above was driven through `IxClient.stitch` and recorded.

Also: the `unparsed` bucket briefly went into the *backend's* contract when the change is client-side; a 2xx with an unreadable body was accepted as proof and then, correctly, un-accepted (see below); and a failed `rmSync` of the cooldown after a successful stitch was swallowed — the one failure here that fails *closed*, blocking the endpoint for fifteen minutes silently.

**Rounds 16–18** (9). Round 11's gate was too narrow *and* too wide. Too narrow: it kept parse-nulls out, and `parseFile` returns null deterministically when a grammar is an unavailable optional dependency — `tree-sitter-sas` has no win32 prebuild — so one such file blocked stitching forever, `--force` included. Too wide: a *crashed* parse worker is a real loss, and `ParsePool` collapsed both into the same `null`. It reports crashed tasks separately now: **0 stitches** with a simulated crash, **1** with a `parseFile` null, and the empty-`__init__.py` case still works.

And Path B — which re-parses everything — still carried the stat loop's mtime skips into the gate, so "add one new file to a mapped repo", the commonest incremental shape there is, silently declined to stitch.

**Rounds 19–23** (10). The 2xx arm is reverted: a proxy answering 200 with its own HTML error page produces exactly that shape (it is the #528 pattern), and clearing the marker there admits the next map onto a running join — the only proof arm that can cause #568 itself. `positiveEnvMs` used bare `parseInt`, so `IX_STITCH_COOLDOWN_MS=15m` silently meant fifteen *milliseconds*.

And a hang I shipped and then had to fix twice: adding a `'exit'` listener to the parse pool closed a real hole (a worker that dies without `'error'` strands its task forever) and opened a worse one — `destroy()` terminates every worker, each `'exit'` was read as a crash, replacements were spawned, and since a worker thread refs the event loop the CLI printed its summary and **hung forever**. Measured: exit 0 in 5s with the shutdown guard, timed out at 61s without. The first fix guarded only the `'exit'` route; `'error'` reaches the same code and hung the same way, so the guard moved into `onError`, and the respawn is capped.

Finally, the reporting contract: `stitch_skipped: null` means "the cross-repo edges are current", and it was null for the two commonest reasons a stitch does not happen — an incremental map with nothing complete to send, and a run with parse or commit errors. Both are reported now, with a machine-readable rule, and neither prints a human Note.
</details>

---

## Known limits

* **A skipped stitch is not re-attempted by an *incremental* map.** The block is gated on no file having been skipped as unchanged, and an incremental map that skips one never reaches it — nor does it have the registration data to send, having parsed only what changed. This predates the PR (it is equally true of a stitch that *failed*), but the guard makes the path more common. `ix ingest <root> --force` recovers it, and as of round 10 that is actually true rather than merely advertised. The round-8 revert is the record of an attempt to make it automatic that did not work.
* **The cooldown is measured from the stitch's start, not the join's real duration.** With the reported 2–38 minute joins and a 15-minute cooldown, a second copy can still be pushed onto a long-running first. This bounds the pile-up at ~2–3 rather than eliminating it; `IX_STITCH_COOLDOWN_MS` is tunable.
* **The lock fails open.** `acquireLockAt` treats any non-`EEXIST` error as acquired, deliberately, since it is shared with the map lock. An unwritable `~/.ix/locks` makes both rules inert; a failed cooldown write now warns on stderr, which `ix map --silent` discards.

## What this does not fix

#568 asks for four things. Two are here:

- [x] One stitch job at a time
- [x] No retry while the previous job is still running
- [ ] **Client timeout cancels the AQL** — needs the backend to bind the query to the request and kill it on disconnect.
- [ ] **Join indexed / incremental rather than a full scan** — backend-side.

A backend already carrying runaway queries stays saturated until they finish or it restarts; what changes is that the CLI stops adding to the pile. The two remaining items belong in the memory-layer repo.

## Verification

| check | result |
|---|---|
| `stitch-guard.test.ts` | 51/51 |
| `npx vitest run` (full) | **1543 passed, 3 failed, 21 skipped** |
| `ix ingest` exits (not hangs) | exit 0 in ~4s, 12 consecutive runs |
| empty-`__init__.py` stitch repro | 0 → **1** stitch request |
| crashed-worker stitch repro | **0** stitch requests (correctly refused) |
| `skipReasons` buckets vs `filesSkipped` | reconcile |

## One drive-by, unrelated to the guard

`stitch-scope-cache.test.ts` redirects `HOME` to a temp dir, but `config.ts` reaches the cache through `os.homedir()`, which on Windows reads `USERPROFILE` and ignores `HOME`. So on Windows the redirect was inert and these tests wrote `stitch_scope_*.json` into the developer's real `~/.ix`.

That made the file self-poisoning rather than merely untidy: "asks the backend on a miss and records the answer for next time" recorded `ws-fresh` in the real home on its first run and found it cached on every run after, so the assertion that the backend was consulted passed exactly once per machine and failed forever. The neighbouring "does not ask the backend" case passes either way, which is why it read as one flaky test rather than a broken fixture. CI never saw it because every job starts on a clean home.

Found while verifying this PR; it fails on `main`. Fixed in its own commit — running the file twice now gives 10/10 then 10/10, where it gave 10/10 then 9/10, and leaves nothing behind in the real `~/.ix`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,12 +14,27 @@ target/
 *.jar
 .DS_Store
 .worktrees/
-node_modules/
-ix-cli/node_modules/*
-core-ingestion/node_modules/*
-core-ingestion/dist/
-ix-cli/dist/
-ix-cli/coverage/
+# No trailing slash, deliberately: `node_modules/` matches directories only, so
+# it did not cover a *symlink* named node_modules -- which is how an absolute
+# symlink into a contributor's home directory came to be committed at
+# ix-cli/node_modules in #545 and stayed tracked until #567. Without the slash
+# the rule matches the entry whatever its type, at any depth, which also makes
+# the two package-specific lines that used to sit here redundant.
+#
+# The same applies to the build outputs below. A symlinked `dist` is how a git
+# worktree borrows a build from the main clone -- the same practice that
+# produced the node_modules symlink -- so a directory-only rule there leaves
+# exactly the defect above open under a different name.
+node_modules
+core-ingestion/dist
+ix-cli/dist
+ix-cli/coverage
+
+# watch-dedup.test.ts used to build a throwaway CLI into
+# ix-cli/.watch-child-runtime-*; it now builds inside node_modules/.cache,
+# which the rule above already covers. Kept so a directory left behind by an
+# interrupted run on a pre-#567 checkout does not surface as untracked work.
+ix-cli/.watch-child-runtime-*
 .vscode/
 .mcp.json
 
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -284,6 +284,7 @@ data to send, having only parsed what changed. A run that re-ingests every file
 | `IX_STITCH_COOLDOWN_MS` | `900000` | How long to hold off after a stitch that did not prove it stopped. `0` disables the cooldown; single-flight stays. |
 | `IX_STITCH_WAIT_MS` | `30000` | How long to wait for an in-flight stitch before skipping. `0` sheds immediately. |
 | `IX_LOCK_DIR` | `~/.ix/locks` | Where the stitch lock and cooldown record live (shared with the map lock). |
+| `IX_MAP_LOCK_MAX_MS` | `1200000` | Shared with the map lock: how old a held lock must be before it is presumed abandoned and stolen. Lowering it to a few seconds so a wedged `ix map` self-heals faster also lets a second process steal the stitch lock from an in-flight stitch. The cooldown normally catches that on the next read, so it only matters together with `IX_STITCH_COOLDOWN_MS=0` — which is the one case where "single-flight stays" stops being true. |
 
 This bounds the client. Cancelling the server-side query when the client hangs
 up, and making the join indexed rather than a full scan, are backend concerns
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -240,10 +240,9 @@ retried exactly as before. Nor does **any 4xx**, however long it took to arrive
 spend tens of seconds there before a 413 comes back; a refused request is
 decisive evidence that no join started — with **408** as the exception, since a
 proxy reporting that *it* gave up waiting says nothing about whether the backend
-did, and that is the whole bug. An abort raised by a run deadline that
-had *already* expired sets no cooldown either — `fetch` rejects on an
-already-aborted signal without contacting the backend, so there is no join to
-wait for.
+did, and that is the whole bug. An abort has no rule of its own: one that reached
+the backend passes the elapsed test anyway, and one raised before the request
+left (a run deadline expiring during serialization) correctly does not.
 
 Both the lock and the cooldown are keyed on a normalised endpoint, so
 `http://localhost:8090`, `http://localhost:8090/` and `http://127.0.0.1:8090`
@@ -256,7 +255,7 @@ already on disk, so setting it to `0` releases an active one rather than only
 affecting the next.
 
 A skipped stitch is reported as `stitchSkipped` in `ix ingest --format json`,
-as `stitch_skipped` in `ix map --format json`, and in the ingest summary, so an
+as `stitch_skipped` in `ix map --format json` and `--format llm`, and in the ingest summary, so an
 automated consumer can tell it apart from a clean run. It is not an error: it does not set a non-zero exit code and does
 not count towards `stitchErrors`, and the previous registration stands — the
 same position a stitch that *failed* already left the graph in. `ix map` prints
@@ -271,7 +270,7 @@ data to send, having only parsed what changed. A run that re-ingests every file
 | Variable | Default | Effect |
 |---|---|---|
 | `IX_STITCH_COOLDOWN_MS` | `900000` | How long to hold off after a cut-off stitch. `0` disables the cooldown. |
-| `IX_STITCH_SLOW_FAILURE_MS` | `20000` | Failures at or past this wall-clock are treated as "the backend may still be working". `0` turns the elapsed rule **off** (only an abort cools down) — it does not mean "everything is slow". |
+| `IX_STITCH_SLOW_FAILURE_MS` | `20000` | Failures at or past this wall-clock are treated as "the backend may still be working". `0` turns the rule **off** — it does not mean "everything is slow". |
 | `IX_STITCH_WAIT_MS` | `30000` | How long to wait for an in-flight stitch before skipping. `0` sheds immediately. |
 | `IX_LOCK_DIR` | `~/.ix/locks` | Where the stitch lock and cooldown record live (shared with the map lock). |
 
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -289,6 +289,15 @@ apart from a clean run. It is not an error: it does not set a non-zero exit code
 and does not count towards `stitchErrors`, and the previous registration stands —
 the same position a stitch that *failed* already left the graph in.
 
+The wait happens **inside** `ix map`'s per-workspace lock, which the run holds
+until it exits. So while one map is waiting out another repo's stitch — up to
+`IX_STITCH_WAIT_MS`, 30s by default — any further `ix map` fired for that same
+workspace (an auto-map hook, for instance) finds the lock held, coalesces, and
+exits 0. Edits made in that window get no graph refresh and nothing says so.
+That is a new source of staleness, bounded by `IX_STITCH_WAIT_MS`; set it to `0`
+to shed on contention immediately instead, at the cost of losing that map's
+cross-repo registration.
+
 Both the lock and the cooldown are keyed on the **endpoint**, not on the
 workspace, because the join they bound is cross-workspace. That is the point of
 the guard, and it is also its cost: one transient failure while mapping repo A
@@ -321,7 +330,7 @@ count rather than every skip, and `skipReasons.emptyFile`, previously hardcoded 
 |---|---|---|
 | `IX_STITCH_COOLDOWN_MS` | `900000` | How long to hold off after a stitch that did not prove it stopped, measured from when the attempt ended. `0` disables the cooldown; single-flight stays. Values under ~2 min are not honoured after a killed process — see above. |
 | `IX_STITCH_WAIT_MS` | `30000` | How long to wait for an in-flight stitch before skipping. `0` sheds immediately. |
-| `IX_LOCK_DIR` | `~/.ix/locks` | Where the stitch lock and cooldown record live (shared with the map lock). |
+| `IX_LOCK_DIR` | `~/.ix/locks` | Where the stitch lock and cooldown record live (shared with the map lock). `ix reset` clears the cooldown, so the full re-ingest that follows one is not refused by it. |
 | `IX_MAP_LOCK_MAX_MS` | `1200000` | Shared with the map lock: how old a held lock must be before it is presumed abandoned and stolen. Lowering it to a few seconds so a wedged `ix map` self-heals faster also lets a second process steal the stitch lock from an in-flight stitch. The cooldown normally catches that on the next read, so it only matters together with `IX_STITCH_COOLDOWN_MS=0` — which is the one case where "single-flight stays" stops being true. |
 
 This bounds the client. Cancelling the server-side query when the client hangs
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -289,6 +289,16 @@ apart from a clean run. It is not an error: it does not set a non-zero exit code
 and does not count towards `stitchErrors`, and the previous registration stands —
 the same position a stitch that *failed* already left the graph in.
 
+Both the lock and the cooldown are keyed on the **endpoint**, not on the
+workspace, because the join they bound is cross-workspace. That is the point of
+the guard, and it is also its cost: one transient failure while mapping repo A
+refuses the stitch for repos B..E on the same backend for the whole cooldown,
+and when it expires none of them re-attempt on their own for the reason below.
+In a multi-repo setup that means every repo's cross-repo edges stay as they were
+until somebody runs a full re-ingest in each. The alternative -- keying per
+workspace -- does not bound the query at all, since the query is not per
+workspace.
+
 Note that re-registration is not automatic on the next map, and was not before
 this change: the stitch is gated on every file having been parsed this run, so an
 incremental map that skips an mtime- or hash-unchanged file neither reaches it nor
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -304,14 +304,18 @@ apart from a clean run. It is not an error: it does not set a non-zero exit code
 and does not count towards `stitchErrors`, and the previous registration stands —
 the same position a stitch that *failed* already left the graph in.
 
-The rule is machine-readable: `in-flight`, `cooling` and `deadline` are the
-guard's, and **`incomplete`** is `ix ingest`'s own completeness gate — an
-incremental map that did not re-parse every file has no complete registration to
-send. That last one is by far the commonest, and it is reported for the same
-reason as the others: a consumer asking "are the cross-repo edges current?" gets
-the wrong answer if the field is absent whenever the stitch was never attempted.
-It is deliberately the one rule that prints no human Note, because it would then
-appear on every incremental map.
+The rule is machine-readable, and carried alongside the prose as
+`stitchSkippedRule` / `stitch_skipped_rule`: `in-flight`, `cooling` and
+`deadline` are the guard's, and **`incomplete`** is `ix ingest`'s own
+completeness gate — an incremental map that did not re-parse every file has no
+complete registration to send. That last one is by far the commonest, and it is
+reported for the same reason as the others: a consumer asking "are the cross-repo
+edges current?" gets the wrong answer if the field is absent whenever the stitch
+was never attempted. Branch on the rule, never on the sentence.
+
+`incomplete` is deliberately the one rule that prints no human Note and emits no
+`--silent` token, both of which would otherwise appear on nearly every
+incremental map. `--format json` and `--format llm` carry it either way.
 
 The wait happens **inside** `ix map`'s per-workspace lock, which the run holds
 until it exits. So while one map is waiting out another repo's stitch — up to
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -290,10 +290,22 @@ and does not count towards `stitchErrors`, and the previous registration stands 
 the same position a stitch that *failed* already left the graph in.
 
 Note that re-registration is not automatic on the next map, and was not before
-this change: the stitch is gated on `filesSkipped === 0`, so an incremental map
-that skips any mtime-unchanged file neither reaches it nor has the registration
-data to send, having only parsed what changed. A run that re-ingests every file
-(`ix ingest <root> --force`, a post-reset re-map) is what picks it back up.
+this change: the stitch is gated on every file having been parsed this run, so an
+incremental map that skips an mtime- or hash-unchanged file neither reaches it nor
+has the registration data to send, having only parsed what changed. A run that
+re-ingests every file (`ix ingest <root> --force`, a post-reset re-map) is what
+picks it back up.
+
+The gate counts only files skipped **as unchanged**. It used to be the whole
+`filesSkipped` total, which also counts zero-byte files, ones that look minified
+and ones the parser returned nothing for — none of which have anything to
+contribute to the registration, so their absence does not make the collected set
+partial. Counting them meant a single empty `__init__.py` disqualified a repo from
+stitching permanently, `--force` included: the run printed nothing, exited 0, and
+left the cross-repo edges stale, while the message above advertised `--force` as
+the way to fix it. `skipReasons.unchanged` in `--format json` is now that narrower
+count rather than every skip, and `skipReasons.emptyFile`, previously hardcoded to
+`0`, is the real number.
 
 | Variable | Default | Effect |
 |---|---|---|
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -152,18 +152,17 @@ Ingest a path into the graph. Long-running: the client allows **30 minutes**.
   "filesSkipped": 3,
   "entitiesCreated": 640,
   "latestRev": 217,
-  "skipReasons": { "unchanged": 1, "emptyFile": 1, "parseError": 0, "unparsed": 1, "tooLarge": 0 }
+  "skipReasons": { "unchanged": 1, "emptyFile": 1, "parseError": 1, "tooLarge": 0 }
 }
 ```
 
-`unchanged`, `emptyFile`, `minifiedLikely` and `unparsed` are the buckets that
-are subsets of `filesSkipped`; `parseError` and `tooLarge` are counted
-separately and always have been. `unchanged` means specifically *skipped because
-we assumed it had not changed* — an mtime- or hash-clean file — and not every
-skip; see the stitch section for why that distinction is load-bearing.
-`unparsed` counts files the parse pool returned nothing for, which is not the
-same as `parseError` (that also covers stat, read and patch-build failures, most
-of which are not skips at all).
+> The **CLI's** `ix ingest --format json` emits a different, narrower breakdown
+> under the same key: `unchanged` there counts only files skipped as *mtime- or
+> hash-unchanged*, `emptyFile` is a real count rather than a hardcoded `0`, and
+> there is an extra `unparsed` bucket for files the parse pool returned nothing
+> for. That is the client's own summary of its own run and is not this response;
+> see the stitch section below for why the narrower `unchanged` is load-bearing
+> there.
 
 #### POST `/v1/map`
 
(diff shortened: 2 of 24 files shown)


KageBinary:
Reviewed at `ab4eb5d`. The core of this is right and the reasoning in `stitch-guard.ts` is the good kind — the elapsed-based classifier over a text matcher, and the 4xx-except-408 carve-out, are both the correct calls and are argued rather than asserted. Five findings, none of them in the guard's logic; four are in what four rounds of edits left behind, and they cluster on the same seam.

---

### 1. `stitchAbortReachedBackend` is dead code, and its five tests now pin a rule nothing runs

Round 4 removed the abort arm from `failureMayStillBeRunning` — `stitch-guard.ts` says so directly ("An abort used to get its own arm… so the arm added nothing"). The same commit removed the only call site:

```console
$ git show ab4eb5d -- ix-cli/src/cli/commands/ingest.ts | grep stitchAbortReachedBackend
-            // Sampled BEFORE the request -- see stitchAbortReachedBackend.
-                aborted: stitchAbortReachedBackend(err, deadlineAlreadySpent),
```

The function, its doc comment and its five tests survived:

```console
$ grep -rn "stitchAbortReachedBackend" ix-cli/src/
ix-cli/src/cli/commands/ingest.ts:622:export function stitchAbortReachedBackend(...)
ix-cli/src/cli/__tests__/ingest-retry.test.ts:  (5 assertions)
```

Worse than merely dead: those tests are the ones round 3 added to prove that a backend body saying `"AQL: transaction aborted"` is *not* read as a client abort. They still pass, and they no longer touch the code path that decides it — `failureMayStillBeRunning` does, on elapsed and status. Anyone changing abort classification later gets a green suite that proves nothing.

`knip` does not catch it, because the test file imports it. Worth knowing about that gate generally.

**Fix:** delete the function and the `describe` block, or restate those five cases against `failureMayStillBeRunning`. `deadlineAlreadySpent` goes with it; `isAbortError` is still used at line 635 and stays.

---

### 2. `ix ingest --format json` drops `stitchSkipped`, which is the thing `map.ts` explicitly refuses to do

`map.ts` states the rule and follows it:

```ts
// `?? null`, not left undefined: JSON.stringify drops an undefined
// value, so a consumer could not tell the field apart from an older
// CLI that never emitted it. Its siblings are always present too.
stitch_skipped: localIngest?.stitchSkipped ?? null,
```

`ingest.ts` emits the bare `string | undefined` into its own JSON body:

```ts
stitchErrors,
stitchSkipped,        // dropped entirely when undefined
elapsedSeconds: ...
```

```console
$ node -e 'console.log(JSON.stringify({stitchErrors:0, stitchSkipped:undefined, elapsedSeconds:1.2}))'
{"stitchErrors":0,"elapsedSeconds":1.2}
```

Same argument, same fix (`?? null`), and its siblings there — `commitErrors`, `stitchErrors` — are numbers that are always present, so it is the only key in that body whose absence is ambiguous.

---

### 3. `stitch_skipped` has no test coverage on either surface

It is the whole of round 3's finding 4 ("landed where the hooks do not look"), and nothing asserts it:

```console
$ grep -rn "stitch_skipped\|stitchSkipped" ix-cli/src/cli/__tests__/
(no matches)
```

The 35 tests in `stitch-guard.test.ts` are all on the guard module. `llm-renderers.test.ts` does cover `renderMapLlm`, but calls it as `renderMapLlm(result, regions)` — no `ingest` argument — so the new record is never emitted in a test. A refactor that drops the field from either JSON body or from the llm line is silent, which is exactly the failure mode the field exists to prevent.

---

### 4. `IX_MAP_LOCK_MAX_MS` now bounds the stitch lock, and nothing says so

`acquireLockAt` is shared, so `isStale` applies `lockMaxMs()` — `IX_MAP_LOCK_MAX_MS`, default 20 min — to the stitch lock too. At the default this is a harmless backstop. Lowered, it inverts rule 1: a knob whose only documented purpose is "outlast a legitimately slow map" would let a second `ix map` treat a live 60s stitch as stale, steal the lock and issue a second join — precisely when the backend is slow, which is when the guard is supposed to matter.

The new table in `docs/api/README.md` lists `IX_STITCH_COOLDOWN_MS`, `IX_STITCH_SLOW_FAILURE_MS`, `IX_STITCH_WAIT_MS` and `IX_LOCK_DIR` (correctly flagged as "shared with the map lock") but not this one. Either give the stitch lock its own floor, or add the row and say what lowering it costs.

---

### 5. Two `Pick`s widened for a member they never read

```ts
export function describeDroppedFiles(
  ingest: Pick<IngestFilesSummary, "parseErrors" | "commitErrors" | "stitchSkipped"> | undefined,
): string | undefined {
  const parse = ingest.parseErrors;
  const commit = ingest.commitErrors;   // stitchSkipped never referenced
```

Same in `emitDroppedFileWarning`. Only `renderMapLlm` actually uses it. The cost is that every caller and every test fixture now has to supply a field the function ignores — the widening is load-bearing on the tests, not on the code.

---

### Not findings, for the record

- **The 15-minute cooldown does not probe**, so a backend that recovers in 30s is still refused for 14.5 minutes, and any workspace whose only full map falls in that window loses its registration until someone runs `ix ingest <root> --force`. That is a genuine cost, but a health check is not evidence the *old join* finished, so there is no cheap probe that would be sound. Called out in `describeStitchSkipped` and the Note. Reads as the right trade to me.
- The settle-then-rethrow ordering, the second cooldown read after acquisition, and the `elapsedMs`-covers-upload reasoning behind the 4xx short-circuit all hold up.
- `stitchKey`'s loopback folding is justified on its cost rather than on a claim about deployments, which is the right way to justify a heuristic.

Verified locally at `ab4eb5d`: `typecheck` exit 0, full suite **90 files, 1541 passed, 2 skipped** (one earlier run had a single unrelated failure and took 242s against ~100s — the shape of #567).

KageBinary:
Second pass, still at `ab4eb5d`. Eight further findings that survived checking, and two I am flagging as **rejected** so nobody spends time on them.

### Rejected

- **`stitchKey` and default ports.** The concern is that `https://ix.example` and `https://ix.example:443` would hash to different locks. They do not — the WHATWG parser strips a scheme's default port, so `new URL("https://ix.example:443").port === ""` for both. `stitchKey` is fine as written.
- **`stitchAbortReachedBackend`'s doc comment does not mark itself unused.** I read all 20 lines of it; it still describes the function as live. My earlier finding stands, but nobody has annotated it.

### New

**6. `ix map --silent` is the one machine mode RULE 5 actually prescribes, and it carries nothing.** The `--silent` branch returns after a fixed `map: N files · …` line, before either JSON body. So `stitch_skipped` reaches `--format json` and `--format llm` only, while the comment above the llm record says "`--format llm` and `--silent` are what the hooks this field was added for actually read". The Note *does* reach stderr on that path (it is written before the `suppressOutput` return in `ingest.ts`), so it is not invisible — but a hook parsing the silent line gets no field, which is round 2 finding 3 for the third time.

**7. The `in-flight` remedy's premise is false for `ix ingest`.** It says contention "is always between DIFFERENT workspaces … the other run is registering its own, not this one", resting on `ix map` single-flighting per workspace. But `acquireMapLock` has exactly one call site:

```console
$ grep -rn "acquireMapLock" ix-cli/src/ | grep -v __tests__
ix-cli/src/cli/commands/map.ts:374:      const mapLock = acquireMapLock(cwd, ...)
```

`ix ingest` takes no workspace lock, and the stitch block is reached from `ingestFiles`. Two concurrent `ix ingest <root> --force` runs — or one alongside `ix map` on the same workspace — contend for the endpoint lock with the *same* `workspaceId`, and the loser is told to go do a redundant full re-ingest. Same class of wrong advice round 3 finding 2 removed.

**8. `docs/api/README.md` still documents the rule round 4 deleted, and contradicts itself six lines later.**

```
line 234: … or is aborted by the client's own timeout, no further stitch is sent …
line 243: An abort has no rule of its own …
```

A reader tuning `IX_STITCH_SLOW_FAILURE_MS=0` reads the table and believes aborts still cool down.

**9. `stitchSkippedRule` never reaches the wire.** Round 3 finding 7 replaced `reason.startsWith("another ix map")` with an explicit `rule`, because branching on prose is fragile. Only the prose is serialised — `stitch_skipped` is one English sentence in both JSON bodies — so any consumer that wants to retry on contention and back off on a cooldown has to do the exact sniff the code was forbidden from doing. Emitting the rule beside the reason costs one field.

**10. `readCooldown` validates the one field the expiry math does not use.** The guard is `typeof parsed?.until === "number"`, but `coolingUntil` computes `Math.min(cooldown.until, cooldown.at + stitchCooldownMs())`. A record missing `at` gives `Math.min(n, NaN) === NaN`, `NaN > now` is false, and the cooldown silently does not apply. Fail-open is the right default for this module, but here it is unintentional: validate `at` too, or clamp only when it is a number.

**11. The Note is false on a first map.** "cross-repository edges are unchanged since the last successful stitch" — on a workspace that has never been stitched there is no such point, and its cross-repo edges do not exist rather than being stale. The remedy line below it is right; the reassurance above it is not.

**12. `describeStitchSkipped(reason, rule?)` — the optional `rule` falls through to the cooling remedy.** So any caller that omits it silently reproduces the message round 3 finding 2 removed, printed under "another ix map is already stitching …". The rule is always available at the one call site; making the parameter required lets the compiler enforce what a hand-written test currently checks.

**13. Round 4 re-opened round 2 finding 5 through a documented knob.** Removing the abort arm means a pre-flight abort is classified on elapsed alone. At the 20s default that is safe. At `IX_STITCH_SLOW_FAILURE_MS=2000` — documented as tunable — a run deadline expiring during `JSON.stringify` of a megabyte-scale payload gives elapsed ≈ 2.5s, and the guard writes a 15-minute cooldown saying "cut off … and may still be running" for a request the backend never received. That is the case `stitchAbortReachedBackend` was written for, which is also why it is worth deciding its fate deliberately rather than just deleting it (finding 1).

Findings 6–13 came from a second reviewer pass; I verified each against the source before relaying, and dropped the two above that did not hold up.

Alot1z (commented):
Independent verification of the current head (`1387538`), plus one coordination finding.

**Verified against the code, not just the body**

- Read `stitch-guard.ts` end to end (561 lines): the lock-first/cooldown-while-holding order, the start-time marker with end-time re-stamp, the four proof arms, the 2xx-unparseable exclusion, and `positiveEnvMs`'s strict `^\d+$` match all match the documented design. The `parseOrThrowWithStatus` change in `api.ts` is what makes the 2xx-exclusion honest — a garbled 2xx now lands in the catch carrying its status.
- `settle()` is called on every exit path of the stitch attempt: `{ok:true}` after success, and in the catch before rethrow with `status`/`neverConnected` — the lock is never held past the request.
- Ran the PR's own suites on this head: `stitch-guard.test.ts` **51/51** and `stitch-scope-cache.test.ts` **10/10**, and the scope-cache file again **10/10** (the "passes twice" claim holds; nothing landed in the real `~/.ix`).
- Independent mutation: deleting the `neverConnected` proof arm turns **exactly one** of the 51 tests red — the claim is pinned, not decorative.

**One coordination finding — this PR and #591 fix the same bug in the same file**

`ix-cli/src/cli/__tests__/stitch-scope-cache.test.ts` is in this PR's diff (the "drive-by" USERPROFILE/HOME isolation, commit `d816a03`'s neighbourhood) **and** in the open Alot1z PR #591 (`feat/skill-install-toolscan`, commit `445c7fe`), which fixed the same defect independently — same root cause (`os.homedir()` honours USERPROFILE, not HOME), same mechanism (set `USERPROFILE = home` in `beforeEach`, restore in `afterEach`), same hunk position. Whichever merges second needs a one-line rebase there, and carrying the identical fix in two open PRs invites a divergence the next editor will have to reconcile.

Concrete resolution: keep it in whichever PR the maintainers would rather carry it; the other drops its copy on rebase. Happy to drop `445c7fe`'s hunk from #591 if this PR keeps it.

PR #572, opened 2026-09-02
fix(ingest): stop turning one failed bulk commit into one doomed request per patch

### LINKED ISSUE #560: [bug] ArangoDB transaction begin timeout during incremental ingest on healthy backend
## Summary

On Windows / Docker Desktop, Ix 0.10.5 with Memory Layer 1.0.28 fails to begin ArangoDB transactions during incremental ingest, although `ix doctor` reports a healthy, conflict-free graph and the host has no meaningful CPU/RAM/disk pressure.

This is not fixed by lowering batch size or concurrency.

## Environment

- OS: Windows 11, Docker Desktop / WSL2
- Ix CLI: 0.10.5
- Memory Layer: 1.0.28
- ArangoDB: 3.12.10-1 enterprise
- Workspace: mixed Python / TypeScript repository
- Discovered files: 160
- Changed files: 17–23
- Graph schema: v3

## Healthy state before ingest

```text
ix doctor

✓ Server reachable
✓ Graph has nodes and edges
✓ No unresolved conflicts
✓ Backend is the released image
✓ Graph schema matches engine
All checks passed.

### PR DESCRIPTION
**Pull request by josephismikhail**
Fixes #560.

When ArangoDB cannot begin a transaction, `ix map` turns one failed bulk commit into **one failed request per patch**, serialized behind the fallback mutex, each waiting out its own timeout — against the database that is already the reason they fail. The report says lowering batch size and concurrency does not help, and it cannot: the fan-out is per patch, so it ignores both knobs. Meanwhile `ix doctor` passes, because it asks whether the backend is reachable and the graph consistent, and a saturated ArangoDB is both.

---

## Reproducing it, on real infrastructure

`docker-compose.standalone.yml` — `ghcr.io/ix-infrastructure/ix-memory-layer:latest` against **ArangoDB 3.12.10-1 enterprise**, the versions in the report.

The precondition is the hard part: the backend must be *healthy* and unable to *write*. Holding a real **exclusive stream transaction** on the collections the memory layer writes (`nodes`, `edges`, `revisions`, `patches`, `idempotency_keys`, `claims`, `branches`) does exactly that — every read still works, `ix doctor` is green, and `beginStreamTransaction(exclusive)` inside the backend genuinely cannot start. ArangoDB expires an idle stream transaction at ~60s, so the holder keeps it warm with a trivial in-transaction read.

Under that lock, a 12-file ingest on `main`:

```console
  [bulk failed, falling back to per-file] files 1-12 (12 patches): TimeoutError: The operation was aborted due to timeout
  [commit error] src/m1.ts:  TimeoutError: The operation was aborted due to timeout
  [commit error] src/m2.ts:  TimeoutError: The operation was aborted due to timeout
  ...
```

Note what the failure actually *is*: a **client-side timeout with no HTTP status at all**. A saturated ArangoDB is not obliged to answer 500, and when it does not, the timeout is the only signal there is. That fact drives most of the design below, and it is why an earlier revision that ignored aborts was inert in exactly the shape it was written for.

---

## The fix

A run-scoped cutoff (`ix-cli/src/cli/commit-breaker.ts`) with three pieces, each a decision table that can be read and asserted on its own:

| | what it decides |
|---|---|
| `createCommitBreaker` | Has the backend refused `IX_COMMIT_FAILURE_LIMIT` (5) commits in a row without accepting one? A success un-trips it. |
| `perFileAction` | For the next patch: send it, hold it for the drain, or count it as already-landed. |
| `drainInPasses` / `createDrainGate` | How held patches get re-sent, and when to stop re-sending. |

The rule is deliberately about **behaviour, not error text**. Which message ArangoDB, the memory layer or an intervening proxy produces for "saturated" varies by deployment, and a classifier assembled from guessed strings is inert in exactly the deployments it was written for. What is unambiguous is a backend that has refused N commits in a row without accepting one: nothing about sending the next request differs from the last N.

### Nothing is abandoned that the backend would have taken

This is the half that took the most rounds, and it is worth stating as a property rather than a mechanism: **a patch is only ever reported as an error if it was sent, or if the backend has already proved it is taking nothing.**

Getting there needs two levels.

**Within a batch — `drainInPasses`.** Patches held by a trip are re-sent through the same per-file loop, bounded by a *budget* of failures rather than a streak, because the held set begins with whatever run of bad patches stopped the fan-out. A single bounded pass cannot avoid stranding good patches: it stops wherever the budget runs out and hands back everything past that point, and "handed back" means reported as a commit error, which suppresses the mtime baseline, which makes the next `ix map` repeat the same order and strand the same patches *forever*. Reversing the pass helps with exactly one cluster of bad patches and no more.

So each pass changes direction, and the third starts at the **midpoint**. What ends it is a pass that reached the end, or three passes — tail, head, middle — that between them placed nothing at all.

Every weaker rule strands committable patches, and each of them shipped here before the next was found. This is the part of the PR that took the most revisions, so the counterexamples are worth stating:

| rule | defeated by | what it stranded |
|---|---|---|
| stop when a pass places nothing | a *leading* cluster — the pass says nothing about what it never reached | `[g x 400, b x 15]` → 400 |
| stop after two consecutive empty passes | clusters at both ends of the *remaining* region | `[b x 5, g x 400, b x 10, h x 485]` → 400 |
| stop when the whole drain placed nothing, after two passes | two passes only ever sample the two **ends** | `[b x 5, g x 400, B x 10]` → 400 |
| a pass **cap** of any size | more failing patches than `passes x budget` | 1,000 held with 20 scattered bad → 294 |

There is deliberately **no pass cap**. Termination does not need one — a pass that continues has attempted at least `budget` items, so the unreached set strictly shrinks — and any cap hands back committable patches unsent, which is a real loss where this is only slowness.

The residue is stated rather than hidden: `budget` consecutive indicting failures at the head, the tail **and** the midpoint still gives up with work left. That needs fifteen bad patches in three specific places. A `mayGiveUp` flag briefly removed even that, keyed on whether the batch had accepted anything — it is gone, because that evidence is from the wrong moment (the successes come *before* the degradation, since the breaker trips after them), so one lucky patch disabled the stopping rule and let the drain issue ~494 serialized commits against a dead backend. Worse than the residue it closed.

**Across batches — `createDrainGate`.** A run has many batches and each can hold patches, so a genuinely dead backend would otherwise be re-probed once per batch for the length of the run — which is the #560 amplification, moved rather than removed. After two drains in a row that place nothing, the gate closes. It **reopens the moment a patch lands anywhere**: closing is an inference drawn from two drains, and the run keeps going, so anything landing afterwards falsifies it.

### What the user is told

`ix doctor` passing is the whole reason #560 was filed as "Ix is broken" rather than "the database is saturated", so the run prints the diagnosis once instead of one error per file behind `--debug`:

```text
Error: Commits against http://127.0.0.1:8092 kept failing, so ix stopped fanning out one request per patch.
  3 patches were not sent — sending them one at a
  time would have added load to a backend that is already the reason they fail.
  Last error: TimeoutError: The operation was aborted due to timeout
  The graph is unchanged. Re-run `ix map` once it is fixed.
  Either the database cannot keep up — `ix doctor` still passes, since an ArangoDB that
  cannot begin a transaction is reachable and consistent — or the backend is rejecting these
  particular patches. The error above is the backend's own answer and says which.
  Set IX_COMMIT_FAILURE_LIMIT=0 to send every patch regardless.
```

It prints whenever the cutoff **fired**, not when the run happened to lose patches. Gating it on the leftovers meant it vanished exactly when the drain worked: the 12-file reproduction below at the *default* limit trips on 5 failures, the drain sends the remaining 7, and the user got `Ingest committed nothing: all 12 patches failed to commit. Re-run with --debug to see why.` and no explanation of why `ix doctor` passes. The "N were not sent" line is dropped when N is 0 rather than printing a zero, and the headline says "kept failing" rather than "stopped committing" because the run may well have gone on to commit everything.

Naming **both** causes is not hedging. Verified against the released image: the memory layer answers `500 {"error":"internal_error","message":"Invalid message body: …"}` for a patch body it will not accept, and the saturated path produces a `TimeoutError` with no status at all. The CLI has nothing to tell them apart with, so asserting the database is the cause would send a user with a rejected patch to `docker stats` and `/_api/query/current` — the wrong subsystem entirely.

---

## Results

**Real backend, 12 files, exclusive Arango lock held throughout, `IX_COMMIT_FAILURE_LIMIT=3`:** 9 commit requests sent, 3 held and never sent, and the banner above instead of twelve identical `[commit error]` lines.

**Stub backend, 30 files** (a memory layer that fails or rejects on demand, logging every commit request):

| profile | `main` | this PR |
|---|---|---|
| every commit fails | 31 requests | **22 requests** |
| healthy | 1 bulk, 0 per-file | 1 bulk, 0 per-file |
| backend recovers mid-batch | 31 requests | **7 requests** |
| poison cluster of 12, **middle** | 31 requests, 12 errors | **32 requests, 12 errors** |
| poison clusters 5 + 12 | 31 requests, 17 errors | **32 requests, 17 errors** |
| poison cluster of 12, **leading** | 31 requests, 12 errors | **32 requests, 12 errors** |
| backend saturates part-way through a batch | 26 requests, 20 errors | **27 requests, 20 errors** |
| bulk 409s naming every patch as landed | 31 requests | **6 requests** |
| 1,200 files, 25 rejects (5 adjacent + 20 scattered) | 1003 requests, 25 errors | **1003 requests, 25 errors** |

The rows where the error count equals `main`'s are the property: on a backend that refuses particular patches, this costs nothing and loses nothing, in every cluster placement. The `+1` on those rows is the drain's bulk probe.

The saturated row is a third off. The recovery row is the probe's whole point — one bulk instead of thirty serialized commits — and the 409 row is the replay bound: without it a single bulk answering "409, all landed" sent 30 doomed serialized commits with nothing able to stop them.

For contrast, measured rather than argued: the two-cluster profile against the **single-pass** drain of round 10 gives 24 requests and 22 commit errors — five committable patches reported as failures without ever being sent.

---

## Review history

Twenty-six rounds. Several rounds fixed a defect and introduced another, and the recurring cause was diagnosed rather than patched over: the decision lived in a closure inside `ingestFiles` that no test could reach, so every revision of it was verified by reading rather than by running. Extracting `perFileAction`, then `drainInPasses` and `createDrainGate`, is the response to that pattern — each extraction immediately caught a bug the PR had already shipped.

<details>
<summary>Rounds 1–5: the shape of the cutoff</summary>

**Round 1** (6): the cutoff skipped reconcile; it skipped a chunk's BULK commit as well as the fan-out, so a recovered backend could not prove it; `IX_COMMIT_FAILURE_LIMIT` parsing; the summary did not distinguish commit errors from parse errors; the banner quoted nothing.

**Round 2** (4): abandoned patches were counted as applied; the mtime baseline was written for files whose patches never landed, so the next run skipped them and they stayed missing until a `--force`.

**Round 3** (5): the lock-conflict exclusion made the cutoff inert in a real saturation mode — `timeout waiting to lock key` and `error: 1200` are exactly what a RocksDB-backed ArangoDB emits under sustained write contention, *and* those same patterns are in the retry list, so every doomed patch was being multiplied by `retryOnConflict`. Reverted.

**Round 4** (3): the cutoff abandoned patches rather than holding them; it consulted the breaker before a bulk, turning one request into N.

**Round 5** (5): the drain re-sent the held set front-first, spending its budget on the same bad patches that stopped the fan-out; the failure budget was inert because the streak always preempted it; a success did not un-trip the latch.
</details>

<details>
<summary>Rounds 6–10: the extraction, and three bugs it caught</summary>

**Rounds 6–8** (12 across three rounds): the replay guard's condition was inverted, so nothing was held back at all — caught by measurement, not by reading: the saturated profile went from 14 requests to 35. The budget never bound because the streak preempted it. An entry check turned a bulk into N serialized commits.

Those are all decision-table bugs in one closure, so **round 9 extracted the table** as `perFileAction`, with seven tests. Writing them found three behaviours the PR had shipped and nobody had noticed.

**Round 10** (5): 413s could exhaust the drain budget, so a handful of oversized generated files stranded hundreds of committable patches on a healthy backend; `cutoffRetrySpent` latched on one unlucky drain; the "landed before it stopped" count was a run total on the bulk path; the banner claimed a consecutive streak the run may never have seen; a troubleshooting row keyed on text the CLI no longer emits.
</details>

<details>
<summary>Round 11: five fixed, one answered</summary>

1. **The drain stranded committable patches** whenever the failures ran longer than its budget — see *Nothing is abandoned* above. Fixed by `drainInPasses`; measured end-to-end at 22 commit errors before, 17 after, against 17 poison patches.
2. **`cutoffRetrySpent` was a one-way latch**, and the line resetting its counter lived *inside* the branch the latch disables, so nothing could ever reopen it. A thirty-second blip in batches 3 and 4 disabled the drain for the rest of a forty-batch run. Fixed by `createDrainGate`, which reopens when a patch lands.
3. **`appliedAtCutoff` latched on the first hold** and was never cleared, so a run that tripped in batch 1, recovered and committed 19,000 more could print "The graph is unchanged." above a summary saying 19,200 files processed. Deleted rather than fixed: it existed to support "N landed *before it stopped*", and since a cutoff holds patches and the run continues, there is no such moment.
4. **`perFileAction` returned "hold" for a replay with a spent budget**, but the only consumer turns every hold into a commit error — and a replay's patches are confirmed landed by the server's own 409. Now returns `count-applied`. Latent (no caller passes both), but a decision table must not contradict its consumer.
5. **Two load-bearing comments described mechanisms that do not run**: "the breaker is ACTIVE, so this stays bounded" (with a budget in play `perFileAction` ignores the streak entirely) and "if NOTHING landed, they are left for the next run" (a check the old code did not make).
6. **Not changed**: `commitFailureIndictsBackend`'s default of `true`. The finding asks that per-patch rejections stop feeding the run-wide cutoff, and there is no way to detect them here — see the verified 500-for-both above. What the finding is right about is the *message*, and that is fixed.

Every new guarantee is mutation-validated: removing the direction flip, ignoring a pass that placed nothing, closing the gate on one miss, never reopening it, and holding a spent-budget replay each turn exactly one test red.
</details>

<details>
<summary>Round 12: the same mistake in a new place, twice</summary>

1. **`drainInPasses` still stranded committable patches** — it returned on the first pass that placed nothing, which is not evidence when the pass stopped early. See *Nothing is abandoned* above; measured at 410 stranded before, 0 after. The round-11 test missed it because it put the cluster in the *middle*, where the first pass places the trailing block before meeting it: the assertion was true and proved less than it claimed.
2. **The banner was gated on `skipped() > 0`**, so the diagnostic #560 was filed for vanished whenever the drain placed everything it held — including on this PR's own 12-file reproduction at the default limit. Now gated on `everTripped()`, with `describeCommitOutcome` taking the same flag so it no longer says "re-run with `--debug` to see why" directly beneath a message that just said why.
3. **`count-applied` raises `patchesApplied` without sending a request** (those patches are confirmed landed by a 409 body), and `createDrainGate` read that counter — so a full-landed replay could reopen a gate that had closed against a dead backend. The gate now reads a counter incremented only where a request came back OK.
4. **Once the run deadline fired, the cutoff went inert.** `commitFailureIndictsBackend` returns false for every error past the deadline, so the drain budget never decremented and the breaker never tripped: nothing could stop the per-file fan-out for the tail of the run. That classification is right — the deadline is our clock, not the backend's answer — so the loop now holds the remainder the moment the deadline has fired, and the drain does not run past it.
5. **A comment claimed the mtime baseline is not marked for `count-applied` patches.** It is: `persistIngestBaselineIfClean` is gated on the run-wide commit-error count and that branch adds none. Corrected, with the reason it is *correct* here stated rather than left to be re-derived.

Findings 1 and 3 are both "a number that looks like evidence and is not", which is the same shape as rounds 5 and 8. That is the argument for the extractions: each of the three is now a decision table with its own tests, and all three of round 12's mechanical findings were in code that had already been reviewed twice.
</details>

<details>
<summary>Rounds 13–26: the replay path, the drain's bound, and five rules that were wrong</summary>

The later rounds are dominated by two things: the full-landed replay path, and successive attempts to state when the drain may stop. Both are covered above; what follows is the record.

**Round 13** (5). The **full-landed replay was unbounded**: when a bulk fails with 409 naming every patch id as already committed, `commitBulkWithPayloadSplit` hands the whole chunk to the per-file loop as a replay and returns *before* its `shouldStop` check — and a replay failure took the branch that increments `patchesApplied` and never called `recordFailure`. Measured against a stub answering 409-with-every-id: `main` 31 requests, round 12 **31**, round 13 **6**. Also: deadline-held patches were attributed to the cutoff; a message steered readers away from the only per-file detail available; an orphaned JSDoc; a comment describing the rule round 12 removed.

**Round 14** (3), two of them consequences of round 13 — a replay caught by the clock had confirmed-landed patches counted as errors, and a run could trip, place everything, exit 0, and still print `Error: Commits against <endpoint> kept failing`.

**Rounds 15–19** (16). The drain's bulk probe (recovery 31 → 7 requests) and its guards: deletions must not go to the bulk endpoint, the probe must be chunked, a partial 409 must be honoured. The banner's quoted error going stale. And one commit whose message described three fixes while its diff carried one — my patch script aborts on a failed anchor and discards the earlier edits in the same run, and I wrote the message from intent rather than from the diff.

**Rounds 20–26** (21). The drain's stopping rule, four times (the table above), plus: the gate treating "nothing stranded" as proof the backend is alive; the gate unable to reopen because a closed gate blocks the very progress it waits for; the replay bound indicting the whole run for an idempotency error the CLI knows says nothing about the backend; and a steady tail of comments describing mechanisms that no longer ran.

</details>

<details>
<summary>Rounds 13–14 in detail</summary>

**Round 13** (5). The **full-landed replay was unbounded**: when a bulk fails with 409 naming every patch id as already committed, `commitBulkWithPayloadSplit` hands the whole chunk to the per-file loop as a replay and returns *before* its `shouldStop` check — and a replay failure took the branch that increments `patchesApplied` and never called `recordFailure`. The streak could neither be consulted nor advanced, so up to `IX_COMMIT_HTTP_MAX_FILES` doomed commits went out one at a time behind the global fallback mutex, blocking every other chunk's fallback, unstoppable. Measured against a stub that answers 409-with-every-id and then fails each single commit, 30 patches: `main` 31 requests, round 12 **31**, round 13 **6**.

Also: patches held by the *run deadline* went into the same list as cutoff-held ones, so the banner blamed the backend for what the clock abandoned — reproduced at a 4s `IX_MAP_DEADLINE_MS`, printing "23 patches were not sent ... a backend that is already the reason they fail" directly above "Ingest ran out of time ... raise IX_MAP_DEADLINE_MS". Plus a message that steered readers away from the only per-file detail available, an orphaned JSDoc, and a comment describing the rule round 12 removed.

**Round 14** (3), two of them consequences of round 13. Putting the deadline check above the replay check meant a replay caught by the clock had confirmed-landed patches counted as commit errors. Making replay failures feed the breaker meant a run could trip, place everything, persist the baseline, be classified `{ kind: "ok" }` and exit 0 while stderr printed `Error: Commits against <endpoint> kept failing` — measured: banner printed before, absent after. And past the deadline the drain is skipped, so anything still held was the clock's, yet still went to `recordSkipped`; that third one is a one-line gate I could **not** reproduce against the stub (every timing had the drain finish first) and the commit says so.
</details>

---

## Behaviour and knobs

* Default limit **5**; `IX_COMMIT_FAILURE_LIMIT=0` restores the pre-#560 behaviour exactly (send every patch regardless).
* A chunk's **bulk** commit is never skipped. One request per chunk is not the amplification — the fan-out at one request per patch is — and sending it is what lets a backend that recovered mid-run prove so and unstick the rest.
* Only failures that say something about the backend feed the cutoff: the run's own deadline is excluded (that is our clock, not the backend's answer), and so is payload-too-large (about one patch, not the server). A per-request timeout is **not** excluded — a backend that hangs produces nothing else.
* A lock conflict is not excluded either, and by the time one reaches this code `retryOnConflict` has already spent six backed-off attempts on it.
* The mtime baseline is not written on a run with commit errors, so the next `ix map` re-ingests whatever did not land.

## Verification

| check | result |
|---|---|
| `commit-breaker.test.ts` | 53/53 |
| `npx vitest run` (full) | **1561 passed, 3 failed, 21 skipped** |


KageBinary:
Reviewed at `ccab3c4`. This one is in better shape than its sibling — I went looking for the same residue I found on #571 and mostly did not find it. Round 4 removed `isCommitLockConflict` cleanly (`COMMIT_LOCK_CONFLICT_PATTERNS` is still live in the retry list, nothing orphaned), and the integration behaviour is genuinely tested: `ingest-retry.test.ts` asserts the abandonment, the check between payload-split halves, and the replay flag, not just the pure helpers. Three findings, one of which I think matters.

I also checked the load-bearing claim in the module header — *"the mtime baseline is not written on a run with commit errors, so the next `ix map` retries every file it did not commit"* — because that is what makes a false trip cheap. It holds. `persistIngestBaselineIfClean` gates on `ingestCompletedCleanly`, and the one ungated `saveIngestBaseline` (the deletion-recovery callback) writes `previousMtimes` at the *previous* rev, so it cannot mark this run's files as ingested. Worth saying out loud since the whole risk argument rests on it.

---

### 1. The cutoff is invisible to every machine consumer

This is the one I would not merge without a decision on.

The PR's own framing is that the diagnostic *is* the deliverable — "This is the diagnostic #560 did not have", against a reporter whose symptom was an ingest failing while `ix doctor` passed. But nothing about the cutoff reaches a machine:

```console
$ grep -n "cutoff\|breaker" ix-cli/src/cli/commands/ingest.ts | grep -iE "IngestFilesSummary|json"
(no matches)
```

`IngestFilesSummary` is unchanged, so there is no field in `ix ingest --format json`, none in `ix map --format json`, none in `--format llm`. What a consumer sees is `commit_errors: N` — and it cannot tell **"N patches were attempted and rejected"** from **"5 were attempted, and N−5 were abandoned unsent because we gave up on the backend"**. Those have opposite remedies: retry now versus wait for the database. That distinction is the entire point of the change, and it exists only as English on stderr.

Two things sharpen it:

- Under `ix map --format json`, a cutoff with `patchesApplied === 0` is `kind: "fatal"`, which throws, and map.ts's `emitError` has an `llm` branch and an *else* — so the JSON path prints prose to stderr and **nothing to stdout**. The llm path does get `error code=backend_error message="Ingest stopped early: …"`, which is fine; json gets no structured record at all.
- #571, by the same author and landing alongside this, adds `stitchSkipped` to the summary *and* both JSON bodies, on the argument that hooks running `ix map` read machine output. That condition does not even move the exit code. This one does, and reports less.

`commitBreaker.skipped()` is already computed at the call site. Adding it to the summary and to both bodies is a few lines and makes the two PRs agree with each other.

### 2. `consecutiveFailures()` is only ever called by tests

```console
$ grep -rn "consecutiveFailures" ix-cli/src/
  commit-breaker.ts:74   (interface)
  commit-breaker.ts:95   (implementation)
  __tests__/commit-breaker.test.ts:37,57,78
```

Round 3 gave `describeCommitCutoff` the comment explaining why it reports `breaker.limit` and deliberately *not* the live streak ("Requests already in flight … keep incrementing, so the streak is whatever the race happened to produce"). That is the moment this accessor lost its production purpose. It is a fair test seam for the streak/latch interaction, so this is "decide and say so" rather than "delete": either drop it and assert the latch through `tripped()`, or note in the interface that it exists for tests. As on #571, `knip` will not tell you — a test-file import counts as usage.

### 3. The reconcile guard exists in two shapes

Round 4 correctly extended the skip to the deletion phase, and its comment says why: *"Guarding only the two patch-build loops and not this one is the omission that makes a guard read as complete when it is not."* But the two loops guard inline —

```ts
if (previousHash && reconcileWorthDoing()) {
  patch = await reconcileRemovedEntities(…)
```

— while the deletion site goes through a wrapper, `reconcileRemovedEntitiesIfWorthIt`. Both are correct. The cost is that a reader has to find both forms to know the guard is complete, which is the same "reads as complete when it is not" problem one level up. Routing all three through the wrapper would make it one thing to check.

---

### Checked and not findings

- The replay arm (`patchesApplied += items.length - itemIndex`) does not corrupt the baseline. `latched` is only ever set in `recordFailure`, which is called immediately after `commitErrors++`, so `tripped()` implies `commitErrors >= limit >= 1` and `persistIngestBaselineIfClean` refuses. `latestRev` under-reports on that path, which is harmless for the same reason.
- Concurrency: the breaker's counters are mutated from up to `IX_COMMIT_CONCURRENCY` chunks, but a bulk failure does not feed the streak, so the bound is roughly (concurrency) bulk requests + `limit` per-file attempts before the top-of-function `shouldStop` closes the rest. That matches the measurements in the description.
- `IX_COMMIT_FAILURE_LIMIT=0` really does mean "never trip" (`limit > 0 &&` in `recordFailure`), and the message that advertises it is only printed when `skipped() > 0`, which requires a trip. Consistent.
- `IX_COMMIT_FAILURE_LIMIT` is documented in `skills/ix/references/troubleshooting.md`, along with the "doctor can still pass" note — which is the part of #560 that actually confused the reporter.

Alot1z (commented):
## Independent E2E verification — saturated-backend behavior (measured)

I stood up `docker-compose.standalone.yml` (memory-layer **1.0.28** + Arango **3.12.10-1** enterprise — the exact versions from #560), held an exclusive stream transaction on the 7 write collections via a heartbeat holder, and ran this PR's head against the locked backend. Baseline happy-path ingest on the same backend first (32 files, rev 8, clean).

### The #560 failure class, reproduced verbatim
With the lock held, the bulk commit path fails with a **client-side `TimeoutError: The operation was aborted due to timeout` and no HTTP status** — exactly what #560 reported (write-blocked Arango, healthy otherwise), not the 512MB-ceiling 500.

### What this PR's breaker does under it (head `8280361`)
- Bulk attempt times out → falls back to per-file patches → refusals accrue → **breaker trips after 5 consecutive refusals**.
- Cutoff holds the remaining patches (27 held) and runs **bounded drain passes** — the drain gives up only after three passes (tail/head/middle) place nothing.
- Terminal state is honest and non-zero:
  `Error: Ingest stopped early: 32 of 32 file patches are missing from the graph, 12 of them never sent. See the cutoff above for why.`
- **Measured: 21 commit attempts total, 22 timeouts, then stop** — vs. main's unbounded fan-out that keeps attempting every patch against a backend that cannot answer.

### Mutation pin (independent of the PR's own tests)
Removing the replay-first stop decision → **2 tests red**; removing the `neverConnected`-style proof arm on the companion PR → **1 test red**. Both decision tables are pinned, not decorative.

### Disclosure
The scratch client's per-request commit timeout was shortened **5 min → 15 s** (breaker logic untouched, same real saturated backend, same `TimeoutError` class) so the full sequence completes in minutes instead of ~40 min wall; the two code paths are equivalent for refusal/streak purposes since every attempt times out identically. Full-speed behavior follows from the same decision tables (verified by reading the breaker at the 5-min timeout).

Verdict: the breaker behaves exactly as designed on real saturated infrastructure — bounded attempts, honest reporting, no silent drop. No change requested.

PR #577, opened 2026-09-02
docs: document ix savings, its flags, and its llm records

### LINKED ISSUE #573: docs: ix savings has no command documentation — usage, flags, and llm output are undocumented
**Finding.** `ix savings` is a registered top-level command (`ix-cli/src/cli/commands/savings.ts`,
with llm rendering and a `reset` subcommand) but is absent from every doc surface: no
`skills/ix` entry, no reference row, no README mention, no usage line. The only traces in
the repo are `docs/llm-format.md`'s Tier-5 command list and the *backend* API doc
(`docs/api/README.md` → `GET/DELETE /v1/savings`), which a CLI user will not connect to the
command.

Because `docs/llm-format.md` promises `--format llm` works unconditionally on the commands
it lists, the omission is load-bearing for agents: `savings` is listed as llm-capable but
nothing documents what that output contains or how the command behaves.

**Ask (documentation only).** Add a `skills/ix`/references entry for `ix savings`: what it
reports, its flags (`--format`, the `reset` subcommand, any others), and an example of the
`--format llm` output. Happy to make this a one-line-PR-sized doc change if preferred.

**Method note.** Found by a log-only scan (`docs-scan`) that inventories registered commands
and diffs them against the repo's markdown; the scan changes nothing. This is the only
top-level command it found in that state.

KageBinary:
Closes #573.

Confirmed the finding before writing anything. `savings` on `main` (`a2cab21`):

| surface | mentions |
|---|---|
| `docs/api/README.md` | 6 — but those are the **backend** endpoints, `GET`/`DELETE /v1/savings` |
| `docs/llm-format.md` | 1 — its name in the Tier-5 list, nothing about the output |
| `skills/ix/SKILL.md`, `references/*.md`, `README.md`, `CLAUDE.md`, `CONTRIBUTING.md` | **0** |

Meanwhile `registerSavingsCommand` ships `--detail`, `--model <model>`, `--format <fmt>`, a `reset` subcommand, and `renderSavingsLlm`. The llm-format promise is what makes this load-bearing rather than cosmetic: the doc tells agents `--format llm` works on every command it lists, `savings` is on that list, and nothing said what came back.

## Change

**`skills/ix/references/commands.md`** — a `## Session Metrics` section: routing rows in the same shape as the other tables, what the `session` and `lifetime` scopes actually mean, a flag table with values and defaults, `ix savings reset`, and the llm record shape.

**`docs/llm-format.md`** — an `ix savings --detail` example beside the other per-command examples, placed before the `Error line:` block so that lead-in still owns the snippet under it. Plus the two things the shape does not show on its own:

- `command` records appear **only** under `--detail`, and carry their own `scope=` because both scopes are broken down into one flat stream.
- `--model` moves `money_saved` and nothing else — token and water figures are model-independent. Verified: `--model sonnet` on identical data gives `money_saved=5.50` against opus's `27.48`, with `tokens_saved` and `water_saved_ml` byte-identical.

## Sourcing

Flags, defaults and choices come from the registered commander tree, not from prose; the HTTP calls from `IxClient.savings` / `savingsReset` (`GET /v1/savings?detail=true`, `DELETE /v1/savings`), which matches `docs/api/README.md`. Record shapes are real `ix savings --format llm` output against a live backend, with the sample figures swapped for modest ones rather than publishing a real account's lifetime totals.

Documentation only — no source, no tests, no behaviour touched.

### DIFF
--- a/docs/llm-format.md
+++ b/docs/llm-format.md
@@ -171,6 +171,20 @@ of two field sets it is holding. `saved_at` is on it because the `context`
 record that follows says whether the snapshot was fresh when it was taken, not
 when that was.
 
+`ix savings --detail`:
+
+```
+savings model="Claude Opus ($15/MTok in, $75/MTok out)"
+scope name=session commands=726 tokens_saved=832739 naive_tokens=1099164 actual_tokens=266425 money_saved=27.48 water_saved_ml=1665.478
+scope name=lifetime commands=4820 tokens_saved=6142880 naive_tokens=8003104 actual_tokens=1860224 money_saved=202.71 water_saved_ml=12285.76
+command scope=session name=callers count=329 tokens_saved=334995
+```
+
+The two `scope` records are always emitted; `command` records appear only under
+`--detail` and carry their own `scope=` because both scopes are broken down in
+one stream. `money_saved` is the only field `--model` moves — the token and
+water figures are model-independent.
+
 Error line:
 
 ```
--- a/docs/llm-format.md
+++ b/docs/llm-format.md
@@ -176,14 +176,22 @@ when that was.
 ```
 savings model="Claude Opus ($15/MTok in, $75/MTok out)"
 scope name=session commands=726 tokens_saved=832739 naive_tokens=1099164 actual_tokens=266425 money_saved=27.48 water_saved_ml=1665.478
-scope name=lifetime commands=4820 tokens_saved=6142880 naive_tokens=8003104 actual_tokens=1860224 money_saved=202.71 water_saved_ml=12285.76
 command scope=session name=callers count=329 tokens_saved=334995
+scope name=lifetime commands=4820 tokens_saved=6142880 naive_tokens=8003104 actual_tokens=1860224 money_saved=202.72 water_saved_ml=12285.76
+command scope=lifetime name=callers count=2104 tokens_saved=2210488
 ```
+
+(Truncated: one `command` record is shown per scope, and a real run emits one
+per entry in that scope's breakdown.)
 
 The two `scope` records are always emitted; `command` records appear only under
 `--detail` and carry their own `scope=` because both scopes are broken down in
-one stream. `money_saved` is the only field `--model` moves — the token and
-water figures are model-independent.
+one stream. They are emitted inside the scope loop, so they follow their own
+`scope` record rather than being grouped at the end — a parser reading the
+stream in order sees session's breakdown before the lifetime `scope` line.
+
+`money_saved` is the only field `--model` moves — the token and water figures
+are model-independent.
 
 Error line:
 
--- a/skills/ix/references/commands.md
+++ b/skills/ix/references/commands.md
@@ -65,6 +65,43 @@ These aggregate multiple graph operations into single bounded responses.
 | Start backend | `ix docker start` | `ix docker start` |
 | Graph statistics | `ix stats` | `ix stats --format json` |
 
+## Session Metrics
+
+| Goal | Command | Example |
+|---|---|---|
+| Token savings so far | `ix savings` | `ix savings --format llm` |
+| Per-command breakdown | `ix savings --detail` | `ix savings --detail --format llm` |
+| Clear lifetime totals | `ix savings reset` | `ix savings reset` |
+
+`ix savings` reports how many tokens the graph saved against a naive
+read-the-files baseline, in two scopes — `session` (this shell's commands) and
+`lifetime` (every command against this backend). Each scope carries the command
+count, `tokens_saved`, the `naive_tokens` / `actual_tokens` it was derived
+from, and money and water estimates.
+
+| Flag | Value | Default | Effect |
+|---|---|---|---|
+| `--detail` | — | off | Add one record per command type, within each scope |
+| `--model <model>` | `opus\|sonnet\|haiku\|gpt-4o` | `opus` | Pricing table for `money_saved` only — token and water figures do not change |
+| `--format <fmt>` | `text\|json\|llm` | `text` | See references/output-formats.md |
+
+`ix savings reset` clears the **lifetime** totals (`DELETE /v1/savings`). It
+takes no flags and does not prompt.
+
+`--format llm` emits one `savings` header record and one `scope` record per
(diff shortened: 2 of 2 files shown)


docs/llm-format.md
```diff
@@ -171,6 +171,20 @@ of two field sets it is holding. `saved_at` is on it because the `context`
 record that follows says whether the snapshot was fresh when it was taken, not
 when that was.
 
+`ix savings --detail`:
+
+```
+savings model="Claude Opus ($15/MTok in, $75/MTok out)"
+scope name=session commands=726 tokens_saved=832739 naive_tokens=1099164 actual_tokens=266425 money_saved=27.48 water_saved_ml=1665.478
+scope name=lifetime commands=4820 tokens_saved=6142880 naive_tokens=8003104 actual_tokens=1860224 money_saved=202.71 water_saved_ml=12285.76
```
josephismikhail: `money_saved=202.71` is not a value this command can print. `renderSavingsLlm` renders `estimateMoney(tokensSaved, model).toFixed(2)`, and for the `tokens_saved=6142880` on this same line that is `(6142880*0.7*15 + 6142880*0.3*75)/1e6 = 202.71504`, which `toFixed(2)` gives as `202.72`. The session row above checks out exactly (27.480387 -> `27.48`, water `1665.478`), and both `naive - actual` subtractions are consistent, so this one cell is the only thing in the block that cannot be reproduced -- which is what makes it costly: a reader treating the block as golden output (or a doc test asserting it) disagrees with the CLI by a cent, on the one field the prose below singles out as model-dependent. The same line is duplicated at `skills/ix/references/commands.md:97` and needs the same fix.

docs/llm-format.md
```diff
@@ -171,6 +171,20 @@ of two field sets it is holding. `saved_at` is on it because the `context`
...
+command scope=session name=callers count=329 tokens_saved=334995
```
josephismikhail, line 179: The `--detail` sample shows only `scope=session` command records, but `renderSavingsLlm` iterates `[["session", ...], ["lifetime", ...]]` and, when `detail` is set, emits the `byCommandType` rows for **both** scopes -- so real `ix savings --detail --format llm` output on this data (lifetime `commands=4820`) also contains `command scope=lifetime ...` rows, interleaved after the `scope name=lifetime` record rather than grouped at the end. A consumer writing a parser or a size estimate from this block will not expect them. One `command scope=lifetime ...` row in the sample would match the code and would demonstrate the very point the paragraph below makes -- that `scope=` is on the record because both scopes share one stream. `skills/ix/references/commands.md:98-99` has the same gap.

skills/ix/references/commands.md
```diff
@@ -65,6 +65,43 @@ These aggregate multiple graph operations into single bounded responses.
...
+## Session Metrics
+
+| Goal | Command | Example |
+|---|---|---|
+| Token savings so far | `ix savings` | `ix savings --format llm` |
+| Per-command breakdown | `ix savings --detail` | `ix savings --detail --format llm` |
+| Clear lifetime totals | `ix savings reset` | `ix savings reset` |
+
+`ix savings` reports how many tokens the graph saved against a naive
+read-the-files baseline, in two scopes — `session` (this shell's commands) and
```
josephismikhail: `session` (this shell's commands) is a mechanism claim the CLI cannot support. `IxClient.savings()` issues `GET /v1/savings` (plus `?detail=true`) with no session id, no header, and no pid -- `savingsReset` likewise -- so the backend has nothing to attribute a count to a shell with; whatever `session` means, it is a window the shared backend keeps, not per-terminal. Concrete failure: an agent runs `ix savings` in a second terminal, or in a fresh shell against a docker backend that has been up all day, and reports `commands=726 / tokens_saved=832739` as "what I saved in this session" when the figures include every other shell's traffic since the backend last started. Since this line is in a skill reference an agent acts on, prefer wording that does not claim shell scoping (the CLI's own labels are "This session" / "All time"), or state the actual reset boundary if it is known.

skills/ix/references/commands.md
```diff
@@ -65,6 +65,43 @@ These aggregate multiple graph operations into single bounded responses.
...
+`lifetime` (every command against this backend). Each scope carries the command
+count, `tokens_saved`, the `naive_tokens` / `actual_tokens` it was derived
+from, and money and water estimates.
+
+| Flag | Value | Default | Effect |
+|---|---|---|---|
+| `--detail` | — | off | Add one record per command type, within each scope |
+| `--model <model>` | `opus\|sonnet\|haiku\|gpt-4o` | `opus` | Pricing table for `money_saved` only — token and water figures do not change |
+| `--format <fmt>` | `text\|json\|llm` | `text` | See references/output-formats.md |
```
josephismikhail: Wrong relative path: this file already lives in `skills/ix/references/`, so `references/output-formats.md` resolves to `skills/ix/references/references/output-formats.md`, which does not exist. The `references/...` form used in `SKILL.md:62,77` is correct only because `SKILL.md` sits one level up. An agent told to load that path gets a missing file and either skips the format rules or wastes a turn searching. Sibling reference is `output-formats.md`.

```suggestion
| `--format <fmt>` | `text\|json\|llm` | `text` | See output-formats.md |
```

skills/ix/references/commands.md
```diff
@@ -65,6 +65,43 @@ These aggregate multiple graph operations into single bounded responses.
...
+
+`ix savings reset` clears the **lifetime** totals (`DELETE /v1/savings`). It
```
josephismikhail: The bolded **lifetime** asserts more than anything in the repo establishes. The only contract for this call is `docs/api/README.md:448` -- "DELETE `/v1/savings` -- Reset savings metrics" -- with no scope, and `savingsReset()` sends a bare DELETE with no body or query, so nothing on this side of the wire selects a scope; the CLI's `.description("Reset lifetime savings totals")` is the same unverified prose being re-published here as fact. Bolding it invites the reading that session counters survive: an agent that wants to zero only its all-time figures runs `ix savings reset` and finds the `session` scope zeroed too, with no undo (the command does not prompt -- that part is right). Either verify against the backend and say what it clears, or drop the emphasis and describe it as resetting the savings metrics.

Issue #576, opened 2026-09-02
RFC: add a docs-parity check to CI so a registered command or flag cannot merge without a doc mention


Alot1z:
**Problem.** Today nothing stops a command or flag from merging undocumented. Measured on
`main` (2026-09-02): `ix savings` is registered but has no doc mention anywhere; 51 of 77
implemented flags never appear on the doc surfaces. The last time a suite existed but no CI
job ran it, the project added a gate (#558); docs are in that same position now.

**Proposed check (minimal, opinionated).** A CI job (or a step in the existing test job)
that:
1. inventories top-level commands from the register layer and flags from `.option()` /
   `.requiredOption()` in `ix-cli/src/cli/commands/*`;
2. fails when a top-level command has no `ix <cmd>` mention (or backticked command-table
   form) on the canonical surfaces (`skills/ix/SKILL.md`, `skills/ix/references/*`,
   `README.md`, `docs/api/README.md`);
3. reports (does not fail on) flag-coverage, so it does not block mid-refactor work while
   still surfacing the ratio in the check log.

**Design questions for maintainers.**
- Should the check fail on new undocumented flags too, or only commands at first?
- Which surfaces are canonical? (The list above is a proposal.)
- Is a soft threshold (e.g. coverage %) better than a hard list for the transition period?

**Offered.** A working, log-only reference implementation exists on the fork side
(`repo-fork/scripts/docs-scan.mjs` — inventories commands/flags, diffs docs, changes
nothing, exits 0 clean or `--fail-on-drift`). Happy to adapt it to this repo's structure and
open a PR if the maintainers want the check; this issue is the discussion, not the PR.

**Scope guard.** This RFC asks for a *check*, not for rewriting the docs in one pass — the
check makes the drift visible and stops it growing; the doc pass is the sibling issue.

KageBinary:
Delivered, and wider than the RFC asked for — `ci-success` now requires **three** committed docs gates rather than one:

| job | script | what it gates | current |
|---|---|---|---|
| `doc-parity` | `check-doc-parity.mjs` (#584) | every registered command and flag has a doc entry, and no entry outlives its flag | 54 commands, 196 flags |
| `api-parity` | `check-api-parity.mjs` (#587) | four surfaces: paths+methods the client calls, `$ref` integrity, schema-name ↔ `types.ts` names, README endpoint coverage | 42 documented, 42 called, 13 schemas, 13 refs |
[...]

Issue #575, opened 2026-09-02
docs: flag-level documentation covers 26/77 implemented flags — request a per-command flags pass and a committed parity check


Alot1z:
**Finding.** A log-only scan of the CLI source against the repo's doc surfaces finds that
most implemented command flags are never spelled out in the docs. Coverage measured today:
**26/77 distinct flags (34%)** appear on the canonical surfaces. Representative examples
from the 51 never-documented set (full list in the evidence file):
`--as-of`, `--edge-cap`, `--exclude-kind`, `--god-module-chunks`, `--max-evidence`,
`--member-file-cap`, `--min-confidence`, `--no-open`, `--orphan-max-connections`,
`--regions`, `--remove-all-data`, `--semantic`, `--since`, `--tests-only`, `--upstream`,
`--weak-max-neighbors`.

This is not a single command's gap: it is spread across the command set, which suggests the
docs were written for the flags that existed when each surface was last edited rather than
kept in parity with the implementation.

**What this costs.** `docs/llm-format.md` tells agents to pass `--format llm`
unconditionally; a flag that exists but is undocumented is invisible to both humans and
agents until they read source. The project already treats machine-readable output as a
first-class surface; the flags that shape it deserve the same.

**Ask.**
1. A per-command documentation pass so every implemented flag appears in the docs
   (usage lines, references, or a flags table per command).
2. A lightweight, committed parity check so this stops regressing silently — see the
   companion RFC draft (`docs-parity-ci-proposal.md`) for a minimal CI shape; a working
   log-only reference implementation exists on the fork side and can be shared.

**Method note.** Findings come from a report-only scanner (inventories `.command()` /
`.option()` in source, diffs against the repo's markdown). It never patches anything.
Numbers carry a `scanned_at` stamp and full inventory in the evidence file above.

KageBinary:
Both halves of this request are on `main`.

**The per-command flags pass** — #578 landed `skills/ix/references/flags.md`, generated from the CLI's own registration rather than written by hand, via `ix-cli/scripts/dump-cli-surface.mjs`. It covers **54 commands and 196 flags**, against the 26/77 this issue measured.

**The committed parity check** — #584 landed `ix-cli/scripts/check-doc-parity.mjs` as a required `doc-parity` job in `ci-success`. It runs against the real reference and currently reports:

```
check-doc-parity: 54 commands, 196 flags — parity with skills/ix/references/flags.md
```

[...]

PR #578, opened 2026-09-02
docs: a complete per-command flag reference, generated from the CLI itself

KageBinary:
Addresses part 1 of #575 (the documentation pass). Part 2 — the committed CI parity check — is left to #576, deliberately; see the last section.

## The finding reproduces

Measured independently before writing anything, by walking the tree `registerOssCommands` actually builds rather than regexing source:

| | issue | measured here |
|---|---|---|
| implemented long flags | 77 | **76** |
| documented | 26 (34%) | **27 (35%)** |
| undocumented | 51 | **49** |

Close enough that the one- and two-flag differences are just where you draw the boundary (`--help`, Pro stubs). Every representative flag the issue listed is in my missing set. The conclusion stands as filed.

## Why this is generated, not written

The issue's own diagnosis is the important part: *"the docs were written for the flags that existed when each surface was last edited rather than kept in parity with the implementation."* Hand-writing 49 more rows reproduces that state on a later date. So:

**`ix-cli/scripts/dump-cli-surface.mjs`** — walks the registered commander tree and prints every command, argument, flag, default and choice list as JSON; `--flags` prints just the sorted long flags. It imports `dist/`, not `src/`, because that is what ships, and it fails with "run npm run build first" rather than a module-resolution stack trace.

**`skills/ix/references/flags.md`** — 39 top-level commands plus their subcommands, **196 flag rows**, values and defaults taken from the command objects. Coverage is now **76/76**, verified with the same script that measured the gap:

```
implemented long flags : 76
present in flags.md    : 76
still undocumented     : 0  []
commands not in flags.md: []
```

## Editorial decisions

- **Repetition kept.** `--format` appears on ~30 commands and gets a row on each, because the issue's ask is that every implemented flag appear under its command — a reader looking up `ix rank` should not have to know to check a shared table. Its Effect column points at `output-formats.md` instead of repeating one sentence 30 times.
- **Nothing paraphrased.** Values, defaults and choice lists are the command's own strings. Where a description already restates the choices, it is left as-is rather than reworded into something that could drift from the code.
- **A short preamble carries what a table cannot**: `ix diff` refusing to rank conflicting volume flags rather than picking a winner, `ix reset` being global (cross-referenced to the existing Gotchas section rather than duplicated), `--detailed` requiring `--list`, `--pick` being 1-based, and why an "off" default on a `--no-` flag means the positive behaviour is on. Each of those was checked against source, not assumed — the `ix diff` one against `diff.ts`'s explicit-conflict messages.

Cross-linked from `SKILL.md` (References list and step 4), `commands.md` (header), and `CLAUDE.md`'s repo map, so the file is reachable from the surfaces an agent already loads.

## On part 2

I have not wired a parity gate into CI. That is a policy question — what it blocks on, whether a new flag can merge undocumented, whether the check regenerates or only compares — and it is exactly what #576 exists to decide. This PR ships the inventory half that either answer needs; `dump-cli-surface.mjs --flags` is a one-line diff away from being that check once the shape is agreed.

Documentation and one new script. No source, no tests, no behaviour touched.

### DIFF
--- a/ix-cli/scripts/dump-cli-surface.mjs
+++ b/ix-cli/scripts/dump-cli-surface.mjs
@@ -1,0 +1,73 @@
+#!/usr/bin/env node
+/**
+ * Print the CLI's registered command/flag surface as JSON, one object per
+ * command (subcommands included, `parent sub` as the name).
+ *
+ * The point is that `skills/ix/references/flags.md` is checked against the
+ * *registered commander tree* rather than against a regex over source or, worse,
+ * someone's memory of what a command took. #575 measured 27 of 76 implemented
+ * flags documented; the gap was not carelessness, it was that nothing connected
+ * the two surfaces.
+ *
+ * Needs a build first (`npm run build`) — it imports `dist/`, not `src/`,
+ * because that is what actually ships.
+ *
+ *   node scripts/dump-cli-surface.mjs            # JSON to stdout
+ *   node scripts/dump-cli-surface.mjs --flags    # one long flag per line, sorted
+ *
+ * Deliberately NOT wired into CI: what a parity gate should block on is the
+ * open question in #576, and this is the inventory half either answer needs.
+ */
+import { Command } from "commander";
+import { createRequire } from "node:module";
+import { dirname, join } from "node:path";
+import { fileURLToPath, pathToFileURL } from "node:url";
+
+const here = dirname(fileURLToPath(import.meta.url));
+const ossPath = join(here, "..", "dist", "cli", "register", "oss.js");
+
+let registerOssCommands;
+try {
+  ({ registerOssCommands } = await import(pathToFileURL(ossPath).href));
+} catch (err) {
+  process.stderr.write(
+    `cannot load ${ossPath}\nRun 'npm run build' in ix-cli first.\n${err}\n`,
+  );
+  process.exit(1);
+}
+
+const program = new Command();
+program.name("ix");
+registerOssCommands(program);
+
+const commands = [];
+function walk(cmd, parentPath) {
+  const name = parentPath ? `${parentPath} ${cmd.name()}` : cmd.name();
+  commands.push({
+    command: name,
+    description: cmd.description(),
+    arguments: (cmd.registeredArguments ?? []).map(a => ({
+      name: a.name(),
+      required: a.required,
+    })),
+    options: (cmd.options ?? []).map(o => ({
+      flags: o.flags,
+      long: o.long,
+      short: o.short,
+      description: o.description,
+      default: o.defaultValue === undefined ? null : String(o.defaultValue),
+      choices: o.argChoices ?? null,
+    })),
+    subcommands: (cmd.commands ?? []).map(c => c.name()),
+  });
+  for (const sub of cmd.commands ?? []) walk(sub, name);
+}
+for (const cmd of program.commands) walk(cmd, "");
+
+if (process.argv.includes("--flags")) {
+  const longs = new Set();
+  for (const c of commands) for (const o of c.options) if (o.long) longs.add(o.long);
+  for (const f of [...longs].sort()) console.log(f);
+} else {
+  console.log(JSON.stringify(commands, null, 2));
+}
--- a/ix-cli/scripts/dump-cli-surface.mjs
+++ b/ix-cli/scripts/dump-cli-surface.mjs
@@ -13,7 +13,7 @@
  * because that is what actually ships.
  *
  *   node scripts/dump-cli-surface.mjs            # JSON to stdout
- *   node scripts/dump-cli-surface.mjs --flags    # one long flag per line, sorted
+ *   node scripts/dump-cli-surface.mjs --flags    # one "<command>\t<flag>" pair per line, sorted
  *
  * Deliberately NOT wired into CI: what a parity gate should block on is the
  * open question in #576, and this is the inventory half either answer needs.
@@ -37,7 +37,12 @@ try {
 }
 
 const program = new Command();
-program.name("ix");
+// Mirror main.ts's root registration. `--version` / `-V` is declared on the
+// program itself (main.ts:64), not on any subcommand, so without this the one
+// place a flag can be added with no per-command file changing is also the one
+// place this inventory cannot see it.
+const pkg = createRequire(import.meta.url)("../package.json");
+program.name("ix").version(pkg.version ?? "0.0.0");
 registerOssCommands(program);
 
 const commands = [];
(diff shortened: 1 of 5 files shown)


Alot1z:
Measured confirmation from the other half of #576, for the record: I built `ix-cli` from main and ran the part-2 gate (PR #584) against this branch's flags.md. The registered tree and the reference are in exact parity — 40 commands, 196 flag rows, 0 missing, 0 stale — and mutating the doc (removing a `--kind` row, adding a `--bogus` row) fails the gate naming the flag each time. The 76/76 coverage claim holds against the real registered surface, not just the --flags set.

KageBinary:
Pushed `4b8523d`: a section for **`ix help [topic]`**, which was the one registered command this reference missed.

It surfaced while adding the command-coverage direction to #584's parity gate (pushed there as `b7cdb63`). With that direction in place, the gate reports exactly one gap against this file:

```
$ node scripts/check-doc-parity.mjs --doc skills/ix/references/flags.md
missing doc section: ix help
check-doc-parity: 1 undocumented command(s), 0 missing flag row(s), 0 stale
exit 1
```

and with this commit:

```
check-doc-parity: 54 commands, 196 flags — parity
exit 0
```

`ix help` is a real command, not commander's generated help — `registerWorkflowsHelpCommand` (`src/cli/commands/workflows.ts:68`) registers it with an action, the `workflows`/`workflow` and `advanced` prose topics, per-command help lookup, and the collapsed-plural forwarding that sends `goals`→`goal` and `bugs`→`bug` (Ix-pro#103/#108). An unrecognised topic exits non-zero. The section documents all of that; the command takes no flags, so it uses the same `No flags.` form as `ix init` and the `ix config` family.

Coverage is unchanged at 196 flag rows — this adds a command section, not flags. Merge order with #584 is free in both directions: that gate is inactive while this file is absent, and this file is inert without the gate.

ix-cli/scripts/dump-cli-surface.mjs
```diff
@@ -0,0 +1,73 @@
...
+    subcommands: (cmd.commands ?? []).map(c => c.name()),
+  });
+  for (const sub of cmd.commands ?? []) walk(sub, name);
+}
+for (const cmd of program.commands) walk(cmd, "");
+
+if (process.argv.includes("--flags")) {
+  const longs = new Set();
+  for (const c of commands) for (const o of c.options) if (o.long) longs.add(o.long);
+  for (const f of [...longs].sort()) console.log(f);
```
josephismikhail: **`--flags` cannot serve as the parity check it is advertised as.** It de-duplicates long names across the whole tree, collapsing the 196 (command, flag) pairs this script walks into 76 names. Per-command drift is therefore invisible: add `--limit <n>` to `ix stats`, or drop `--path` from `ix contains`, and this output stays byte-identical while `flags.md` is a row stale — the exact failure mode #575 filed, and the check #576 would build on this. Emitting the pair catches both directions and still sorts stably:

```suggestion
  const pairs = new Set();
  for (const c of commands) for (const o of c.options) if (o.long) pairs.add(`${c.command}\t${o.long}`);
  for (const p of [...pairs].sort()) console.log(p);
```

ix-cli/scripts/dump-cli-surface.mjs
```diff
@@ -0,0 +1,73 @@
...
+    arguments: (cmd.registeredArguments ?? []).map(a => ({
+      name: a.name(),
+      required: a.required,
+    })),
+    options: (cmd.options ?? []).map(o => ({
+      flags: o.flags,
+      long: o.long,
+      short: o.short,
+      description: o.description,
+      default: o.defaultValue === undefined ? null : String(o.defaultValue),
```
josephismikhail: **Negated options dump the wrong default, and the table inverts because of it.** Commander leaves `defaultValue` `undefined` on a `--no-x` option and supplies `true` at parse time instead (verified: a command with only `.option("--no-recursive", ...)` yields `{ recursive: true }` when the flag is absent). So `--no-recursive` and `--no-open` come out as `default: null` here, and `flags.md` renders them as Default `off` (lines 308 and 603) — which by this file's own legend ("`off` for a boolean") states the opposite of the truth: a reader of the `ix view start` row concludes no browser is opened, and one is. The prose bullet at flags.md:48 patches it by hand, but the generator stays wrong, so a gate that ever compares defaults inherits the inversion. `o.negate && o.defaultValue === undefined ? "true" : ...` fixes it at the source.

ix-cli/scripts/dump-cli-surface.mjs
```diff
@@ -0,0 +1,73 @@
...
+      choices: o.argChoices ?? null,
+    })),
+    subcommands: (cmd.commands ?? []).map(c => c.name()),
+  });
+  for (const sub of cmd.commands ?? []) walk(sub, name);
+}
+for (const cmd of program.commands) walk(cmd, "");
```
josephismikhail: **The root program's own options are never walked.** Starting at `program.commands` means anything declared on `program` itself is dropped. The shipped CLI declares `.version(cliVersion)` on the root in `src/cli/main.ts:64`, so `ix --version` / `-V` is a real flag that appears in neither the JSON nor `--flags`, while flags.md:3 claims "Every flag the OSS CLI registers". The forward-looking cost is the one that matters for #576: a global option added to `program` in main.ts would be invisible to a gate built on this output, so the one place a flag can be added without any per-command file changing is also the one place the inventory cannot see.

ix-cli/scripts/dump-cli-surface.mjs
```diff
@@ -0,0 +1,73 @@
...
+ * because that is what actually ships.
+ *
+ *   node scripts/dump-cli-surface.mjs            # JSON to stdout
+ *   node scripts/dump-cli-surface.mjs --flags    # one long flag per line, sorted
+ *
+ * Deliberately NOT wired into CI: what a parity gate should block on is the
+ * open question in #576, and this is the inventory half either answer needs.
+ */
+import { Command } from "commander";
+import { createRequire } from "node:module";
```
josephismikhail, line 22: `createRequire` is imported and never used — the module is loaded with dynamic `import()` on line 31 instead. Nothing catches it: `npm run lint` is `eslint src`, and knip's `project` is `src/**/*.ts`, so a script under `ix-cli/scripts/` is outside both. Drop the line.

```suggestion
```

skills/ix/references/flags.md
```diff
@@ -0,0 +1,625 @@
...
+| `--path` | `<path>` | — | Prefer symbols from files matching this path substring |
+| `--pick` | `<n>` | — | Pick Nth candidate from ambiguous results (1-based) |
+| `--format` | `text\|json\|llm` | `text` | Output format — see [output-formats.md](output-formats.md) |
+| `--raw` | — | off | Show raw metadata dump (legacy format) |
+
+### `ix help [topic]`
+
+Additional help topics. With no topic, prints the top-level help. `workflows`
+(or `workflow`) and `advanced` are prose topics; any other topic is looked up as
+a registered command and shows that command's help. The retired plurals `goals`
```
josephismikhail, line 263: **This claim is false on the surface the file scopes itself to.** "...forward to `goal` and `bug`, whose help documents the subcommand that absorbed them" only holds with @ix/pro installed. On OSS, `goal`/`bug` exist solely as the `registerProStubs` placeholders (`.argument("[args...]")`, no subcommands), so `workflows.ts` finds the stub and `outputHelp()` prints:

```
Usage: ix bug [options] [args...]

Manage bugs

Options:
  -h, --help  display help for command
```

Verified against a local `dist/cli/main.js`. An agent following this paragraph runs `ix help bugs` expecting to be told about `ix bug list` and learns nothing. Worth a clause noting the forwarding only pays off when Pro is installed.

skills/ix/references/flags.md
```diff
@@ -0,0 +1,625 @@
...
+## Flags with behaviour worth knowing
+
+- **`ix diff` refuses conflicting volume flags rather than picking a winner.**
+  `--summary` ignores both `--limit` and `--full`, and `--full --limit` is
+  rejected on its own; each combination errors with which flag to drop.
+- **`ix reset` is global.** `--code` narrows *what kind* of data goes, not
+  which workspace — see the Gotchas section of [commands.md](commands.md)
+  before running it in a shared backend.
+- **`--detailed` on `ix subsystems` requires `--list`**, and `--offset`
+  disables the auto-pagination that `--limit` otherwise drives.
```
josephismikhail: **Two things off in this bullet, and the omission is the one that bites.** (1) `--limit` does not drive auto-pagination: `loadSubsystemScores` pages whenever `--detailed` is set and `explicitPage` is false, i.e. neither `--offset` **nor `--regions`** was passed (`subsystems.ts:527`); `--limit` only sets page size, and `--detailed` alone still auto-pages. (2) The constraint an agent actually trips is the reverse of the documented one: `parseDetailedListQuery` rejects `--limit`, `--offset`, `--regions`, `--edge-cap` and `--member-file-cap` unless `--detailed` is present (`subsystems.ts:487`, "--limit, --offset, --regions, --edge-cap, and --member-file-cap require --detailed."). So `ix subsystems --list --limit 50` — the natural read of the `--limit` row at line 533, whose text mentions only "detailed mode" as context — exits with an error rather than a shorter list.

PR #580, opened 2026-09-02
fix: drop the trailing slash from the remaining artifact ignore rules

KageBinary:
**Rescoped after #569 merged.** That PR landed the untrack and the `node_modules` rule an hour ago, so this is now only the part it stopped short of.

## What is left

#569 states the rule exactly — a trailing slash matches **directories only**, so it does not cover a *symlink* named `node_modules`, which is how an absolute path into a contributor's home directory came to be committed. It fixed that one line. The three below it kept the slash:

```
core-ingestion/dist/
ix-cli/dist/
ix-cli/coverage/
```

so the identical defect stayed open under a different name:

```console
$ ln -s /tmp ix-cli/dist
$ git status --short -- ix-cli/dist
?? ix-cli/dist
```

Not a hypothetical shape: a symlinked `dist` is how a git worktree borrows a build from the main clone — the same practice that produced the `node_modules` symlink in the first place.

## Change

Three characters of `.gitignore`, plus the comment above them extended to say the reasoning covers build outputs too.

And #569's regression test now probes **all four** rules rather than `node_modules` alone — it would have stayed green through every line of this. One wrinkle worth the extra line: an existing directory matches a trailing-slash rule perfectly well, so probing `ix-cli/dist` where a real build lives proves nothing. The test asks about a non-existent sibling in that case.

## Verified

```console
$ git check-ignore -v --no-index <each rule>
.gitignore:28:node_modules          ix-cli/__ix_ignore_probe__/node_modules
.gitignore:29:core-ingestion/dist   core-ingestion/dist
.gitignore:30:ix-cli/dist           ix-cli/dist
.gitignore:31:ix-cli/coverage       ix-cli/coverage
```

With real symlinks at all three paths, `git status` reports only this change; before it, each showed as untracked.

## One thing for a follow-up, not this PR

#569's `builds the child CLI somewhere git cannot see` test cannot run in a checkout whose `ix-cli/node_modules` is itself a symlink:

```console
```

`git check-ignore` refuses any path beyond a symlink, so `explain()` returns `""` and the `cacheRoot` assertion fails. Identical on unmodified `main`, so it is not from this change — and it is unreachable in CI, where `npm ci` creates a real directory. But it fails in exactly the worktree layout #569 was diagnosing, which is worth knowing before someone chases it as a flake.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,10 +20,15 @@ target/
 # ix-cli/node_modules in #545 and stayed tracked until #567. Without the slash
 # the rule matches the entry whatever its type, at any depth, which also makes
 # the two package-specific lines that used to sit here redundant.
+#
+# The same applies to the build outputs below. A symlinked `dist` is how a git
+# worktree borrows a build from the main clone -- the same practice that
+# produced the node_modules symlink -- so a directory-only rule there leaves
+# exactly the defect above open under a different name.
 node_modules
-core-ingestion/dist/
-ix-cli/dist/
-ix-cli/coverage/
+core-ingestion/dist
+ix-cli/dist
+ix-cli/coverage
 
 # watch-dedup.test.ts used to build a throwaway CLI into
 # ix-cli/.watch-child-runtime-*; it now builds inside node_modules/.cache,
--- a/ix-cli/src/cli/__tests__/watch-dedup.test.ts
+++ b/ix-cli/src/cli/__tests__/watch-dedup.test.ts
@@ -183,12 +183,29 @@ describe("canonical watch refresh", () => {
       // committable `ix-cli/node_modules` symlink of #545 with it. A path that
       // does not exist is not a directory as far as git is concerned, so only a
       // rule without the trailing slash matches it.
-      const nonDirectory = path.join(packageRoot, "__ix_ignore_probe__", "node_modules");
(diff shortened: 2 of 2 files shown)


KageBinary:
Heads-up: **#569 does the same untrack**, arrived at independently while chasing #567 (its finding 3). The two conflict in `.gitignore`:

```console
$ git merge-tree --write-tree origin/fix/567-watch-dedup-temp-dir origin/fix/untrack-node-modules-symlink
CONFLICT (content): Merge conflict in .gitignore
```

The difference is scope. #569 drops the trailing slash from `node_modules` only; this PR drops it from `core-ingestion/dist`, `ix-cli/dist` and `ix-cli/coverage` too, which have the identical defect — on #569's branch a symlinked `ix-cli/dist` is still committable (`?? ix-cli/dist`), and a symlinked `dist` is how a worktree borrows a build from the main clone, the same practice that produced the `node_modules` symlink in the first place.

#569 also adds `ix-cli/.watch-child-runtime-*`, which this PR does not and should if it is the one that lands.

Happy either way — I have no attachment to this one landing, only to it landing once. If #569 goes first I will close this; if this goes first, #569 drops its finding 3 and keeps the watch-child line.

PR #581, opened 2026-09-02
fix(deps): override fast-uri to 3.1.7, clearing four HIGH advisories

josephismikhail:
The `Trivy (repo)` check went red on every open PR today, including branches that touch no dependency at all.

## It is not a regression from any open PR

`fast-uri@3.1.5` is already on `main`, and four advisories against it were published at once:

| CVE | Title |
|---|---|
| CVE-2026-75899 | SSRF via repeated hostname percent-decoding |
| CVE-2026-75931 | Host confusion via skipped IDN canonicalization |
| CVE-2026-75975 | SSRF via malformed IPv6 normalization |
| CVE-2026-76172 | URI parsing flaw enabling SSRF and redirects |

```console
$ git show origin/main:ix-cli/package-lock.json | grep -A1 '"node_modules/fast-uri"'
"version": "3.1.5"
```

#569 and #571 touch no lockfile at all and are red; #572 was green only because it happened to run before the advisory landed, and went red on its next push. So this is repo-wide and independent of the work in flight.

## Why it cannot be bumped directly

It is transitive, three levels down, and nothing in this repo depends on it:

```
@ix/cli
`-- @modelcontextprotocol/sdk@1.30.0
  +-- ajv-formats@3.0.1
  | `-- ajv@8.20.0
  |   `-- fast-uri@3.1.5
  `-- ajv@8.20.0
    `-- fast-uri@3.1.5
```

## The fix

`ajv@8.20.0` declares `fast-uri: ^3.0.1`, so **3.1.7 satisfies the range its own dependent asked for** — this is a bump inside that range, not a pin fighting it, which is why it needs no patched fork and carries no compatibility risk from ajv's side. Trivy's own report names 3.1.6 as the first fixed 3.x; 3.1.7 is the current patch.

An `overrides` entry is how this repo already handles exactly this situation for `vite`:

```json
"overrides": {
  "vite": ">=7.3.2 <8.0.0",
  "ajv": {
    "fast-uri": "^3.1.7"
  }
}
```

**Scoped under `ajv`, not tree-wide** (review finding). A bare `"fast-uri"` override applies to every request at any depth regardless of who asked, and `fast-uri`'s `latest` tag is already 4.1.4 — 3.x survives only under the `three` tag. The first dependency that moves to `^4` would be silently resolved back to 3.1.7 and run against a major it never declared support for, failing at import rather than at install. Forward hazard only: both `ajv` paths ask for `^3.0.1` today. (npm rejects the `"ajv > fast-uri"` spelling with `EINVALIDTAGNAME`; the nested object is the supported form.)

## Verification

```console
$ npm ls fast-uri --all
@ix/cli@0.10.5
`-- @modelcontextprotocol/sdk@1.30.0
  +-- ajv-formats@3.0.1
  | `-- ajv@8.20.0
  |   `-- fast-uri@3.1.7 deduped
  `-- ajv@8.20.0
    `-- fast-uri@3.1.7
```

One deduped copy, no second tree.

| check | result |
|---|---|
| `mcp-runner.test.ts` — the only consumer of the SDK that pulls ajv | **25/25** |
| `npx vitest run` (full) | **1487 passed, 3 failed, 21 skipped** — unchanged from `main` |
| `npm audit` | **no vulnerabilities at all** |
| lockfile diff | **3 lines** — version, resolved, integrity |

The lockfile is regenerated with `--package-lock-only` from `main`'s. A plain `npm install` also refreshed a dozen unrelated transitives (@babel/parser, esbuild, …), turning a security bump into a 1,091-line diff; that is reverted.

## Scope

Deliberately its own PR. It shares no file with the three issue fixes in flight (#569, #571, #572), and a dependency bump should be revertable on its own.


KageBinary:
Same conclusion, and yours is the better shape — closing #579 in favour of this. I had the lockfile bump only; the `overrides` entry is what makes it survive the next regeneration, and it matches how `vite` is already handled here.

Two things I verified independently that are worth having on the record, since our lockfile hunks are byte-identical:

**`npm ci` from a clean directory** against this lockfile: exit 0, installs `fast-uri 3.1.7`. That is itself the integrity check — `npm ci` verifies the hash and fails hard on a wrong one — and the hash matches what the registry serves:

```console
$ curl -s https://registry.npmjs.org/fast-uri/3.1.7 | jq -r .dist.integrity
sha512-dOvZVzjdZdz7phd9v6jCbwxrBW3fK6n8Rc0CtdmM4bumzMnxywBYhuph6J819RRw/ku+rLbelwfMunktuzVVHg==
```

**The CLI suite against that install**, not the pre-existing one: 89 files, 1509 passed, 2 skipped, parser smoke passed. Worth doing rather than assuming, because ajv is what pulls `fast-uri` in and the `ix-context-bundle/1` validation tests exercise it.

One note in case it surprises someone later: the lockfile's root `packages[""]` carries no `overrides` field, here or on `main` — the existing `vite` override is not recorded there either. So `npm ci` installs the pinned tree as written and the override only asserts itself when the lockfile is regenerated. That is exactly the durability you want from it, but it does mean the override is not what makes *this* PR's `npm ci` produce 3.1.7 — the lockfile hunk is.

Trivy is green on the branch. Also confirms the finding was repo-wide: it fired on a branch of mine that touches only markdown.

PR #584, opened 2026-09-03
ci(parity): gate registered flags against the flag reference (#576)

Alot1z:
Implements part 2 of #576: a committed CI gate so a registered command or flag cannot merge without a doc mention. Part 1 (the inventory half — `scripts/dump-cli-surface.mjs` + `skills/ix/references/flags.md`) ships in #578; this is the compare half, and it is built to be green before and after that merge.

## What the gate does

`ix-cli/scripts/check-doc-parity.mjs` walks the *registered* commander tree from `dist/` (the same surface dump-cli-surface.mjs prints — never a regex over source) and compares both directions against the flag reference:

- a registered long flag with no doc row -> fail, flag and command named
- a doc row for a flag the CLI no longer registers -> fail, stale row named
- `--help` is excluded: commander auto-registers it on every command and the reference deliberately carries no help rows

## Inactive until the reference exists

While `skills/ix/references/flags.md` is absent the gate exits 0 with a notice. With zero coverage it would only report the whole command surface; it activates the moment #578 merges and is strict from then on.

## Verified locally against the real surface

Built `ix-cli` from main and ran the gate against the exact flags.md from #578's branch:

- parity: 40 commands, 196 flags — exit 0
- mutation (removed the `--kind` rows): exit 1, 18 rows named
- mutation (added a stale `--bogus` row): exit 1, row named
- main state (no flags.md): exit 0, inactive notice

## Policy question left open (from #576)

The RFC asked what the gate blocks on. This PR wires it into the `CI Passed` aggregate, so it blocks once active. If maintainers prefer a report-only first window, the inactive branch is the only skip point and it is a one-line change.


KageBinary:
Reviewed at `1191c1a`. The core decision is the right one and it is the reason this is worth having: it reads the **registered commander tree from `dist/`**, so the gate cannot drift from what ships the way a regex over source would. Comparing both directions is also right. I reproduced the headline claim against #578's actual `flags.md`:

```console
$ node ix-cli/scripts/check-doc-parity.mjs
check-doc-parity: 40 commands, 196 flags — parity with skills/ix/references/flags.md
exit=0

$ grep -v '^| `--kind`' flags.md > flags.md   # the mutation from the description
missing doc row: --kind (on ix trace)
missing doc row: --kind (on ix context)
check-doc-parity: 18 missing, 0 stale
exit=1
```

The CI wiring is correct too, including the part that is easy to do halfway: `doc-parity` is added to `ci-success`'s `needs` **and** to the shell condition, and the build step runs before the gate.

Three findings in the doc parser, one of which lets a real gap through.

---

### 1. A non-command section after the command list is attributed to the last command — and can mask a missing row

`current` is reset only when a heading matches `^(#{2,4}) \`…\`` and the backticked text does not start with `ix `. A **plain** heading — `## Notes`, `## Conventions`, anything without backticks — does not match that regex at all, so `current` keeps pointing at the final command section and every backticked flag below it is recorded as documented for that command.

The harmless direction is a false failure:

```console
$ printf '\n## Notes\n\nAvoid `--bogus-flag` and prefer `--kind`.\n' >> flags.md
stale doc row: --bogus-flag (under ix watch)
stale doc row: --kind (under ix watch)
exit=1
```

The direction that matters is the opposite one. Delete a genuine row and mention that flag in trailing prose:

```console
# remove the real `--path` row from the `ix watch` section, then:
$ printf '\n## Notes\n\nThe `--path` option is unchanged.\n' >> flags.md
check-doc-parity: 40 commands, 196 flags — parity
exit=0                       <- the missing row is masked

# control: same deletion, no Notes section
missing doc row: --path (on ix watch)
exit=1
```

So a normal documentation edit — appending a notes or conventions section — either breaks CI naming the wrong command, or silently satisfies the gap the gate exists to catch. Which one you get depends on whether the prose happens to name a flag the last command registers.

**Fix:** reset `current = null` on *any* heading line, not only backticked ones. One extra branch before the existing `head` match.

### 2. A stale row under a flagless command is invisible

```js
const reg = registered.get(cmd);
if (!reg) continue; // command exists but registers no long flags
```

`registered` only gets an entry when `longs.size > 0`, so every command whose section says "No flags." is exempt from the stale check:

```console
# replace `#### ix config show`'s "No flags." with a table row for a flag that does not exist
| `--totally-invented` | — | off | does not exist |

$ node ix-cli/scripts/check-doc-parity.mjs
check-doc-parity: 40 commands, 196 flags — parity
exit=0
```

That is 15 of the 54 documented sections in #578's reference. The `continue` is guarding against a real thing — `documented.get(cmd)` for a flagless section is an empty set, and iterating it is a no-op — but the guard is broader than the problem: an empty set needs no special case, and skipping the command entirely is what creates the hole. Dropping the `continue` (leaving `const reg = registered.get(cmd) ?? new Set()`) closes it with no other change.

### 3. `--doc` with no value fails open

```console
$ node ix-cli/scripts/check-doc-parity.mjs --doc
check-doc-parity: no flag reference at skills/ix/references/flags.md — gate inactive
exit=0
```

`process.argv[idx + 1]` is `undefined`, `existsSync(undefined)` is `false`, and the run takes the inactive branch and passes. `--doc` is the documented testing path, so the mistake it is most likely to catch is the one it hides. An explicit check that the value is present and exists — erroring rather than deactivating — would do it.

---

### Smaller

**The `ci-success` comment is now orphaned.** The new job is inserted between `ci-success`'s explanatory comment and `ci-success:` itself, so this now reads as describing `doc-parity`:

```yaml
  # One stable required-status-check context. Branch protection requires only
  # this job, so the matrix can be reshaped without orphaning required checks.
  # Fails if any upstream job failed OR was skipped/cancelled.
  # ── Flag-reference parity gate (#576) ──────────────────────────────────
  doc-parity:
```

Move the new job above that block, or below `ci-success`.

**The `--help` exclusion is inert.** Commander keeps its auto help option off `cmd.options`:

```console
$ node -e "const {Command}=require('commander'); const c=new Command().option('--x','x'); console.log(c.options.map(o=>o.long))"
[ '--x' ]
```

Harmless, and defensible as future-proofing against `.addHelpOption`, but the comment states it as something that happens today. Worth either dropping or rewording, since #578's generator has the same shape and neither emits help rows.

### On the policy question you left open

Wiring it into `CI Passed` is the right default — a report-only gate on documentation is a gate nobody reads. The thing I would flag instead is the *deactivation* branch: once #578 lands, "no `flags.md`" stops being a bootstrapping state and becomes a way to pass by deleting the reference. Worth turning into a failure at that point, or asserting the file's existence separately from parity.

That also makes the merge order load-bearing in one direction only: #584 before #578 is green-but-inert, and if #578 never lands the gate stays inert forever with nothing saying so.

KageBinary:
Verified the central claim independently, and it holds exactly. Built `ix-cli` from this branch and ran the gate against `skills/ix/references/flags.md` taken from #578's head:

```
$ node scripts/check-doc-parity.mjs                       # main state, no flags.md
check-doc-parity: no flag reference … gate inactive until the reference ships
exit=0

$ node scripts/check-doc-parity.mjs --doc <#578's flags.md>
check-doc-parity: 40 commands, 196 flags — parity
exit=0
```

Same 40/196 you reported. The inactive-until-the-reference-exists design is the right call and it means this can merge before or after #578 without turning `main` red — which is the thing I most wanted to check.

## One real gap: 14 of 54 commands are invisible to the gate

The PR title and #576 both say a registered **command or flag** cannot merge without a doc mention. As written this gates flags only. `registered` is populated with `if (longs.size > 0) registered.set(name, longs)`, so a command with no long flags is never compared against anything — it lands in `allCommands`, which is used only for the stale-section direction.

Enumerated from the built tree on this branch:

```
total commands: 54
commands with ZERO long flags (ungated): 14
  ix init            ix docker          ix docker start    ix docker status
  ix docker restart  ix config          ix config show     ix config get
  ix config set      ix view stop       ix view status     ix savings reset
  ix mcp             ix help
```

So `ix init` — or any new flagless command — can merge today with no doc section at all, green.

**The fix is free.** I checked all 14 against #578's `flags.md`: **13 already have sections**, and the fourteenth is `ix help`, which commander auto-registers and which the reference deliberately omits — the same exemption `--help` already gets. So adding the direction costs one loop and stays green today:

```js
for (const cmd of allCommands) {
  if (cmd === "ix help") continue;          // auto-registered, like --help
  if (!documented.has(cmd)) missing.push(`section for ${cmd}`);
}
```

That closes the half of #576 the title claims.

## Smaller

- **`--doc` as the last argument crashes.** `process.argv[docFlagIndex + 1]` is `undefined`, and `existsSync(undefined)` throws `TypeError` rather than printing a usage error. Testing-only path, but it is the one flag the script defines. #587's `check-api-parity.mjs` has the same `arg()` shape.
- **Root-level options are out of scope.** `walk` is only called for `program.commands`, so any global option registered on the root program is neither gated nor stale-checked. There are none today (I checked: the root registers no long options), so this is a latent gap rather than a bug — worth a comment saying so, since "the registered surface" reads as if it includes them.
- **Hidden options.** If anyone adds `new Option(...).hideHelp()`, `cmd.options` still carries it and the gate will demand a doc row for a flag that is deliberately undocumented. Skipping `o.hidden` alongside `--help` would pre-empt that.
- **No trailing newline** at EOF. `npm run lint` only covers `src`, so nothing catches it — same in #590.

## Policy question you left open

You asked whether the gate should block. I would say yes and keep it wired into `ci-success` as you have it — a report-only window for a check that is green on day one just gives it time to rot. Maintainers' call, but there is nothing to phase in here.

## Merge coordination

This, #587 and #590 each add a leg to the same `ci-success` `needs:` list; two of the three will need a one-line rebase. Each new leg is also a new required context — worth confirming branch protection tracks `CI Passed` only (the file's comment says it does), otherwise the added jobs are advisory.

With the command direction added I would be happy to see this land.

KageBinary:
I've pushed the command-coverage direction to this branch (maintainer edits are enabled) — `b7cdb63`. Force-push over it or drop the commit if you'd rather take it differently; it's your PR.

**One correction to my review first.** I said the 14th flagless command, `ix help`, was "auto-registered like `--help`" and could be excluded for free. That's wrong. `ix help [topic]` is a **real Ix command** — `registerWorkflowsHelpCommand` in `src/cli/commands/workflows.ts:68`, with its own action, its own prose topics (`workflows`/`workflow`, `advanced`), and the collapsed-plural forwarding that sends `goals`→`goal` and `bugs`→`bug` (Ix-pro#103/#108). Commander's generated help command is a different object and isn't in `program.commands` here; I checked.

So the fix isn't quite free — but that's the gate *working*. `ix help` genuinely has no section in the reference, and a parity gate that surfaces it is doing its job. Rather than carve out an exclusion, I've documented the command: the `ix help` section is pushed to **#578**, which owns `flags.md`.

## What the commit does

- Every registered command name now needs a section, not just the flag-bearing ones.
- **Nothing is excluded.** `--help` is skipped on the flag side because commander puts it on every command; there is no command-side equivalent, so there's no carve-out to write.
- A command missing its section is reported **once, as the section it needs**, rather than once per flag it registers — five missing rows for one absent heading buries the single edit that fixes them.
- The summary line now counts all 54 registered commands rather than the 40 flag-bearing ones, which is what the gate actually checks.
- Also added the trailing newline at EOF, since I was in the file.

## Verified against #578's reference, built from this branch

| case | result |
|---|---|
| parity (reference carries the `ix help` section) | `54 commands, 196 flags` — **exit 0** |
| reference *without* that section | `missing doc section: ix help` — exit 1 |
| delete a **flagless** section (`ix init`) | `missing doc section: ix init` — exit 1 |
| delete a **flag-bearing** section (`ix explain`, 5 flags) | **one** missing section, not 5 rows — exit 1 |
| delete a single flag row (`--pick`) | `missing doc row: --pick (on ix explain)` — exit 1 |
| add a stale row (`--bogus`) | `stale doc row` — exit 1 |
| add a stale section (`ix retired-thing`) | `stale section` — exit 1 |
| no `flags.md` at all (main today) | inactive notice — **exit 0** |

The first and last rows are the ones that matter for merging: your inactive-until-the-reference-exists design is preserved exactly, so this stays green on `main` regardless of merge order with #578. Both orders are safe — the gate is inactive while `flags.md` is absent, and `flags.md` is inert without the gate.

## Two things I left alone

Both pre-existing, both yours to take or leave:

1. `--doc` as the final argument still crashes — `process.argv[i + 1]` is `undefined` and `existsSync(undefined)` throws a `TypeError` rather than printing usage. Same `arg()` shape in #587's `check-api-parity.mjs`.
2. The inactive notice hardcodes `skills/ix/references/flags.md` even when `--doc` pointed somewhere else, so `--doc /tmp/missing.md` reports the wrong path. I only noticed it because it showed up in my own mutation runs.

Alot1z:
Both leftovers are fixed in `fac2317` (pushed to `ci/576-docs-parity`).

**1. `--doc` as the final argument** — now a usage error, not a TypeError. `process.argv[i + 1]` was `undefined`; a missing value (last arg, or followed by another flag) now prints `error: option '--doc <value>' argument missing` to stderr and exits 1. `--doc=<path>` is accepted alongside the two-arg form. The same unguarded `arg()` shape in `check-api-parity.mjs` (#587) is fixed identically in `548b1b8` on `docs/skill-harness-agnostic`.

**2. Inactive notice path** — the notice now names the path actually checked (`--doc /tmp/missing.md` reports `/tmp/missing.md`, not the hardcoded default).

Battery rerun against #578's reference, built from this branch (all rows from your table reproduced):

| case | result |
|---|---|
| parity (reference carries the `ix help` section) | 54 commands, 196 flags — exit 0 |
| reference without that section | missing doc section: `ix help` — exit 1 |
| delete a flagless section (`ix init`) | missing doc section: `ix init` — exit 1 |
| delete a flag-bearing section (`ix explain`, 5 flags) | one missing section, not 5 rows — exit 1 |
| delete a single flag row (`--pick`) | missing doc row: `--pick` (on `ix explain`) — exit 1 |
| add a stale row (`--bogus`) | stale doc row: `--bogus` (under `ix explain`) — exit 1 |
| add a stale section (`ix retired-thing`) | stale section — exit 1 |
| no flags.md at all | inactive notice naming the `--doc` path — exit 0 |
| `--doc` as final argument | usage error, exit 1 (no TypeError) |
| `--doc=/path` equals form | works, same behavior as two-arg form |

The 54/196 numbers hold against the current reference, so the command-coverage change in `b7cdb63` and the reference on #578 agree.

KageBinary:
## The response commit is not on the branch

Head is `cc138ac`, force-pushed at `2026-09-03T23:26:24Z`. The branch is now
`c961528` + `cc138ac` and nothing else — both `fac2317` (your fix) and
`b7cdb63` (the command-coverage commit I pushed through maintainer edits) are
gone. The rebase onto current main dropped them.

Reproduced against `cc138ac`, from `ix-cli/`:

```
$ node scripts/check-doc-parity.mjs --doc
check-doc-parity: no flag reference at skills/ix/references/flags.md — gate inactive …
exit=0

$ node scripts/check-doc-parity.mjs --doc /tmp/missing-flags.md
check-doc-parity: no flag reference at skills/ix/references/flags.md — gate inactive …
exit=0

$ node scripts/check-doc-parity.mjs --doc=/tmp/missing-flags.md
check-doc-parity: no flag reference at skills/ix/references/flags.md — gate inactive …
exit=0
```

- `--doc` with no value: still unguarded (`const flagsDoc = process.argv[docFlagIndex + 1]`).
- `--doc <missing path>`: notice still names the hardcoded `skills/ix/references/flags.md`.
- `--doc=<path>`: still not parsed — `indexOf("--doc")` is `-1`, so it silently falls back to the default.
- Command coverage: gone. `registered` is still keyed on flag-bearing commands only, so a registered command with no doc section is not reported; only *stale* sections are.
- The EOF newline is missing again.

**One correction to my earlier review.** I said `--doc` as the final argument
throws a `TypeError`. It does not — `existsSync(undefined)` returns `false` in
Node 22 rather than throwing. The actual behaviour is worse than I described: a
mistyped or valueless `--doc` makes the gate print the inactive notice and
**exit 0**, so the check passes silently instead of failing loudly. Same shape,
higher stakes.

The identical `argValue` guard *is* present on #587 (`548b1b8`), so the fix
exists — it just isn't on this branch.

Nothing else to re-review until `fac2317` and `b7cdb63` are back. Say the word
and I'll re-push `b7cdb63`; it's your PR either way.

See also the cross-cutting note on #589 about the `docs(skill): … instead of
Freebuff` commit, which is what forced this rebase.

Alot1z:
Rebuilt after the force-push clobber you caught. The response commits are back on `ci/576-docs-parity` in their original order: `b8f93f5` (your command-coverage direction, recreating `b7cdb63`) with `6e2d3ba` (the `--doc` guard) on top — branch head is `6e2d3ba`. The de-branding commit is dropped from here per your merge-order note; it now lives only on #589.

KageBinary:
The lost commits are back and every finding I raised is fixed. Re-verified on `6e2d3ba`, from a built `ix-cli/`.

## The four parsing findings

| | before | on `6e2d3ba` |
|---|---|---|
| `--doc` with no value | inactive notice, **exit 0** | `error: option '--doc <value>' argument missing`, exit 1 |
| `--doc /tmp/missing.md` | named the hardcoded default path | names `/tmp/missing.md` |
| `--doc=/tmp/missing.md` | silently ignored, fell back to the default | parsed |
| trailing newline | missing | present |

The first row was the one that mattered — a mistyped `--doc` made the gate pass silently rather than fail loudly. That is closed.

`b8f93f5` restores command coverage, so a registered command with no section is reported now, not just a stale section.

## Mutation-tested against the real reference

Pointed it at `flags.md` from #578 (641 lines) rather than a toy fixture. Clean:

```
check-doc-parity: 54 commands, 196 flags — parity with .../flags.md
exit 0
```

Then four mutations, each caught with the right message and the right granularity:

| mutation | result |
|---|---|
| delete one flag row (`--pick`) | `missing doc row: --pick (on ix history)` / `(on ix diff)` — exit 1 |
| add a stale flag row under `ix explain` | `stale doc row: --bogus-flag (under ix explain)` — exit 1 |
| add a stale section (`ix retired-thing`) | `stale doc row: section for ix retired-thing has no registered command` — exit 1 |
| delete the whole `ix explain` section | `missing doc section: ix explain` — **1 undocumented command**, 0 missing rows, exit 1 |

That last row is the behaviour I asked for and it is exactly right: removing a flag-bearing section reports one missing section, not one row per flag it happened to contain.

## The two PRs agree

Worth stating since they were written independently: this gate passes clean against #578's reference with no adjustment on either side — 54 commands and 196 flags, zero gaps in either direction. And the default path still prints the inactive notice and exits 0 where `flags.md` is absent, so `main` stays green whichever order they merge in.

No open findings from me.

Alot1z:
Thank you — re-verified on `6e2d3ba` from a built tree, mutation-tested against the real reference, and the parser findings all closed with red/green pins. The cross-agreement between the two independently written gates (this one and #587's) holding at 54 commands / 196 flags is the strongest signal either could have. Ready for merge on my side.

PR #587, opened 2026-09-03
docs: API-reference parity gate (four surfaces)

### LINKED ISSUE #586: docs/api reference has drifted from the client; add a committed parity gate and a harness-agnostic skill install
## Measured drift (2026-09-03)

A fresh parity check between `docs/api/openapi.yaml` and the client source (`ix-cli/src/client/api.ts`) finds one endpoint out of step in each direction:

- **`GET /v1/decisions` is documented but never called.** The client only ever calls `POST /v1/decisions`, with `limit` / `topic` in the request body (`api.ts:146`).
- **`POST /v1/decisions` is called but not documented.** The reference claims to be generated from the client source, so a doc entry describing a call the client never makes is drift in the other direction.

All other 41 endpoints match (42 documented = 42 called after the fix).

## Proposal

1. **Committed parity gate** — `ix-cli/scripts/check-api-parity.mjs` compares every path+method in `docs/api/openapi.yaml` against every call in `ix-cli/src/client/api.ts` and fails either direction with the endpoint named. The visualizer proxy (`view.ts`) forwards every `/v1/*` request and defines no endpoints, so the client is the complete source of truth. Wired into CI as an `api-parity` job on the aggregate. Verified: 42/42 green; both mutation directions fail rc 1 naming the endpoint.
2. **Fix the drift** — `/v1/decisions` becomes `post` with `limit`/`topic` in a body schema.
3. **Harness-agnostic skill install** — `scripts/install-skill.sh` hardcodes `~/.claude/skills` + `~/.agents/skills`; the README describes the skill as a "Claude Code skill". The skill itself is harness-agnostic, so the installer should probe for Claude Code, Freebuff, Codex CLI, Gemini CLI, Cursor, opencode, VS Code and OpenClaw (the same detection `ix mcp install` uses) and deploy to each found harness skill dir, with `--dry-run` and explicit id selection. Verified on this machine: 7/8 harnesses detected and installed, openclaw correctly skipped.

PR to follow.

Alot1z:
Closes #586.

**API-reference parity gate** (`ix-cli/scripts/check-api-parity.mjs`, CI job `api-parity`)

The reference claims to be generated from the client source but nothing enforced it. The gate compares every path+method in `docs/api/openapi.yaml` against every call in `ix-cli/src/client/api.ts` and fails either direction with the endpoint named — a registered call with no doc row, or a doc row for a call the client never makes. The visualizer proxy (`view.ts`) forwards every `/v1/*` request and defines no endpoints, so the client is the complete source of truth.

Measured drift it caught and this PR fixes: `docs/api/openapi.yaml` documented `GET /v1/decisions`; the client only calls `POST /v1/decisions` with `limit`/`topic` in the body. Now documented as the POST it is.

Verified: 42 documented = 42 called, exit 0; both mutation directions (rename a client call, rename a doc path) fail exit 1 naming the endpoint. The gate parses source + yaml directly, so the job needs no install or build.

Scope note: the harness-agnostic skill installer that originally lived here was narrowed out in f1812a0 — the install surface is now #591 (`feat/skill-install-toolscan`), which owns the verified harness table. This PR keeps the API-reference parity gate it was reviewed for; the README/install-skill.sh edits that remain are the simplification that went with the narrow.

Coordination note: #584 (`doc-parity`) edits the same aggregate `needs` block in `ci.yml`; whichever merges second needs a one-line rebase there.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -290,12 +290,12 @@ jobs:
 
   # ── API-reference parity gate ────────────────────────────────────────────
   # docs/api/openapi.yaml claims to be generated from the client source; this
-  # gate enforces it. The visualizer proxy forwards every /v1/* request and
-  # defines no endpoints of its own, so the client surface
-  # (ix-cli/src/client/api.ts) is the complete source of truth: a client call
-  # with no documented path/method, or a documented endpoint the client never
-  # calls, fails this job with the endpoint named. It parses source + yaml
-  # directly, so it needs no install or build.
+  # gate enforces it across four surfaces: endpoint parity (client calls vs
+  # openapi paths), openapi.yaml validity + $ref integrity, the schema contract
+  # (every schema must exist in types.ts or the explicit allowlist), and the
+  # README Endpoints by Area section. A client call pattern the gate cannot
+  # classify fails loudly rather than passing silently. The gate parses yaml
+  # with the `yaml` package, so ix-cli deps are installed first.
   api-parity:
     name: API-reference parity
     runs-on: ubuntu-latest
@@ -306,6 +306,11 @@ jobs:
       - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
           node-version: "22"
+          cache: "npm"
+          cache-dependency-path: ix-cli/package-lock.json
+      - name: Install (ix-cli)
+        working-directory: ix-cli
+        run: npm ci --silent
       - name: docs/api vs client surface
         working-directory: ix-cli
         run: node scripts/check-api-parity.mjs
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -288,6 +288,28 @@ jobs:
         if: failure()
         run: docker compose -f docker-compose.standalone.yml logs --tail 200
 
+  # ── API-reference parity gate ────────────────────────────────────────────
+  # docs/api/openapi.yaml claims to be generated from the client source; this
+  # gate enforces it. The visualizer proxy forwards every /v1/* request and
+  # defines no endpoints of its own, so the client surface
+  # (ix-cli/src/client/api.ts) is the complete source of truth: a client call
+  # with no documented path/method, or a documented endpoint the client never
+  # calls, fails this job with the endpoint named. It parses source + yaml
+  # directly, so it needs no install or build.
+  api-parity:
+    name: API-reference parity
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: "22"
+      - name: docs/api vs client surface
+        working-directory: ix-cli
+        run: node scripts/check-api-parity.mjs
+
   # ── Aggregate gate ───────────────────────────────────────────────────────
   # One stable required-status-check context. Branch protection requires only
   # this job, so the matrix can be reshaped without orphaning required checks.
@@ -295,7 +317,7 @@ jobs:
   ci-success:
     name: CI Passed
     if: always()
-    needs: [static, test, e2e]
+    needs: [static, test, e2e, api-parity]
     runs-on: ubuntu-latest
     steps:
       - name: Verify all required jobs succeeded
@@ -303,6 +325,8 @@ jobs:
           echo "static=${{ needs.static.result }}"
           echo "test=${{ needs.test.result }}"
           echo "e2e=${{ needs.e2e.result }}"
+          echo "api-parity=${{ needs.api-parity.result }}"
           [ "${{ needs.static.result }}" = "success" ] \
             && [ "${{ needs.test.result }}" = "success" ] \
-            && [ "${{ needs.e2e.result }}" = "success" ]
+            && [ "${{ needs.e2e.result }}" = "success" ] \
+            && [ "${{ needs.api-parity.result }}" = "success" ]
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -500,9 +500,9 @@ Record a decision.
 
 **Response** — `{ status, nodeId, rev }`.
 
-#### GET `/v1/decisions`
+#### POST `/v1/decisions`
 
-List decisions. **Query**: `limit`, `topic`. **Response** — array of `GraphNode`.
+List decisions. **Body**: `limit`, `topic`. **Response** — array of `GraphNode`.
 
 ### Patches & Provenance
 
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -3,9 +3,13 @@
 The Ix backend exposes a JSON-over-HTTP API on **`http://localhost:8090`** (the
 local Docker memory-layer). The `ix` CLI and the Compass visualizer are its two
 primary clients. This reference is generated from the client source
-(`ix-cli/src/client/api.ts`), the visualizer server template
-(`ix-cli/src/cli/commands/view.js`), and the shared client types
-(`ix-cli/src/client/types.ts`).
+(`ix-cli/src/client/api.ts`) and the shared client types
+(`ix-cli/src/client/types.ts`). The visualizer proxy
+(`ix-cli/src/cli/commands/view.ts`) forwards every `/v1/*` request to the
+backend and defines no endpoints of its own. A committed parity gate
+(`ix-cli/scripts/check-api-parity.mjs`, wired into CI) enforces that the
+OpenAPI paths/methods match the client surface, so the reference cannot drift
+from what the CLI actually calls.
 
 ## Table of Contents
 
--- a/docs/api/openapi.yaml
+++ b/docs/api/openapi.yaml
@@ -522,17 +522,18 @@ paths:
                 $ref: "#/components/schemas/NodeCreationResult"
 
   /v1/decisions:
-    get:
+    post:
       tags: [planning]
       summary: List decisions
       operationId: listDecisions
-      parameters:
-        - name: limit
-          in: query
-          schema: { type: integer }
-        - name: topic
-          in: query
-          schema: { type: string }
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                limit: { type: integer }
+                topic: { type: string }
       responses:
         "200":
           description: Decisions
--- a/ix-cli/scripts/check-api-parity.mjs
+++ b/ix-cli/scripts/check-api-parity.mjs
@@ -194,7 +194,7 @@ const apiLines = apiSrc.split(/\r?\n/);
 // excluded.
 const KNOWN_HELPERS = new Set(["runReset"]);
 for (let i = 0; i < apiLines.length; i++) {
-  const call = apiLines[i].match(/this\.([a-zA-Z_$][\w$]*)\(/);
+  const call = apiLines[i].match(/this\.([a-zA-Z_$][\w$]*)(?:<[^>]*>)?\(/);
   if (!call) continue;
   const verb = call[1];
   if (VERBS.has(verb) || KNOWN_HELPERS.has(verb)) continue;
--- a/ix-cli/scripts/check-api-parity.mjs
+++ b/ix-cli/scripts/check-api-parity.mjs
@@ -194,7 +194,7 @@ const apiLines = apiSrc.split(/\r?\n/);
 // excluded.
 const KNOWN_HELPERS = new Set(["runReset"]);
 for (let i = 0; i < apiLines.length; i++) {
-  const call = apiLines[i].match(/this\.([a-zA-Z_$][\w$]*)(?:<[^>]*>)?\(/);
+  const call = apiLines[i].match(/this\.([a-zA-Z_$][\w$]*)(?:<.*?>)?\(/);
   if (!call) continue;
   const verb = call[1];
   if (VERBS.has(verb) || KNOWN_HELPERS.has(verb)) continue;
@@ -212,7 +212,7 @@ for (let i = 0; i < apiLines.length; i++) {
 // yield two tokens that normalize to the same key). A call with no literal
 // path at all is a new, unparseable pattern — fail loud instead of passing.
 for (let i = 0; i < apiLines.length; i++) {
-  const call = apiLines[i].match(/this\.(get|post|put|delete|patch)(?:<[^>]*>)?\(/);
+  const call = apiLines[i].match(/this\.(get|post|put|delete|patch)(?:<.*?>)?\(/);
   if (!call) continue;
   const method = call[1].toUpperCase();
   let tokens = [...apiLines[i].matchAll(/\/v1[^\s"'`]*/g)].map((t) => t[0]);
--- a/ix-cli/scripts/check-api-parity.mjs
+++ b/ix-cli/scripts/check-api-parity.mjs
@@ -1,0 +1,162 @@
+#!/usr/bin/env node
+/**
+ * API-reference parity gate for docs/api.
+ *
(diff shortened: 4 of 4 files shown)


KageBinary:
Reviewed both halves on a machine that has all eight harnesses installed. The API-parity half is good and I verified its central claim independently. The skill-install half targets directories that mostly do not exist, and I think it should not ship in this shape.

## Half 1 — API parity gate: correct, and the drift it fixes is real

I checked the `GET` → `POST /v1/decisions` change against the backend rather than against the client, and it holds:

```scala
// Ix-memory: src/main/scala/ix/memory/api/DecisionRoutes.scala:23
case req @ POST -> Root / "v1" / "decisions" =>
```

`POST` is the only route the backend registers for that path, so the old `GET` row documented an endpoint that has never existed. Good catch, and the gate's four surfaces are a sensible shape.

Two notes, neither blocking:

1. **The new `requestBody` documents 2 of 3 fields.** `DecisionListRequest` is `(limit, topic, branchId)`; the schema omits `branchId`. The gate compares path+method only, so it cannot catch this — worth adding the field while you are in there.
2. **The gate pins the API reference to the CLI's call log.** "Every documented endpoint must be called by the client" means a real backend endpoint that `api.ts` does not happen to call must be *deleted* from `docs/api` to keep CI green. That is the right constraint today (42 = 42), but it inverts what an API reference is for. Consider making the stale-doc direction a warning and keeping the undocumented-endpoint direction fatal.

## Half 2 — harness-agnostic install: five of the six new targets are not read by anything

The premise is that `skills/ix` is harness-agnostic and every harness loads it "natively from its own skills directory". I checked that on this machine, which has all eight harnesses installed:

| harness | directory this PR installs to | what is actually there |
|---|---|---|
| claude | `~/.claude/skills` | ✅ correct |
| agents | `~/.agents/skills` | ✅ correct |
| codex | `~/.codex/skills` | ✅ **correct** — exists, populated |
| cursor | `~/.cursor/skills` | ❌ Cursor's skills live in **`~/.cursor/skills-cursor/`** (populated: `babysit`, `canvas`, `create-hook`, `create-rule`, `create-skill`, …) |
| gemini | `~/.gemini/skills` | ❌ `~/.gemini/` has `extensions/`, `commands`; no skills convention |
| opencode | `~/.config/opencode/skills` | ❌ has `agents/`, `commands/`, `plugins/`; no `skills/` |
| openclaw | `~/.openclaw/skills` | ❌ has `agents/`, `extensions/`, `flows/`; no `skills/` |
| vscode | `~/.vscode/skills` | ❌ `~/.vscode/` has `argv.json`, `cli`, `extensions`; no skills concept |

So on a fully-loaded machine the installer creates and populates **five directories nothing reads**, and prints `Installed: …` for each — the user is told the skill is live in five harnesses where it is inert. The README change states this as fact ("any skill-compatible harness loads the same `skills/ix` tree natively from its own skills directory — Codex CLI, Gemini CLI, Cursor, and more"), which is the part I would most want corrected before merge.

### Why the derivation produces wrong paths

The root cause is that `hosts.ts` is an **MCP host registry**, not a skills registry. Deriving a skills directory from an MCP config path is a category error, and the PR's own `OVERRIDES` table is the evidence — four of the eight ids need a hand-written override precisely because the derivation does not hold.

`vscode` is the clearest case. The override comment says the MCP config path "is platform-dependent, not `~/.vscode`" and then hardcodes `~/.vscode` anyway. `vsCodeUserConfig()` returns `~/.config/Code/User/mcp.json` on Linux, so `~/.vscode` is neither the config dir nor a skills dir — but it *does* exist on every machine that has ever installed a VS Code extension, so it is a presence probe that is always true. `vscode` will be reported as present and "installed" on essentially every developer machine.

### Suggested reshape

Make the skills table explicit and small rather than derived:

```js
// Skills are a harness feature, not an MCP feature. Only list a harness
// once its skills directory is verified.
const SKILL_DIRS = {
  claude: "~/.claude/skills",
  agents: "~/.agents/skills",
  codex:  "~/.codex/skills",
  cursor: "~/.cursor/skills-cursor",
};
```

That keeps everything this PR gets right (dry-run, explicit id selection, the not-a-previous-install guard, the loud-skip discipline) and drops the four speculative targets. Adding a harness stays a one-line edit — and a *deliberate* one, backed by having checked where that harness actually looks.

## Smaller things

- **`install-skill.sh` conflict semantics did change**, contrary to "The not-a-previous-install guard and `--force` semantics are unchanged." It used to `exit 1` on the first foreign skill before writing anything; now it counts the conflict, **continues installing to the other harnesses**, and exits 1 at the end. That is arguably better, but it is a behaviour change and the note says otherwise.
- **Dry-run and the real run disagree on exit code.** With one conflict, `--dry-run` prints `would refuse:` and `exit 0`; the real run exits 1. The PR's stated goal is that "the preview and the real run agree" — the guard now runs in both, but the exit code does not follow.
- **`skill-harnesses.mjs` parses `hosts.ts` by brace-depth counting.** It survives today, but an unbalanced `{` or `}` inside a string or comment in `hosts.ts` silently mis-slices the blocks. Since the file is TypeScript that already gets compiled, exporting the table from a built module (or a small JSON emitted at build time) would be sturdier than re-parsing source.
- **`check-api-parity.mjs` and `check-doc-parity.mjs` both end without a trailing newline.** `npm run lint` only covers `src`, so nothing catches it.

## Merge coordination

This edits the same `ci-success` aggregate as #584 and #590 — three PRs adding a required leg to the same `needs:` list. Whichever two land second each need a one-line rebase, and each new leg is another required check that must be added to branch protection or the aggregate silently loses it.

Happy to re-review once the skills table is narrowed to the directories that are actually read.

Alot1z:
Addressed. The branch now carries only the api-parity half; the skill-install half is removed (`f1812a0`).

## Half 2 — skill-install: moved to #591 with the exact reshape you asked for

You were right that the targets mostly do not exist. Rather than ship a second, narrowed copy of the installer here, #591 (`feat/skill-install-toolscan`, based on main) owns that surface with precisely the table you suggested — explicit and verified, no `~/.vscode` presence probe:

```js
const SKILL_DIRS = {
  claude: "~/.claude/skills",
  agents: "~/.agents/skills",
  codex:  "~/.codex/skills",
  cursor: "~/.cursor/skills-cursor",
};
```

That PR also fixes the smaller items you flagged: dry-run and the real run now agree on exit code (a would-refuse exits 1 in both), the conflict behavior change is documented in the script header, the `hosts.ts` brace-depth parsing is gone (the table above replaces it), and the container smoke job asserts the off-PATH case end to end. This PR removing its half means neither PR can ship the speculative directories if it merges first.

## Half 1 — api-parity, your two notes

1. **`branchId` in the requestBody** — deliberate omission, with evidence. `listDecisions` in `ix-cli/src/client/api.ts` posts only `{ limit, topic }` (`api.ts:145`), and no caller holds a branch context to pass through. The reference's stated contract (README: "generated from the client source") means the requestBody mirrors the client's actual call; documenting `branchId` would re-introduce exactly the documented-but-never-sent drift this gate exists to catch. If a caller ever gains a branch scope, the field lands in the client and the schema together.
2. **Stale-doc direction as a warning** — I kept it fatal, for a system-specific reason: `docs/api` is the CLI's wire contract, and the client is the only consumer of this API in the repo, so "every documented endpoint must be called" is the correct invariant *here*. The inversion you describe is real for a public API reference; this one is an internal mirror, and the fatal direction is what keeps the README's generation claim honest. If the API ever becomes public-facing, flipping that direction is the right change and the gate makes it a one-line edit.

## Smaller things

- Trailing newlines: `check-api-parity.mjs` now ends with one (restored in `f1812a0`); `check-doc-parity.mjs` on #584 does too.
- `--flag` as the final argument in `check-api-parity.mjs` no longer crashes — `548b1b8` applies the same guarded parser as the #584 fix (usage error + exit 1 instead of a TypeError on `process.argv[i + 1]`).

## Merge coordination

Noted — `ci-success` now gains `api-parity` here and `harness-install-smoke` on #591; each PR that lands after the first needs the one-line `needs:` rebase, same as #584/#590. Both PRs are green on their own bases.

Alot1z:
Follow-up (085aacf): the README and install-skill.sh now name the install surface as `Claude Code / .agents` instead of a specific client brand — the skill is harness-agnostic, and `~/.agents` is the neutral name for the surface the script deploys to. The `.freebuff/` ignore pattern in .gitignore is kept as a functional ignore for a directory a local tool writes; only the comment naming the brand was dropped.

KageBinary:
Re-reviewed at `085aacf`. The narrowing landed cleanly — the net diff is the
api-parity gate plus the shared de-branding commit, and the `argValue` guard
(`548b1b8`) is present and works: `--doc` with no value is a usage error on
stderr with exit 1, and the `--doc=<path>` equals form is accepted.

Ran the gate from `ix-cli/`: `42 documented, 42 called, 13 schemas, 13 refs` —
exit 0. Then mutated each of the four surfaces:

| mutation | result |
|---|---|
| client calls `/v1/bogus-probe` | `undocumented endpoint: GET /v1/bogus-probe` — exit 1 |
| `this.get<unknown>(PATHS.stats)` (non-literal path) | `unparsed client call … path is not a literal /v1 string` — exit 1 |
| rename `HealthResponse` in `types.ts` | `schema HealthResponse has no types.ts counterpart` — exit 1 |
| `$ref` → `#/components/schemas/NoSuchSchema` | `openapi $ref targets missing schema` — exit 1 |
| document `/v1/ghost` that nothing calls | `stale doc entry` + README mismatch — exit 1 |
| drop every README mention of `/v1/capabilities` | `openapi endpoint missing from README Endpoints by Area` — exit 1 |

Both directions hold on all four surfaces. One real hole.

## The unclassified-verb guard is blind to the house call style

The header promises that `this.request(...)` is reported so "a new call pattern
cannot quietly escape the gate". The detector at line 197 is:

```js
const call = apiLines[i].match(/this\.([a-zA-Z_$][\w$]*)\(/);
```

No generic-argument slot — unlike the known-verb pass at line 216, which has
`(?:<[^>]*>)?`. Every call in `api.ts` is written `this.get<…>(`, so a new
helper written in the same style walks straight past it. Same injected call,
same line, only the type argument differs:

```
this.request("/v1/whatever")           -> unclassified client call verb: this.request(...) near line 43 — exit 1
this.request<unknown>("/v1/whatever")  -> 42 documented, 42 called, 13 schemas, 13 refs — exit 0
```

Fix is the same token the other pass already uses:

```js
const call = apiLines[i].match(/this\.([a-zA-Z_$][\w$]*)(?:<[^>]*>)?\(/);
```

Worth a pin test, since this guard's whole job is to fail on a pattern that
doesn't exist in the tree yet.

Everything else here I'd merge. Note the `argValue` guard you shipped in
`548b1b8` is *not* on #584 — that branch was force-pushed back past it; details
in my comment there.

Alot1z:
Rebuilt — and you were right to have stopped here. The head you re-reviewed at `085aacf` (narrowed, guard present) was reset past **two** commits when the de-branding landed, not one: `548b1b8` (the `argValue` guard) *and* `f1812a0` (the narrowing itself). The branch sat at `7672b23` — the pre-narrow stack — while the PR body described the narrowed state. Same clobber class that cost #584/#590 their response commits; my rebuild now restores the reviewed shape.

**New head `80d9aa2`** (`docs/skill-harness-agnostic`, rebased on current main `fb46941`):

- `8f27e4a` — API-reference parity gate + `/v1/decisions` drift fix (your Half-1, `6cb1128` content)
- `4813132` — four-surface hardening (`a1da5dc` content)
- `bacdbc5` — `argValue` guard (`548b1b8` content, restored)
- `80d9aa2` — your one remaining finding: the unclassified-verb guard now accepts the generic-argument slot, same token the known-verb pass already uses

Net diff vs main is exactly the four parity files — ci.yml, docs/api/README.md, docs/api/openapi.yaml, check-api-parity.mjs. No skill-install content, no install-skill.sh, no README installer section, no brand commit.

Measured on the rebuilt head:

- parity: `42 documented, 42 called, 13 schemas, 13 refs` — exit 0
- injected `this.request<unknown>("/v1/bogus-probe")` (your exact case): `unclassified client call verb: this.request(...) near line 153 — extend check-api-parity`, 1 gap, exit 1 — previously this shape exited 0
- restored tree: parity again

Re-ran from `ix-cli/` on the rebuilt tree; the gate needs no build (parses source + yaml). The branch name still says `skill-harness-agnostic` (GitHub keeps the head ref), but the title now says what the PR is: the API-reference parity gate. #591 owns the install surface.

CI is re-running on the new head.

KageBinary:
`adadd96` fixes the hole I reported — verified against the real gate on this head, not just by reading the regex:

```
probe: this.request<unknown>("/v1/bogus-probe");
   exit=1  unclassified client call verb: this.request(...) near line 599 — extend check-api-parity
```

That was exit 0 before. Clean tree still reports `42 documented, 42 called, 13 schemas, 13 refs`, exit 0.

## The same hole is still open one step further out

`<[^>]*>` cannot cross a nested `>`, so a generic argument that contains one walks past the guard exactly as the un-slotted version did:

```
probe: this.request<Record<string, unknown>>("/v1/bogus-probe");
   exit=0     (no error at all)
```

Same file, same head, same probe — only the type argument changed. `[^>]*` matches up to the `>` in `unknown>`, then the pattern needs a `(` and finds `>`, so the line does not match and the verb is never classified.

Nothing in `api.ts` uses a nested generic today (checked: zero occurrences), so the gate is not currently lying. But `this.get<Record<string, unknown>>(...)` is ordinary house style for this file, and the whole point of the commit you just landed is that a new helper written in house style must not slip through. Right now one shape does and one does not, which is worse than either, because the gate reads as covering the generic case.

## One character fixes it

Make the generic slot lazy so backtracking finds the closing `>` that is actually followed by `(`:

```js
const call = apiLines[i].match(/this\.([a-zA-Z_$][\w$]*)(?:<.*?>)?\(/);
```

Verified on this head, both directions:

| probe | `<[^>]*>` | `<.*?>` |
|---|---|---|
| `this.request("/v1/x")` | caught | caught |
| `this.request<unknown>("/v1/x")` | caught | caught |
| `this.request<Record<string, unknown>>("/v1/x")` | **misses** | caught |
| `this.request<Map<string, Set<number>>>("/v1/x")` | **misses** | caught |
| `this.get<MapResult[]>("/v1/x")` | caught | caught |

And with the lazy version in place the clean tree still passes: `42 documented, 42 called, 13 schemas, 13 refs`, exit 0. The nested probe then fails with the right message:

```
unclassified client call verb: this.request(...) near line 599 — extend check-api-parity
```

Lazy is safe here — the input is the repo's own `api.ts`, one line at a time, not untrusted text.

If you would rather not depend on backtracking at all, `/this\.([a-zA-Z_$][\w$]*)\s*[<(]/` gets the same coverage by not trying to match the type argument in the first place. It would also flag a comparison like `this.count < n`, which `api.ts` does not contain; a false positive in a gate costs one line in `KNOWN_HELPERS`, while a false negative costs the gate its point.

## Also worth carrying over from #600

Your note there — that this gate cannot see a property added *inside* an already-known schema — is right, and belongs in this PR's own description of what the four surfaces cover. `release_version` on `HealthResponse` drifted for months with the client type already carrying it, and nothing here would have said so. Better stated as a known limit than found later by someone assuming the gate covers response shapes.

Alot1z:
Fixed on the branch — head `f37fd497` is live.

Applied to both call-verb passes, not just the demonstrated one: the generic-verb regex and the HTTP-verb regex both carried `<[^>]*>`, so `this.get<Record<string, unknown>>(...)` was equally invisible. The generic slot is now lazy (`<.*?>`) in both.

Re-verified on this head against the branch's real inputs, both directions, from outside the repo tree:

```
[clean]  42 documented, 42 called, 13 schemas, 13 refs — parity, exit 0
[probe]  this.request<Record<string, unknown>>("/v1/bogus-probe")
         → unclassified client call verb: this.request(...) — extend check-api-parity, exit 1
```

Clean tree still passes because nothing in api.ts uses a nested generic today — the fix only widens what the census can see, and the probe now fails loud with the right message instead of exit 0. Your probe table holds: non-nested shapes unchanged, both nested shapes now caught.

PR #589, opened 2026-09-03
docs(readme): point the Docs link at the in-repo docs

### LINKED ISSUE #588: docs(readme): the Docs nav link points at a 404
## Problem

The README nav block links **Docs** to `https://www.ix-infra.com/docs`, which returns **HTTP 404**. Every variant of the link is dead, and no docs exist on the site or anywhere else external.

## Measured evidence (2026-09-03, curl -L --max-time 15)

| URL | Result |
|---|---|
| `https://www.ix-infra.com/docs` | 404 |
| `https://www.ix-infra.com/docs/` | 404 |
| `https://ix-infra.com/docs` (no www) | 404 |
| `https://ix-infra.com/docs/` | 404 |
| `https://ix-infra.com/docs/index.html` | 404 |
| `https://docs.ix-infra.com` | DNS: no such host |
| `https://www.ix-infra.com` (root) | 200 — no docs route or docs link on the page |
| `https://ix-infra.com/sitemap.xml` | empty (no sitemap) |
| GitHub Pages (`GET /repos/ix-infrastructure/Ix/pages`) | 404 — Pages not configured |

All other 25 URLs in the README return 200.

## Where the docs actually live

The repo's `docs/` directory — `docs/prerequisites.md` and `docs/api/` (README + openapi.yaml) — which the README already links to relatively elsewhere.

## Proposed fix

Point the Docs nav item at the in-repo docs: `https://github.com/ix-infrastructure/Ix/tree/main/docs`. (If the site intends to ship `/docs` later, the link can move back then.) While touching the same nav line: `Demo</a> .` ends with a stray `.` where every other item uses `·`.

I'll send a PR with the two-line change.

Alot1z:
Fixes the README's **Docs** nav link, which points at `https://www.ix-infra.com/docs` and returns **404** (measured: every `/docs` variant 404s, `docs.ix-infra.com` has no DNS, the site has no docs route or link, GitHub Pages is not configured — receipts in #588). The only live docs are in-repo (`docs/prerequisites.md`, `docs/api/`), which the README already links to relatively elsewhere.

Changes (2 lines, README nav block):
- Docs href `https://www.ix-infra.com/docs` → `https://github.com/ix-infrastructure/Ix/tree/main/docs`
- `Demo</a> .` → `Demo</a> ·` (stray period; every other nav item uses `·`)

Closes #588

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,8 @@ target/
 # better kept that way so the intent of each is visible.
 /dist/
 
-# Session/tooling artifacts (Freebuff preview state, generated knowledge graphs)
+# Session/tooling artifacts (preview state, generated knowledge graphs; the
+# desktop tool writes .freebuff/ at session start)
 .freebuff/
 .understand-anything/
 .bsp/
--- a/README.md
+++ b/README.md
@@ -219,10 +219,10 @@ Windows (PowerShell):
 irm https://raw.githubusercontent.com/ix-infrastructure/ix-cursor-plugin/main/install.ps1 | iex
 ```
 
-## Claude Code / Freebuff Skill
+## Claude Code / .agents Skill
 
 [`skills/ix/`](skills/ix/SKILL.md) is a self-contained **Claude Code skill** (also
-loaded natively by Freebuff / Codebuff) that teaches any LLM agent to drive the
+loaded natively through the `.agents` surface) that teaches any LLM agent to drive the
 `ix` CLI: build the graph, explain symbols, trace flows, analyze impact, and
 detect smells — instead of grepping and guessing.
 
@@ -251,7 +251,7 @@ powershell -ExecutionPolicy Bypass -File skills/ix/scripts/bootstrap.ps1 .
 The skill follows the progressive-disclosure layout — a lean `SKILL.md` with the
 full command reference, output-format rules, and troubleshooting split into
 `references/`, plus `scripts/` for first-run setup — so it is portable to Claude
-Code, Freebuff, and any other skill-compatible agent. Edit `skills/ix/` and
+Code, the `.agents` surface, and any other skill-compatible agent. Edit `skills/ix/` and
 re-run `scripts/install-skill.sh` to update your installed copy. To produce a
 distributable zip:
 
--- a/README.md
+++ b/README.md
@@ -21,8 +21,8 @@
 
 <p align="center">
   <a href="https://www.ix-infra.com">Website</a> ·
-  <a href="https://www.ix-infra.com/docs">Docs</a> ·
-  <a href="https://compass.ix-infra.com">Demo</a> .
+  <a href="https://github.com/ix-infrastructure/Ix/tree/main/docs">Docs</a> ·
+  <a href="https://compass.ix-infra.com">Demo</a> ·
   <a href="https://discord.gg/ncEYVHVqZ8">Discord</a>
 </p>
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,6 @@
+# Ix documentation
+
+- [HTTP API reference](api/) — the backend JSON API exposed on `http://localhost:8090`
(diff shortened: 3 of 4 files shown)


KageBinary:
Verified both halves. `https://www.ix-infra.com/docs` is dead and the replacement resolves; the `·` fixes a stray `.` that is inconsistent with the other three separators in the same block. Two lines, no side effects — this is right.

One coordination note: #590 currently allowlists `https://www.ix-infra.com/docs` as a known-dead link, keyed to this PR. Its allowlist is deliberately self-cleaning, so whichever of the two merges second goes red until the allow entry is deleted. Merging this one first and dropping the entry from #590 before it lands is the cleaner order.

KageBinary:
Verified the premise and the fix:

```
https://www.ix-infra.com/docs                           404
https://www.ix-infra.com                                200
https://github.com/ix-infrastructure/Ix/tree/main/docs  200
https://compass.ix-infra.com                            200
```

Dead link replaced with a live one, and the stray `.` separator between Demo
and Discord is fixed on the way past. Good to merge.

One thing worth knowing before it lands: `docs/` has no `README.md`, so the new
link renders as a bare file listing — `docs/api/`, `llm-format.md`,
`mcp-plugin-consolidation.md`, `prerequisites.md`. It works, but a short
`docs/README.md` index would make the nav link land somewhere that reads like a
docs home. Either that, or the site gets a `/docs` page and the link goes back;
maintainers' call which is the real fix.

---

## Cross-cutting: the de-branding commit is in all five open PRs

`docs(skill): reference the .agents surface instead of Freebuff` appears as
`cc138ac` (#584), `085aacf` (#587), `0030960` (#589), `b675596` (#590) and
`b971d2b` (#591) — five copies of the same change to `.gitignore`, `README.md`
and `scripts/install-skill.sh`, unrelated to four of the five PRs' stated
purpose. Whichever merges first forces a rebase on the other four, and that
rebase is what just cost #584 and #590 their response commits (`fac2317` and
`4812866` are both absent from their heads).

Suggest keeping it on exactly one branch — this one is the closest fit, since
it already owns the README nav — or splitting it into its own PR, and dropping
it from the rest.

Two small things about the change itself while it's here:

- `.gitignore` still ignores `.freebuff/`, but the comment above it no longer
  says what produces that directory, so the rule now reads as unexplained.
- `install-skill.sh` now labels both destinations "Claude Code":
  ```
  #   ~/.claude/skills/ix   — Claude Code (global)
  #   ~/.agents/skills/ix   — Claude Code / .agents (global)
  ```
  The second line was there to say *which other thing* reads that path; as
  written the two are indistinguishable. `— the .agents skill surface (global)`
  would keep the distinction without naming a product.

Alot1z:
Thanks — both points taken, and the two small things fixed.

**De-branding now lives on exactly one branch.** Kept on this one per your suggestion (it owns the README nav); dropped from #584, #587, #590, #591 — those branches are back to main's wording, so whichever merges first no longer forces four rebases.

**The two lost response commits are restored.** `fac2317` (the `--doc` parsing guard on #584) and `4812866` (the fenced-URL link-gate fix on #590) were recovered from the local object store and re-applied on top of the rebased bases — both heads now carry their full content again. Root cause on my side: the brand cherry-picks were force-pushed against a stale tracking ref, which clobbered the remote-only response commits; that mistake is recorded in my own review notes so it doesn't recur.

**The two wording nits, fixed here:**
- `.gitignore` — the comment now says what writes the directory: "Session/tooling artifacts (preview state, generated knowledge graphs; the desktop tool writes .freebuff/ at session start)".
- `install-skill.sh` — the second destination is now labeled "the .agents skill surface (global)", distinct from Claude Code.

**docs/README.md** — added a small index ([api/](docs/api/), llm-format, mcp-plugin-consolidation, prerequisites) so the nav link lands on something that reads like a docs home rather than a bare listing.

All five branches pushed; CI is re-running on each head now.

PR #590, opened 2026-09-03
ci: fail on dead markdown links

Alot1z:
A link that stops resolving should not be able to merge. `ix-cli/scripts/check-links.mjs` probes every external URL in the tracked markdown (README, docs/, CONTRIBUTING, CLAUDE, skills/, homebrew, templates) and **fails on 404/410**; other 4xx/5xx and network flakes are warnings, so bot-blocked pages (e.g. `desktop.docker.com` answers 403 to anything without a browser) do not trip it. Zero dependencies — Node 22 global fetch, no `npm ci`, runs in seconds.

Verified (measured): 40 URLs checked, 0 dead on this branch; mutation tests prove both directions — a new dead link fails naming the URL, and a **stale allowlist entry** fails too.

**Self-cleaning allowlist**: the two currently-dead links are allowed with tracked replacements — `ix-infrastructure/ix-memory-layer` (fixed by #582) and `www.ix-infra.com/docs` (fixed by #589). Once either PR merges, its URL disappears from the markdown and the gate starts failing on the now-unused allow entry, forcing removal. No allow entry can rot silently.

Coordination: this PR edits the same `ci-success` aggregate block as #584 and #587 — whichever merges second needs a one-line rebase.


KageBinary:
Ran the checker on this branch. It is green, and the self-cleaning allowlist is a genuinely good idea. But the coverage is much smaller than the PR describes, and the number in the body is the giveaway.

## The gate checks 18 of the repo's 40 URLs

The body says "**40 URLs checked**, 0 dead on this branch". Running the committed script:

```
$ node ix-cli/scripts/check-links.mjs
checked 20 references — 0 broken, 0 warnings
```

That 20 is 18 external URLs plus 2 relative targets. Measured with the script's own `proseOf`/`extractUrls`:

```
external URLs actually probed (prose only): 18
external URLs if fences were included     : 40
relative targets checked                  :  2
```

40 is exactly what you get **without** the fenced-code-block exclusion — so the body's headline number appears to predate `proseOf` and was not re-measured after it was added. Shipped, the gate covers 45% of the repo's URLs.

## What falls in the gap is the part that matters most

Per-file, fence-stripping drops:

| file | URLs | probed | dropped |
|---|---|---|---|
| `docs/prerequisites.md` | 12 | **0** | 12 |
| `README.md` | 25 | 16 | 9 |
| `docs/api/README.md` | 5 | 1 | 4 |
| `CONTRIBUTING.md` | 2 | 1 | 1 |
| `docs/llm-format.md` | 1 | 0 | 1 |
| `homebrew/README.md` | 1 | 0 | 1 |

`docs/prerequisites.md` is checked **zero** times, and its 12 URLs are not illustrations — they are the repo's egress allowlist, the endpoints a firewalled user must permit for `ix` to install at all:

```
docs/prerequisites.md:120-131
# https://api.github.com            — resolve latest release version
# https://github.com                — download CLI tarball
# https://raw.githubusercontent.com — download docker-compose.yml
# https://ghcr.io                   — pull Ix Memory Layer image
# https://nodejs.org                — Node.js installer (macOS no-brew)
# https://deb.nodesource.com        — NodeSource apt (Linux)
...
# https://desktop.docker.com        — Docker Desktop .dmg (macOS no-brew)
```

Every one is load-bearing, all 12 sit inside a fence, and the gate skips all 12. Note the last line: `desktop.docker.com` is the body's own worked example for why "other 4xx is a warning" — so the classification logic was written for a URL this script never reaches.

The premise "illustrative URLs are not references" holds for `https://example.com` in a config sample. It does not hold here, where fenced blocks are copy-paste-and-run commands and documented endpoints.

I would drop `proseOf` entirely and rely on the allowlist for the handful of genuine illustrations — that is exactly what the allowlist's self-cleaning property makes safe.

## Merge-order hazard worth stating explicitly

The allowlist entries are keyed to #582 and #589. Whichever of those merges first, its URL disappears from the markdown and the stale-entry check fires — so **#582 and #589 both go red the moment this lands**, and each needs a follow-up commit deleting its allow entry. That is the design working as intended, but it means this PR should merge *after* #582 and #589, not before, and then land with an empty allowlist. Right now it is racing them.

## Smaller

- **Probes are sequential with a 10s timeout each.** 18 URLs today is fine; drop `proseOf` and it is 40, and a bad network turns a "runs in seconds" job into several minutes. `Promise.all` over a small concurrency pool would keep it honest.
- **`targetExists` does `[...allLower].some(...)` per relative target** — spreads the whole tracked-file set into an array on every call. Two targets today, so invisible; worth a `Set` prefix walk before it grows.
- **404-on-HEAD is not retried.** Only 405 falls back to GET, but some CDNs answer 404 to HEAD from a non-browser UA and 200 to GET. Retrying GET on 404 before failing would cost one request on the only path that can fail the build.
- **No trailing newline** at EOF (same as `check-doc-parity.mjs` in #584). `npm run lint` only covers `src`, so nothing catches it.
- The `# ── Aggregate gate ──` banner comment now sits above the `links:` job rather than above `ci-success`, which reads as if `links` is the aggregate.

The core idea is right and I would like to see it land — mainly I want `docs/prerequisites.md` inside the gate rather than outside it.

Alot1z:
Fixed in `4812866` (pushed to `ci/links-check`): `proseOf` is gone and fenced URLs are scanned like prose.

Measured on this branch, before and after:

| | before | after |
|---|---|---|
| references checked | 20 | **42** (40 external + 2 relative) |
| `docs/prerequisites.md` URLs probed | 0 | **12/12** |
| broken | 0 | 0 (exit 0) |
| warnings | 0 | 1 — `desktop.docker.com` HTTP 403, the documented bot-blocking case |

Two extra verifications beyond the count:

- A dead URL inserted inside a fence (`github.com/ix-infrastructure/Ix/nonexistent-path-590`) now fails the gate: `ERROR … exit 1`. Before this change that fence was never scanned.
- The `desktop.docker.com` 403 is the PR body's own worked example for "other 4xx is a warning" — the classification logic now actually reaches the URL it was written for.

The allowlist stays exactly as designed: it absorbs the two dead URLs keyed to #582/#589, and its stale-entry check is what enforces the merge order you flagged — the moment #582 or #589 removes its URL, this gate fails on the unused entry until the allow entry is dropped, so merging those two first is still the right order.

The "40 URLs checked" claim in the body is accurate again (42 with the relative targets).

KageBinary:
## Same as #584 — the response commit is not on the branch

Head is `b675596`, force-pushed at `2026-09-03T23:26:47Z`. `4812866` is not in
it. The branch is `72524c2` → `3ddcb8e` → `b675596`, and `3ddcb8e`
(`2026-09-03 04:35`) is the commit that *added* `proseOf`, not the one that
removed it.

On the current head:

```
$ git show HEAD:ix-cli/scripts/check-links.mjs | grep -n proseOf
57:function proseOf(text) {
141:  const prose = proseOf(text);
```

So fenced code blocks are still stripped before extraction, and the URLs inside
them are still unscanned — the before/after table in your comment describes a
tree that is not what this PR contains.

Also still there: no newline at EOF (`check-links.mjs` ends on `;`), which
`npm run lint` won't catch since it only covers `src`.

Holding the rest of the re-review until `4812866` is back on the branch.

Cross-cutting note about the rebase that caused this is on #589.

Alot1z:
Fixed — the response commit is back on `ci/links-check` as head `4812866` (fenced-URL link gate). The de-branding commit is dropped from this branch per your cross-cutting note; it now lives only on #589.

KageBinary:
`4812866` is back on the branch and it does what the commit says. Re-verified on that head by running the gate, not by reading it.

## `proseOf` is gone, and the coverage more than doubled

```
$ grep -n proseOf ix-cli/scripts/check-links.mjs
(no matches)

$ node ix-cli/scripts/check-links.mjs
warn   https://desktop.docker.com/ — HTTP 403 (in docs/prerequisites.md)
checked 42 references — 0 broken, 1 warnings
exit 0
```

42 references, up from 18. `docs/prerequisites.md` is inside the gate now, which was my main ask — its documented endpoints and install commands live in fences and were the exact content the prose filter was stripping.

## Mutation-tested, four ways

| mutation | result |
|---|---|
| dead URL in prose | `ERROR https://github.com/ix-infrastructure/definitely-not-a-real-repo-9f2a1 (in README.md)` — exit 1 |
| dead relative target | `ERROR ./docs/nope-not-here.md (relative, in README.md)` — exit 1 |
| dead URL **inside a fenced block** | caught — this is the case `4812866` adds, and it is genuinely closed |
| allowlisted URL removed from the markdown | `ERROR stale allowlist entry https://www.ix-infra.com/docs no longer appears in any scanned file — remove it` — exit 1 |

Clean run is 8.9s wall clock, so the sequential probing is fine at this size in practice. The 10s-per-URL worst case still stands as a theoretical ceiling; not worth changing now.

Restricting hard errors to 404/410 and warning on everything else is the right call — `desktop.docker.com` answering 403 to a non-browser UA is exactly the flake that would otherwise make this job untrustworthy, and a link that is actually gone answers 404.

## The merge-order hazard is now demonstrated, not just argued

That last mutation is the real one: I simulated #589 having merged by replacing its URL in `README.md`, and the gate went red immediately. So the moment **either #582 or #589 lands, this branch fails** until its allowlist entry is deleted — and symmetrically, if this lands first, both of those PRs go red on their next CI run.

That is the self-cleaning design working exactly as intended, but it means the order matters: **merge #582 and #589 first, then land this with an empty `ALLOW` map.** Worth putting in the PR description so whoever merges does not discover it from a red check.

## One thing still open

`check-links.mjs` still ends on `;` with no trailing newline. `npm run lint` only covers `src`, so nothing will ever catch it — same as `check-doc-parity.mjs`, which #584 has since fixed.

Otherwise this is good to go, subject to the merge order above.

Alot1z:
Thank you — the fenced-block direction closes the real gap, and the demonstrated merge-order hazard is now on record rather than assumed. Per that constraint: **#582 and #589 are both still open as of right now**, so this branch remains green and there is no action needed from me until one of them lands. When that happens I'll rebase this branch with the allowlist entry dropped in the same push (gated sequence, lease-pinned) so the self-cleaning design lands clean.

PR #592, opened 2026-09-03
fix(test): stop the ignore assertion blaming .gitignore when git will not answer

### LINKED ISSUE #585: watch-dedup: the new git check-ignore assertion fails in a worktree whose node_modules is a symlink
## Summary

The regression test added in #569 — `canonical watch refresh > builds the child CLI somewhere git cannot see, so an interrupted run leaves nothing behind` — cannot run in a checkout whose `ix-cli/node_modules` is a symlink. It fails there with a message that blames `.gitignore`, which is the one thing that is not wrong.

## Cause

`git check-ignore` refuses any path that lies beyond a symbolic link:

```console
$ ls -ld ix-cli/node_modules
lrwxrwxrwx  ix-cli/node_modules -> /home/…/oss/Ix/ix-cli/node_modules

```

The test's `explain()` helper reads only `stdout`, so it gets `""`, `sourceOf("")` returns `""`, and the assertion fails:

```
AssertionError: git does not ignore /…/ix-cli/node_modules/.cache via this repo's .gitignore:
  expected '' to be '.gitignore'
  ❯ watch-dedup.test.ts:178
```

## Why it matters more than it looks

That layout — `node_modules` symlinked to a sibling clone's install — is how a git worktree borrows dependencies, and it is precisely the arrangement that produced the committed `ix-cli/node_modules` symlink of #545 that #569 was untangling. So the test is unavailable in the environment its own subject matter comes from.

It is **not** reachable in CI: `npm ci` creates a real directory there, and the test passes. It also has nothing to do with the ignore rules — reproduced identically on unmodified `main` at `122add8`, and it survives #580.

## Suggested fix

`explain()` could fall back when git refuses to answer, e.g. resolve `packageRoot` through `fs.realpathSync` before building `cacheRoot`, or detect the `fatal: … beyond a symbolic link` on stderr and take the same `else` branch the test already has for "not in this repo" (assert structurally that the build root is under `node_modules`). Either keeps the assertion meaningful where git can answer and stops it failing where git will not.

## Environment

- Ix `main` @ `122add8` and `59933d5`
- git 2.x, Linux

KageBinary:
Closes #585.

`git check-ignore` refuses any path that lies beyond a symbolic link — it writes `fatal: pathspec '…' is beyond a symbolic link` to **stderr** and prints nothing on stdout. The `explain()` helper read only stdout, so the refusal arrived as `""`, `sourceOf("")` returned `""`, and the assertion failed with:

```
AssertionError: git does not ignore …/ix-cli/node_modules/.cache via this repo's .gitignore:
  expected '' to be '.gitignore'
  ❯ watch-dedup.test.ts:178
```

That message blames this repo's `.gitignore`, which is the one thing that is not wrong. The rules are correct; git simply declined to answer.

## Why it is worth fixing rather than tolerating

Symlinking `ix-cli/node_modules` at a sibling clone's install is how a git worktree borrows dependencies — and it is the very practice that produced the committed `ix-cli/node_modules` symlink of #545 that this test exists to pin. So the regression test was unavailable in exactly the environment its own subject matter comes from. (It hit me twice while working #580.)

## The change

`explain()` now reports whether git could answer at all:

```ts
if (probe.status !== 0 && /beyond a symbolic link/.test(probe.stderr ?? "")) {
  return { source: "", answerable: false };
}
```

A refusal is the same epistemic position as being outside a work tree, so it takes the same route the test already had for that case:

- **`cacheRoot`** falls back to the structural half of the claim — the build root is under `node_modules`.
- **A per-rule probe** git cannot answer for is skipped rather than asserted against `""`.

Where git *can* answer — CI, and any checkout with a real `node_modules` — every assertion runs exactly as before, including all four directory-only rule probes from #580. Nothing is weakened on the path that matters.

## Verified in both layouts, same commit

Built the two layouts side by side in one worktree (`node_modules` as a real directory of package symlinks, then as a plain symlink) so the only variable is whether git will answer:

| layout | `main` | this branch |
|---|---|---|
| `node_modules` a **real directory** (git answers) | 16/16 pass | **16/16 pass** |
| `node_modules` a **symlink** (git refuses) | **1 failed**, 15 passed | **16/16 pass** |

The failure on `main` is the #585 assertion with a byte-identical message. Git's behaviour in that layout, for the record:

```console
$ git check-ignore -v --no-index node_modules/.cache/x
fatal: pathspec 'node_modules/.cache/x' is beyond a symbolic link
```

versus the real-directory layout, where the assertion still has something to assert on:

```console
$ git check-ignore -v --no-index node_modules/.cache/x
.gitignore:28:node_modules	node_modules/.cache/x
```

## Gates

- Full suite: **89 files, 1510 passed, 2 skipped**
- `lint`, `typecheck`, `knip` — all match `main`'s baseline exactly (the three "redundant entry pattern" knip warnings are pre-existing on `main`)
- Test-only change; no production code touched

### DIFF
--- a/ix-cli/src/cli/__tests__/watch-dedup.test.ts
+++ b/ix-cli/src/cli/__tests__/watch-dedup.test.ts
@@ -156,17 +156,6 @@ describe("canonical watch refresh", () => {
       fs.realpathSync(topLevel.stdout.trim()) === fs.realpathSync(repoRoot);
 
     if (inThisRepo) {
-      // -v, not --quiet: the exit code alone also counts a rule from the
-      // user's global core.excludesFile or .git/info/exclude. A contributor
-      // whose personal ignore file says `node_modules` (no slash -- a common
-      // spelling) could then restore the directory-only rule here, make
-      // ix-cli/node_modules committable again, and still see this pass. The
-      // source has to be this repo's own .gitignore.
-      const explain = (target: string): string =>
-        spawnSync("git", ["check-ignore", "-v", "--no-index", target], {
-          cwd: packageRoot,
-          encoding: "utf8",
-        }).stdout.trim();
       // git -v prints `<source>:<line>:<pattern>\t<pathname>`. Splitting the
       // source off on the first ":" yields "C" for a Windows path -- which
       // happens to differ from ".gitignore" and so still fails the assertion,
@@ -174,8 +163,44 @@ describe("canonical watch refresh", () => {
       const sourceOf = (out: string): string =>
         /^(.+):\d+:/.exec(out.split("\t")[0] ?? "")?.[1] ?? "";
 
-      expect(sourceOf(explain(cacheRoot)), `git does not ignore ${cacheRoot} via this repo's .gitignore`)
-        .toBe(".gitignore");
+      // -v, not --quiet: the exit code alone also counts a rule from the
+      // user's global core.excludesFile or .git/info/exclude. A contributor
+      // whose personal ignore file says `node_modules` (no slash -- a common
+      // spelling) could then restore the directory-only rule here, make
+      // ix-cli/node_modules committable again, and still see this pass. The
+      // source has to be this repo's own .gitignore.
+      //
+      // `git check-ignore` refuses outright for a path that lies beyond a
+      // symbolic link -- `fatal: pathspec '...' is beyond a symbolic link`, on
+      // stderr, with nothing on stdout. Reading stdout alone turned that
+      // refusal into `sourceOf("") === ""` and failed the assertion blaming
+      // this repo's .gitignore, which is the one thing that is not wrong.
+      //
+      // It is not a corner case: symlinking `ix-cli/node_modules` at a sibling
+      // clone's install is how a worktree borrows dependencies, and it is the
+      // very practice that produced the committed symlink of #545 this test
+      // exists to pin -- so the test was unavailable in the layout its own
+      // subject matter comes from. Git has no answer to give there, which is
+      // the same position as being outside a work tree, so say so and let the
+      // caller fall back to the structural half of the claim. (#585)
+      const explain = (target: string): { source: string; answerable: boolean } => {
+        const probe = spawnSync("git", ["check-ignore", "-v", "--no-index", target], {
+          cwd: packageRoot,
+          encoding: "utf8",
+        });
+        if (probe.status !== 0 && /beyond a symbolic link/.test(probe.stderr ?? "")) {
+          return { source: "", answerable: false };
+        }
+        return { source: sourceOf(probe.stdout.trim()), answerable: true };
+      };
+
+      const cache = explain(cacheRoot);
+      if (cache.answerable) {
+        expect(cache.source, `git does not ignore ${cacheRoot} via this repo's .gitignore`)
+          .toBe(".gitignore");
+      } else {
+        expect(path.relative(packageRoot, cacheRoot).split(path.sep)[0]).toBe("node_modules");
+      }
 
       // And pin the rule that ignores it, not just today's outcome. `cacheRoot`
       // is a real directory, so the old `node_modules/` rule matched it too and
@@ -201,8 +226,14 @@ describe("canonical watch refresh", () => {
         // that happens to be a real build directory proves nothing. Ask about a
         // sibling path that cannot exist instead.
         const target = fs.existsSync(probe) ? path.join(probe, "__ix_ignore_probe__") : probe;
+        const answer = explain(target);
+        // A symlinked artifact directory is the one layout git will not answer
+        // for. Nothing is knowable about the rule from here, and asserting on
+        // the empty string would blame .gitignore for git's refusal -- the
+        // defect this test itself hit in #585.
+        if (!answer.answerable) continue;
         expect(
-          sourceOf(explain(target)),
+          answer.source,
           `the '${rule}' ignore rule is directory-only again; a symlink or file with that name is committable (Ix#545)`,
         ).toBe(".gitignore");
       }
--- a/ix-cli/src/cli/__tests__/watch-dedup.test.ts
+++ b/ix-cli/src/cli/__tests__/watch-dedup.test.ts
@@ -199,7 +199,19 @@ describe("canonical watch refresh", () => {
         expect(cache.source, `git does not ignore ${cacheRoot} via this repo's .gitignore`)
           .toBe(".gitignore");
       } else {
-        expect(path.relative(packageRoot, cacheRoot).split(path.sep)[0]).toBe("node_modules");
+        // git refuses `node_modules/.cache` only because `node_modules` is a
+        // *leading* symlink component -- it answers for the symlink itself, and
+        // `cacheRoot` is under it by construction. Asserting that
+        // `relative(packageRoot, cacheRoot)` starts with "node_modules" would be
+        // a tautology (childBuildCacheRoot joins that literal), so it would pass
+        // with every rule deleted from .gitignore. Ask the question that is
+        // actually answerable instead.
+        const nodeModules = explain(path.join(packageRoot, "node_modules"));
+        expect(nodeModules.answerable, "git could not answer for node_modules itself").toBe(true);
+        expect(
+          nodeModules.source,
+          `git does not ignore ${packageRoot}/node_modules via this repo's .gitignore`,
+        ).toBe(".gitignore");
       }
 
       // And pin the rule that ignores it, not just today's outcome. `cacheRoot`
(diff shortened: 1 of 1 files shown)


ix-cli/src/cli/__tests__/watch-dedup.test.ts
```diff
@@ -201,8 +226,14 @@ describe("canonical watch refresh", () => {
         // that happens to be a real build directory proves nothing. Ask about a
         // sibling path that cannot exist instead.
         const target = fs.existsSync(probe) ? path.join(probe, "__ix_ignore_probe__") : probe;
+        const answer = explain(target);
+        // A symlinked artifact directory is the one layout git will not answer
+        // for. Nothing is knowable about the rule from here, and asserting on
+        // the empty string would blame .gitignore for git's refusal -- the
+        // defect this test itself hit in #585.
+        if (!answer.answerable) continue;
```
josephismikhail: **The skip fires exactly where the probe is most valuable -- and the refusal is avoidable here.**

`fs.existsSync` follows symlinks. In the very layout this PR targets (a worktree borrowing artifacts from a sibling clone), `ix-cli/dist` is a symlink to a real directory, so `existsSync(probe)` is true, `target` becomes `ix-cli/dist/__ix_ignore_probe__`, its *leading* path crosses the symlink, git dies, and line 234 `continue`s. The `dist`, `coverage` and `core-ingestion/dist` assertions from #580 all vanish silently in that layout.

But git refuses only on a **leading** symlink component -- `has_symlink_leading_path()` in `pathspec.c` stops at the last slash. Probing `ix-cli/dist` *itself* is answerable, and it is a strictly better probe than the synthesized sibling: a symlink is not a directory, so a directory-only `dist/` rule does **not** match it while `dist` does. That is precisely the #545/#580 property being pinned -- with a real committable symlink actually present on disk. So the fallback discards the strongest available signal in the one layout where the defect is materially present.

Only a *real* directory needs the sibling-path trick; descend only then:

```suggestion
        // An existing real directory matches a trailing-slash rule too, so a
        // probe that happens to be a real build directory proves nothing. Ask
        // about a sibling path that cannot exist instead. A symlink is NOT a
        // directory to git, so the symlink path itself is already the probe we
        // want -- and descending into it is what makes git refuse to answer.
        const lst = fs.existsSync(probe) ? fs.lstatSync(probe) : null;
        const target = lst?.isDirectory() ? path.join(probe, "__ix_ignore_probe__") : probe;
        const answer = explain(target);
        // A symlinked artifact directory is the one layout git will not answer
        // for. Nothing is knowable about the rule from here, and asserting on
        // the empty string would blame .gitignore for git's refusal -- the
        // defect this test itself hit in #585.
        if (!answer.answerable) continue;
```

ix-cli/src/cli/__tests__/watch-dedup.test.ts
```diff
@@ -156,26 +156,51 @@ describe("canonical watch refresh", () => {
...
+        }
+        return { source: sourceOf(probe.stdout.trim()), answerable: true };
+      };
+
+      const cache = explain(cacheRoot);
+      if (cache.answerable) {
+        expect(cache.source, `git does not ignore ${cacheRoot} via this repo's .gitignore`)
+          .toBe(".gitignore");
+      } else {
+        expect(path.relative(packageRoot, cacheRoot).split(path.sep)[0]).toBe("node_modules");
```
josephismikhail: **This fallback is a tautology, and a checkable claim was available instead.**

`cacheRoot` is `path.join(packageRoot, "node_modules", ".cache")` by construction (`childBuildCacheRoot`), so `path.relative(packageRoot, cacheRoot).split(path.sep)[0]` is the literal string `"node_modules"` no matter what `.gitignore` says. In the symlinked-`node_modules` layout the test therefore makes **no** claim about ignoring at all -- delete every rule from `.gitignore` and this branch still passes.

That is defensible for the outside-a-work-tree branch (line 241), where there is genuinely no git to ask. It is not forced here: git refuses `node_modules/.cache` only because `node_modules` is a *leading* component. `<packageRoot>/node_modules` itself has no symlink ancestor, so `explain()` answers for it normally. The full claim -- "an ancestor of `cacheRoot` is ignored via this repo's `.gitignore`, and `cacheRoot` is under it" -- is fully verifiable in exactly the layout this branch handles:

```ts
} else {
  // git will not answer past the node_modules symlink, but it answers for
  // the symlink itself -- and `cacheRoot` is under it by construction.
  const nm = explain(path.join(packageRoot, "node_modules"));
  if (nm.answerable) {
    expect(nm.source, `git does not ignore ${packageRoot}/node_modules via this repo's .gitignore`)
      .toBe(".gitignore");
  }
  expect(path.relative(packageRoot, cacheRoot).split(path.sep)[0]).toBe("node_modules");
}
```

ix-cli/src/cli/__tests__/watch-dedup.test.ts
```diff
@@ -156,26 +156,51 @@ describe("canonical watch refresh", () => {
...
-      const sourceOf = (out: string): string =>
-        /^(.+):\d+:/.exec(out.split("\t")[0] ?? "")?.[1] ?? "";
-
-      expect(sourceOf(explain(cacheRoot)), `git does not ignore ${cacheRoot} via this repo's .gitignore`)
-        .toBe(".gitignore");
+        });
+        if (probe.status !== 0 && /beyond a symbolic link/.test(probe.stderr ?? "")) {
+          return { source: "", answerable: false };
+        }
+        return { source: sourceOf(probe.stdout.trim()), answerable: true };
```
josephismikhail, line 194: **The detector matches an English string that git translates.**

`"pathspec '%s' is beyond a symbolic link"` is a `die(_(...))` in `pathspec.c` -- it goes through gettext. A contributor whose shell sets `LANG`/`LC_ALL` to a locale git ships a translation for (de, fr, zh_CN, ... all in `po/`) gets the localized text on stderr, `/beyond a symbolic link/` misses, `answerable` comes back `true`, `source` is `sourceOf("")` = `""`, and the assertion fails with `expected '' to be '.gitignore'` -- byte-identical to the #585 failure this PR exists to remove. The bug is simply relocated to a subset of machines rather than fixed, and it is invisible in CI, which runs in the C locale.

`spawnSync` inherits the ambient environment by default, so pin the locale for the probe:

```suggestion
      const explain = (target: string): { source: string; answerable: boolean } => {
        const probe = spawnSync("git", ["check-ignore", "-v", "--no-index", target], {
          cwd: packageRoot,
          encoding: "utf8",
          // The refusal below is matched by its English text, and git's die()
          // messages go through gettext. Pin the locale so a contributor with
          // a translated git does not get #585 back on a localized message.
          env: { ...process.env, LC_ALL: "C", LANGUAGE: "" },
        });
        if (probe.status !== 0 && /beyond a symbolic link/.test(probe.stderr ?? "")) {
          return { source: "", answerable: false };
        }
        return { source: sourceOf(probe.stdout.trim()), answerable: true };
      };
```

Alot1z (commented):
Reviewed the diff and the two-layout claim. Strong fix — one non-blocking robustness point.

**What's right**

The diagnosis is the best part: "refusal arrives as `""`, `sourceOf("")` returns `""`, assertion blames `.gitignore`" is exactly the failure #585 shows, and the fix does not weaken the answering path — when git answers, every assertion (including all four directory-only rule probes) runs exactly as before. The `lstatSync` change fixes a real silent hole: `fs.existsSync` follows the symlink, so the old sibling-path trick took the wrong branch for a symlinked `ix-cli/dist`, and git's refusal then vanished into the `continue` — in the one layout the test exists for. And the fallback deliberately avoids the tautology (asserting `cacheRoot` starts with `node_modules` would pass with every rule deleted) by asserting instead that `node_modules` itself is ignored by this repo's `.gitignore`.

**One thing to consider (non-blocking): the refusal detection is locale-dependent**

```ts
if (probe.status !== 0 && /beyond a symbolic link/.test(probe.stderr ?? "")) {
```

Git's refusal text is localized; the exit code is not. Measured on this machine: ignored path → exit 0, not-ignored path → exit 1, fatal (probe outside a repo) → exit 128. A contributor running a localized git gets the refusal in another language, the regex misses, `answerable` returns true with `source: ""`, and the assertion fails with the misleading `.gitignore`-blame message again — the exact symptom this PR fixes, resurfacing in the very worktree layout the test exists to pin. `probe.status === 128` identifies the same refusal with no language dependency, and cannot be confused with the real "rule missing" failure (that is exit 1, which must keep failing). On CI and on English-locale git the two checks are identical, so this changes none of the measured 16/16 results; it only keeps the fix honest for contributors whose git answers in another language.

**Pre-existing, not introduced here:** the `sourceOf` `:\d+:` anchor would misparse a directory literally named like `foo:123:bar` — same tradeoff as before the move.

PR #593, opened 2026-09-03
feat(ingest): parse CUDA (.cu/.cuh) and Python stubs (.pyi)

KageBinary:
## Why

ML repositories keep their hottest code in CUDA and declare their compiled
extensions in `.pyi` stubs. Ix ingested neither, so a PyTorch-shaped repo lost
both its kernels and the only Python-visible signature of any native op it
calls — the graph stopped at the import.

## `.pyi` — one line

The Python grammar already parses stubs (`def f(x: int) -> str: ...`). Just a
map entry.

## `.cu` / `.cuh` — not one line

CUDA maps onto the C++ grammar, but the kernel-launch syntax
`k<<<grid, block>>>(args)` is not valid C++. Measured against the bundled
`tree-sitter-cpp`:

```
HAS ERROR: true
--- any call_expression found? ---   (none)
```

The grammar raises an `ERROR` and recovers by dropping the call node entirely.
Kernel *definitions* survive, so a naive extension mapping looks like it works
while silently losing the host → kernel edge — the one edge worth having in a
`.cu` file.

The fix blanks the launch config in place before parsing, following the
existing Java-annotation and Rust-cfg-macro precedent already in `parseFile`.
Both character positions *and* newlines are preserved, so entity line numbers
stay accurate across a multi-line launch config (pinned by a test — a
space-only fill would have dragged every subsequent entity upwards).

Scoping, so this cannot touch ordinary C++:

- `(?=\s*\()` — a launch config is always followed by its argument list, which
  leaves `vector<vector<vector<int>>>`, `a << b` and `"<<<HEAD>>>"` in a string
  literal alone.
- excluding `;` — stops a stray `<<<` in a comment running away to a later `>>>`.
- gated on the file extension rather than the language, so `.cpp`/`.hpp`
  parsing is byte-identical.

## Extension parity test

`supported-extensions.ts` states in prose that it "MUST stay in sync" with
`EXT_MAP`, but nothing enforced it. The failure mode is silence: a file type the
walker never yields is simply absent from the graph, with no error. Now
enforced in both directions. The two lists were already in sync — this adds the
guard, it does not fix drift.

## Verification

- `core-ingestion`: 370 passed (5 new CUDA, 4 new parity, 1 new `.pyi`)
- `ix-cli`: 1510 passed + parser smoke
- lint 0 errors, typecheck clean, knip no new findings

Covered: simple launches, expression configs, 4-arg launches with stream +
shared memory, multi-line configs, templated kernels, `.cuh` headers, and three
false-positive traps asserted byte-identical to `.cpp` handling.

### DIFF
--- a/core-ingestion/src/__tests__/extension-parity.test.ts
+++ b/core-ingestion/src/__tests__/extension-parity.test.ts
@@ -1,0 +1,48 @@
+import { readFileSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+
+import { describe, expect, it } from 'vitest';
+
+import { languageFromPath } from '../languages.js';
+
+// supported-extensions.ts documents that it "MUST stay in sync" with EXT_MAP,
+// but nothing enforced it, so every new parser was one forgotten line away from
+// ingesting nothing: a file type the walker never yields is silently absent from
+// the graph rather than failing loudly.
+//
+// Both files are read as text so this stays independent of the ix-cli build.
+const read = (relative: string) =>
+  readFileSync(fileURLToPath(new URL(relative, import.meta.url)), 'utf8');
+
+const parsedExtensions = (() => {
+  const source = read('../languages.ts');
+  const map = source.slice(source.indexOf('const EXT_MAP'));
+  return new Set([...map.matchAll(/'(\.[a-z0-9]+)'\s*:/gi)].map((match) => match[1]));
+})();
+
+const walkedExtensions = new Set(
+  [...read('../../../ix-cli/src/cli/supported-extensions.ts')
+    .matchAll(/"(\.[a-z0-9]+)"/gi)].map((match) => match[1]),
+);
+
+describe('extension parity between EXT_MAP and SUPPORTED_EXTENSIONS', () => {
+  it('reads a non-empty set from both files', () => {
+    // Guards against a regex that silently matches nothing after a refactor,
+    // which would make every assertion below vacuously true.
+    expect(parsedExtensions.size).toBeGreaterThan(50);
+    expect(walkedExtensions.size).toBeGreaterThan(50);
+  });
+
+  it('walks every extension a parser handles', () => {
+    expect([...parsedExtensions].filter((ext) => !walkedExtensions.has(ext))).toEqual([]);
+  });
+
+  it('has a parser for every extension it walks', () => {
+    expect([...walkedExtensions].filter((ext) => !parsedExtensions.has(ext))).toEqual([]);
+  });
+
+  it('resolves each walked extension to a language', () => {
+    const unresolved = [...walkedExtensions].filter((ext) => !languageFromPath(`file${ext}`));
+    expect(unresolved).toEqual([]);
+  });
+});
--- a/core-ingestion/src/__tests__/queries.cuda.test.ts
+++ b/core-ingestion/src/__tests__/queries.cuda.test.ts
@@ -87,4 +87,25 @@ void f(std::vector<std::vector<std::vector<int>>>& v, int a, int b) {
     expect(callTargets(cu)).toEqual(callTargets(cpp));
     expect(callTargets(cu)).toEqual(expect.arrayContaining(['consume']));
   });
+
+  it('stays linear on adversarial input', () => {
+    // Ingest input is an arbitrary repository, so the blanking pass is
+    // attacker-controlled. The obvious regex (`/<<<[^;]*?>>>(?=\\s*\\()/g`)
+    // is quadratic here: 16k `<` took 67ms and grew 4x per doubling.
+    const timeFor = (n: number) => {
+      const evil = `void f() { ${'<'.repeat(n)}>>>x }`;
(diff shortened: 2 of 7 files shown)


core-ingestion/src/languages.ts
```diff
@@ -42,13 +42,18 @@ const EXT_MAP: Record<string, SupportedLanguages> = {
   '.mjs':  SupportedLanguages.JavaScript,
   '.cjs':  SupportedLanguages.JavaScript,
   '.py':   SupportedLanguages.Python,
+  '.pyi':  SupportedLanguages.Python,
```
josephismikhail, line 45: @C:/Users/josep/AppData/Local/Temp/rev593/c1.md

core-ingestion/src/languages.ts
```diff
@@ -42,13 +42,18 @@ const EXT_MAP: Record<string, SupportedLanguages> = {
...
   '.java': SupportedLanguages.Java,
   '.c':    SupportedLanguages.C,
   '.h':    SupportedLanguages.C,
   '.cpp':  SupportedLanguages.CPlusPlus,
   '.cc':   SupportedLanguages.CPlusPlus,
   '.cxx':  SupportedLanguages.CPlusPlus,
   '.hpp':  SupportedLanguages.CPlusPlus,
+  // CUDA is C++ plus GPU qualifiers; kernel-launch syntax is neutralised at
+  // parse time (see blankCudaLaunchConfigs in index.ts).
+  '.cu':   SupportedLanguages.CPlusPlus,
```
josephismikhail, line 55: @C:/Users/josep/AppData/Local/Temp/rev593/c2.md

core-ingestion/src/index.ts
```diff
@@ -2102,6 +2175,13 @@ export function parseFile(filePath: string, source: string): FileParseResult | n
     if (language === SupportedLanguages.Rust) {
       parseSource = unwrapRustCfgMacros(parseSource);
     }
+
+    // CUDA: neutralise kernel-launch configs so the host -> kernel call survives.
+    // Gated on the extension rather than the language so .cpp/.hpp parsing stays
+    // byte-identical.
+    if (filePath.endsWith('.cu') || filePath.endsWith('.cuh')) {
```
josephismikhail, line 2182: @C:/Users/josep/AppData/Local/Temp/rev593/c3.md

core-ingestion/src/index.ts
```diff
@@ -1989,6 +1989,79 @@ function unwrapRustCfgMacros(source: string): string {
...
+ *
+ * A launch config is always followed immediately by the argument list, so
+ * requiring `(` after the closing `>>>` is what keeps this off ordinary C++ —
+ * nested templates (`vector<vector<vector<int>>>`), shift operators and
+ * `"<<<HEAD>>>"` in a string literal are all left alone. Stopping the search
+ * at `;` keeps a stray `<<<` in a comment from reaching a later `>>>`.
+ */
+const _isSpace = (ch: string) => ch === ' ' || ch === '\n' || ch === '\t' || ch === '\r';
+
+function blankCudaLaunchConfigs(source: string): string {
```
josephismikhail, line 2018: @C:/Users/josep/AppData/Local/Temp/rev593/c4.md

core-ingestion/src/languages.ts
```diff
@@ -42,13 +42,18 @@ const EXT_MAP: Record<string, SupportedLanguages> = {
...
+  '.cuh':  SupportedLanguages.CPlusPlus,
```
josephismikhail, line 56: @C:/Users/josep/AppData/Local/Temp/rev593/c5.md

core-ingestion/src/__tests__/queries.cuda.test.ts
```diff
@@ -0,0 +1,111 @@
...
+      return performance.now() - started;
+    };
+
+    timeFor(1_000); // warm the parser so the comparison is not first-call cost
+
+    const small = timeFor(8_000);
+    const large = timeFor(32_000);
+
+    // 4x the input. Quadratic would be ~16x; linear scanning stays well under.
+    expect(large).toBeLessThan(Math.max(small, 1) * 8);
```
josephismikhail, line 108: @C:/Users/josep/AppData/Local/Temp/rev593/c6.md

core-ingestion/src/__tests__/extension-parity.test.ts
```diff
@@ -0,0 +1,48 @@
...
+// ingesting nothing: a file type the walker never yields is silently absent from
+// the graph rather than failing loudly.
+//
+// Both files are read as text so this stays independent of the ix-cli build.
+const read = (relative: string) =>
+  readFileSync(fileURLToPath(new URL(relative, import.meta.url)), 'utf8');
+
+const parsedExtensions = (() => {
+  const source = read('../languages.ts');
+  const map = source.slice(source.indexOf('const EXT_MAP'));
```
josephismikhail, line 19: @C:/Users/josep/AppData/Local/Temp/rev593/c7.md

PR #597, opened 2026-09-04
test(ingest): drive ingestFiles end to end, closing the seam #560 and #568 bled from

josephismikhail:
`ingestFiles` had no test coverage at all, and that is where **every** mechanical defect in #560 and #568 lived. This adds an integration harness that drives it against a real fake backend.

## Why

Both of those PRs extracted decision tables — `perFileAction`, `drainInPasses`, `createDrainGate` — and each extraction immediately caught a bug that had already shipped. But it *moved* the untested boundary rather than closing it: the next defect always turned up in the wiring between the tables. A counter fed from the wrong place. A list the deadline diverted to. A probe that read proof as failure.

Across roughly thirty review rounds, several findings were the previous round's fix. Reviewers named the cause each time: `ingestFiles` is imported by no test, so ~350 lines of new wiring were verified by hand-run stubs that live nowhere in the repo.

## What this does

Drives the real function against a real HTTP server on a loopback port. `ingestFiles` builds its own `IxClient` from `getEndpoint()`, which reads `IX_ENDPOINT` — so nothing is mocked, and the backend's request log is the same measurement both PR bodies were built from.

Nine cases, chosen as the shapes the rounds kept breaking:

| | asserts |
|---|---|
| healthy repo | one bulk, zero per-file fan-out |
| backend refuses everything | requests bounded well under `main`'s N+1, every patch accounted for |
| poison patches mid-repo | errors equal the poison count **exactly** |
| poison at the leading edge | same, with no successes before the trip |
| poison at **both ends** | same — the shape that defeats every single-pass rule |
| bulk 409, all landed | confirmed patches counted, not reported as errors |
| deadline fires | the clock's losses reported as the clock's, not the backend's |
| stitch guard | a second attempt is refused while the marker stands |
| `IX_STITCH_COOLDOWN_MS=0` | releases a cooldown already on disk |

## Validated by restoring the bugs

Not "the tests pass" — the tests were checked against the real defects:

- removing the cutoff entirely (the pre-#560 fan-out) → **red**
- the drain stopping on its first empty pass → **red** (this is why the both-ends fixture exists; the middle and leading fixtures do not reproduce it)
- the stitch marker cleared by something that is not proof → **red**

The header also records what these do **not** catch: the finer rules inside `drainInPasses` — two empty passes versus three, and the direction flip — which need a precisely shaped held set and are mutation-validated in `commit-breaker.test.ts` instead. Coarse wiring here, decision tables there. Saying so beats implying blanket coverage.

## Two things the harness found immediately

**Why the seam had no coverage.** `loadIngestionModules` reaches `core-ingestion/dist` through a `new Function("return import(specifier)")` indirection, which keeps a specifier pointing four levels outside the package away from anything that would try to resolve it. Inside a `vm` context with no `importModuleDynamically` hook — vitest's vite-node — that throws `A dynamic import callback was not specified`, from the indirection itself rather than the module. The indirection stays first and unchanged; a direct `import()` is the fallback, taken only for that one error so a genuine module-not-found still propagates.

**A residue, pinned rather than accepted.** On a healthy backend, a bulk answering `409` with every patch id already landed still re-sends all thirty one at a time. The #560 bound only engages when those replays *fail*. Skipping them outright needs a revision for `onCommitted` that the 409 does not carry, so the test pins the exact number with a comment — if someone fixes it, the test says so rather than silently drifting.

## Verification

| check | result |
|---|---|
| `ingest-files.test.ts` | 9/9, ~10s |
| `npx vitest run` (full) | **1633 passed, 3 failed, 21 skipped** |

The 3 failures are the pre-existing Windows-only `EPERM: operation not permitted, symlink` cases in `read-containment.test.ts`; they fail on `main` on this machine too.

### DIFF
--- a/ix-cli/src/cli/__tests__/ingest-files.test.ts
+++ b/ix-cli/src/cli/__tests__/ingest-files.test.ts
@@ -1,0 +1,376 @@
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { createServer, type Server, type ServerResponse } from "node:http";
+import type { AddressInfo } from "node:net";
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import { ingestFiles } from "../commands/ingest.js";
+
+/**
+ * Integration tests that drive `ingestFiles` end to end against a fake backend.
+ *
+ * They exist because that function is where every mechanical defect in Ix#560
+ * and Ix#568 lived, and none of them were reachable from a unit test. The
+ * decision tables were extracted (`perFileAction`, `drainInPasses`,
+ * `createDrainGate`) and each extraction immediately caught a bug -- but the
+ * WIRING between them stayed untested, and that is where the next one appeared
+ * every time: a counter fed from the wrong place, a list the deadline diverted
+ * to, a probe that treated proof as a failure.
+ *
+ * Nothing is mocked. `ingestFiles` builds its own `IxClient` from
+ * `getEndpoint()`, which reads `IX_ENDPOINT`, so a real HTTP server on a
+ * loopback port is the whole seam -- and the request log it keeps is exactly
+ * the thing the PRs' measurements were about.
+ *
+ * What these catch, verified by restoring the bugs and watching them go red:
+ *
+ *   - removing the cutoff entirely (the pre-#560 fan-out);
+ *   - the drain stopping on the first pass that placed nothing, which strands
+ *     every good patch between two clusters of bad ones;
+ *   - the stitch marker being cleared by a failure that is not proof, which is
+ *     the #568 stacking itself.
+ *
+ * What they do NOT catch, and are not the right tool for: the finer stopping
+ * rules inside `drainInPasses` -- two-empty-passes versus three, and the
+ * direction flip. Those need a held set shaped precisely, which at this level
+ * means contorting a fixture until it happens to produce one. They are
+ * mutation-validated directly in `commit-breaker.test.ts`, and that division is
+ * deliberate: coarse wiring here, decision tables there.
+ */
+
+/** A backend that answers the endpoints an ingest touches, and records them. */
+class FakeBackend {
+  readonly requests: Array<{ path: string; patches: number }> = [];
+
+  private server: Server | undefined;
+  private rev = 0;
+
+  /** Patch-source substrings this backend refuses, whatever else is healthy. */
+  poison: string[] = [];
+  /** Fail every commit, the Ix#560 shape. */
+  refuseEverything = false;
+  /** Milliseconds to sit on each commit before answering. */
+  commitDelayMs = 0;
+  /** Answer a bulk with 409 naming every patch as already committed. */
+  bulk409AllLanded = false;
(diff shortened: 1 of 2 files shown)


KageBinary:
Reviewed and verified locally. This closes the right seam and the harness has teeth — no findings.

## What I checked

**The loader fallback is safe under the shipped build.** The `new Function` indirection exists partly to keep a specifier away from a bundler, so a bare `import(specifier)` beside it is worth confirming rather than assuming. `ix-cli` builds with `tsc` alone (`build` is `build-core-ingestion.mjs` → `tsc -p tsconfig.build.json`, no bundler), `package.json` is `"type": "module"` and `tsconfig` is `module: Node16`, so the ESM emit preserves the dynamic import verbatim. Under a CJS emit tsc would have downlevelled it into a `require` of an ESM file and the fallback would have been worse than the error it replaces; that is not what happens here. Narrowing the catch to the `dynamic import callback` message is the right shape.

**The suite is deterministic here.** Five consecutive runs on Linux / Node 22, 10/10 each time, no flakes.

**It catches what it claims.** Mutation-checked independently rather than taken on trust — turning the cutoff effectively off (`IX_COMMIT_FAILURE_LIMIT=100000` in `beforeEach`, standing in for the pre-#560 fan-out) turns exactly the two tests red that should go red:

```
× bounds the requests to a backend that refuses everything (Ix#560)
× bounds the replay, and counts it applied, when the re-sends fail too
  Tests  2 failed | 8 passed (10)
```

The other eight stay green, which is the right blast radius — the poison-cluster tests assert an exact error count and are not measuring request volume.

**It does not touch the real `~/.ix`.** Worth checking, because `backend-version.ts` and `backend-status.ts` compute `IX_HOME` from `homedir()` at **module load**, i.e. before `beforeEach` redirects `HOME` — so a stamp write on this path would land in the developer's real home no matter what the fixture sets. Snapshotted all 5,777 files under `~/.ix` before and after a full run: byte-for-byte identical, no writes.

The mechanism that saves it is narrower than it looks, though, and worth a comment in the fake: `ingestFiles` reaches `fetchBackendHealth`, `isLocalEndpoint` is true for `127.0.0.1`, and `recordBackendRelease` runs — it writes nothing only because the fake's `/v1/health` answers `{ version: "1.0.28" }` and the stamp reads `release_version`, which is falsy, so `writeVersionStamp` refuses it. Give the fake a `release_version` (which #599/#600 are about right now) and the suite starts stamping the developer's real `~/.ix/.backend-version`. One line in the health route saying *why* it deliberately omits the field would keep that from being "fixed".

## Merge order — please land #598 first

These two are independent branches off `main`, not stacked, and merging this one alone makes CI measurably flakier. This suite does ten pool teardowns per vitest process instead of one, which is exactly the multiplier #598 reports raising the segfault rate to ~18% per run — the crash that PR was found through. Landing #598 first costs nothing and keeps the harness from being blamed for a crash it only exposed.

## Two nits

- The body says nine cases; there are ten.
- `refuseReplays` refuses every `/v1/patch`, not only replays. Harmless as used — the test that sets it also sets `bulk409AllLanded`, so every per-file commit on that path really is a replay — but the name promises a discrimination the fake does not make, and the next test to set it alone would get something other than what it asked for.

PR #598, opened 2026-09-04
fix(ingest): stop ix map segfaulting at teardown by asking parse workers to end themselves

josephismikhail:
## The bug

`ParsePool.destroy()` shut the pool down with `Worker.terminate()`. Terminating a worker thread that has loaded the tree-sitter native bindings **segfaults the whole process**. The thread does not have to be busy — an idle worker that has parsed a single file is enough, because the crash is in disposing an isolate that still holds the addon.

`destroy()` runs in `ingestFiles`'s outermost `finally`, so this is roughly **one `ix map` in twelve exiting 139 after a completely successful ingest**: patches committed, mtime baseline written, summary printed, and a non-zero exit status for anything that reads it.

That is why it has gone unreported. Interactively it looks like a clean run — the summary is already on screen when the process dies. Only a script, a hook, or a CI step checking `$?` ever sees it.

## Reproduction

Twenty pool teardowns per process, six runs of each variant, Windows / Node 26:

| teardown | result |
|---|---|
| `w.terminate()` — what `destroy()` did | **5 of 6 runs died with SIGSEGV (139)** |
| worker closes its own port on request | 0 of 6 |
| worker calls `process.exit(0)` on request | 0 of 6 |

A control isolating the phase: create + destroy with **no parsing at all** was 0/6. The addon has to have been loaded, which is why every real ingest is exposed and an empty pool is not.

## The fix

`parse-worker` now understands `{ __shutdown: true }` and closes its own port, letting the thread unwind its event loop and dispose its isolate in order. `destroy()` asks, and waits on `'exit'`.

`close()` rather than `process.exit(0)`: both avoid the crash, but `exit()` from a worker can take the process's exit code with it, and this runs during teardown.

**The fallback is as load-bearing as the fix.** A worker wedged in a long native parse never reaches its message loop, so `shutdown` terminates it after a 2s grace period. Waiting unbounded there would be the same CLI hang this file has already produced three separate ways. The timer is `unref`'d so an early teardown cannot hold the process open for two seconds after everything else has finished.

`onError` asks too. It fires on `'error'` as well as `'exit'`, and a worker that merely emitted `'error'` is still a live thread holding the addon. Fire-and-forget, exactly as the `terminate()` there always was.

## Tests

Two new cases in `parse-pool.test.ts`, both **mutation-validated**: putting `terminate()` back in `destroy()` turns both red.

- **the graceful path**, asserted through a marker file the worker writes when it is *asked* to go. The crash is probabilistic, so a test that merely tore pools down would pass against the bug most of the time — this pins the mechanism instead of the symptom.
- **the fallback**, asserting `destroy()` waits out the grace period and then returns, rather than waiting forever.

End to end, 240 create/parse/destroy cycles: **7 of 12 runs segfaulted before, 0 of 12 after.**

Full `ix-cli` suite: 1626 passed, 3 failed — the three `read-containment` symlink tests that always fail locally with `EPERM` (Windows needs admin or Developer Mode) and pass on CI's `windows-2022` runner.

## How it was found

While validating the ingest harness in #597. That harness does ten pool teardowns per vitest process instead of one, which raised the crash rate to ~18% per run and made it visible as an intermittent `Worker exited unexpectedly`. I had previously seen the same `exit=139` in local harnesses and wrongly attributed it to stale stubs holding ports.

### DIFF
--- a/core-ingestion/src/parse-worker.ts
+++ b/core-ingestion/src/parse-worker.ts
@@ -1,7 +1,7 @@
 /**
  * Worker thread entry point for parallel file parsing.
  * Each worker maintains its own Parser singleton (safe — module state is per-thread).
- * Receives: { filePath: string, source: string }
+ * Receives: { filePath: string, source: string } | { __shutdown: true }
  * Posts:    { ok: true, result: FileParseResult } | { ok: false }
  */
 import { parentPort } from 'node:worker_threads';
@@ -9,7 +9,35 @@ import { parseFile } from './index.js';
 
 if (!parentPort) throw new Error('parse-worker must run inside a worker thread');
 
-parentPort.on('message', ({ filePath, source }: { filePath: string; source: string }) => {
+/**
+ * A message asking this thread to end itself, rather than be terminated.
+ *
+ * `Worker.terminate()` tears a thread down from the outside, and doing that to
+ * a thread that has loaded the tree-sitter native bindings SEGFAULTS the whole
+ * process -- the parses need not even be in flight, an idle worker that has
+ * parsed once is enough. Measured on Windows/Node 26 at ~8% per teardown, and
+ * the pool terminates every worker at the end of every run, so roughly one
+ * `ix map` in twelve exited 139 after a completely successful ingest: patches
+ * committed, summary printed, non-zero status for anything reading `$?`.
+ *
+ * Closing the port from INSIDE lets the thread unwind its own event loop and
+ * dispose its isolate in order. Measured 0 crashes in the same experiment.
+ * `ParsePool.destroy` still falls back to `terminate()` if a worker does not
+ * answer, so a wedged thread cannot hang the CLI.
+ */
+export type ShutdownMessage = { __shutdown: true };
+
+type ParseMessage = { filePath: string; source: string };
+
+parentPort.on('message', (msg: ParseMessage | ShutdownMessage) => {
+  if ((msg as ShutdownMessage).__shutdown) {
+    // `close()`, not `process.exit()`. Both avoid the crash, but `exit()` from
+    // a worker takes the process's exit code with it if the main thread is
+    // already on its way out, and this runs during teardown.
+    parentPort!.close();
+    return;
+  }
+  const { filePath, source } = msg as ParseMessage;
   try {
     const result = parseFile(filePath, source);
     parentPort!.postMessage({ ok: result !== null, result: result ?? null });
--- a/core-ingestion/src/parse-worker.ts
+++ b/core-ingestion/src/parse-worker.ts
@@ -22,10 +22,21 @@ if (!parentPort) throw new Error('parse-worker must run inside a worker thread')
  *
  * Closing the port from INSIDE lets the thread unwind its own event loop and
  * dispose its isolate in order. Measured 0 crashes in the same experiment.
- * `ParsePool.destroy` still falls back to `terminate()` if a worker does not
- * answer, so a wedged thread cannot hang the CLI.
+ *
+ * The teardown contract, precisely, because the pool's half of it changed and
+ * this is the only place the worker states it: `ParsePool` never terminates a
+ * worker. If one does not answer, the pool stops waiting and `unref()`s it, so
+ * the CLI can exit while the thread is still alive. Nothing reclaims a wedged
+ * thread before the process ends -- do not write code here that relies on
+ * being killed, and do not "restore" a `terminate()` fallback in the pool,
+ * which is the segfault itself.
+ *
+ * Not exported: `ix-cli` does not depend on this package and inlines the
+ * literal, and this module throws at load outside a worker thread, so an
+ * importer could not use the type anyway. The binding between the two is the
+ * test that runs `ParsePool` against this built file.
  */
-export type ShutdownMessage = { __shutdown: true };
+type ShutdownMessage = { __shutdown: true };
 
 type ParseMessage = { filePath: string; source: string };
 
--- a/ix-cli/src/cli/__tests__/parse-pool.test.ts
+++ b/ix-cli/src/cli/__tests__/parse-pool.test.ts
@@ -1,5 +1,7 @@
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
 import { existsSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { dirname } from "node:path";
+import { fileURLToPath } from "node:url";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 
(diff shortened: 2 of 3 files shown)


KageBinary:
The teardown fix is right and the segfault evidence is convincing. One finding on the busy-worker branch, which I think is a hang regression rather than a nit.

## `destroy()` never returns for a JS-CPU-bound parse, where `main` returned in 2ms

`parse-pool.ts:205` re-arms the grace timer for as long as `this.active.has(w)`, and never terminates. The justification at `:156-168` rests on one premise:

> `terminate()` cannot cut short a native call: V8 raises a termination interrupt that is only checked at JS boundaries, and a tree-sitter parse does not reach one until it returns.
> […] teardown is bounded by the longest in-flight parse, and **always was** — `main`'s bare `terminate()` had the same floor.

That is true for a tree-sitter parse. It is not true for `parseFile`, because seven of its language families never reach tree-sitter at all — `core-ingestion/src/index.ts:2066-2071` returns before `getParser()` for YAML, Dockerfile, SQL, JSON, TOML, Markdown and LaTeX. Those parsers are pure JavaScript scanning arbitrary repo content with ~100 regex operations, including a `RegExp` built at runtime and run against the remaining source (`index.ts:1403`). A worker burning JS **is** at a JS boundary, so the interrupt lands.

Measured, Node 22, worker in a pure-JS loop with no native call:

```
__shutdown posted   -> still alive after 1500ms (it never reaches its message loop)
terminate()         -> resolved in 2ms
```

So on this branch the interrupt that `main` relied on is exactly the one that still works, and the code has stopped using it.

### Reproduction against the real `ParsePool`

Fixture worker stands in for `parseYamlFile` et al. — pure JS, no native call:

```js
parentPort.on('message', (m) => {
  if (m && m.__shutdown) { parentPort.close(); return; }
  const end = Date.now() + 60_000;
  let x = 0;
  while (Date.now() < end) { x = (x + 1) % 97; }   // stands in for a catastrophic regex
  parentPort.postMessage({ ok: true, result: x });
});
```

```ts
const pool = new ParsePool(WORKER, 1, 200);   // 200ms grace, every chance to fire
pool.init();
void pool.parse('/repo/pathological.yaml', 'x: 1');
await new Promise(r => setTimeout(r, 300));   // it is in the burn loop

const outcome = await Promise.race([
  pool.destroy().then(() => 'destroyed'),
  new Promise(r => setTimeout(() => r('hung'), 5000)),
]);
```

| `parse-pool.ts` from | outcome |
|---|---|
| `main` | `destroyed` after **2ms** |
| this PR (`c3d235b`) | `hung` — still pending at the 5s cutoff, and the timer re-arms forever |

Same fixture, same test, only the one file swapped.

### Why it matters here

`destroy()` is in `ingestFiles`'s outermost `finally`, and the busy-at-teardown case is one this PR already concedes is reachable — the re-arm branch exists for it, and its own note describes reproducing it with a 400ms busy fixture. The reachable trigger is not only a pathological file: the run deadline firing with parses in flight puts a busy worker into `destroy()` on an ordinary repo, and `main` cut that short instantly.

The failure mode is also the one this file's comments say it has spent three rounds eliminating: `ix map` prints its summary and then hangs forever, with an `unref`'d timer that cannot rescue it because the worker thread itself refs the event loop.

### Suggested fix

Keep the re-arm, but bound it — cap the total wait and then terminate regardless of `active`:

```ts
static readonly SHUTDOWN_HARD_MS = 30_000;
// ...
const deadline = Date.now() + ParsePool.SHUTDOWN_HARD_MS;
const arm = (): void => {
  timer = setTimeout(() => {
    if (this.active.has(w) && Date.now() < deadline) { arm(); return; }
    w.terminate().catch(() => {}).finally(done);
  }, graceMs);
  timer.unref?.();
};
```

Past the cap a native parse is terminated too, which reintroduces the segfault risk for that one case — but that is the trade this file already makes one branch above, in its own words: *"That can still crash the process, but a hung CLI is certain and this is not."* A crash after a completed ingest is the bug being fixed; an unbounded hang is strictly worse than the 139 it replaces.

A test for it is cheap and mutation-checkable the same way the existing two are: the fixture above with a `graceMs` well under the cap, asserting `destroy()` resolves.

## Everything else checks out

- `w.threadId === -1` after exit, and `postMessage` to a dead worker is a silent no-op — both confirmed here on Node 22, so the early-return and the no-try/catch note hold.
- `close()` over `process.exit(0)` is the right call for the reason given.
- Draining `this.queue` in `destroy()` and counting the stranded tasks closes a real gap, though note it is only observable when `ingestFiles` throws — on the completing path the summary is already built before the `finally` runs.
- The `onError` path keeping `FAULTED_GRACE_MS` short, and terminating unconditionally there, is reasoned through correctly.
- The one local failure I saw, `shuts the REAL parse worker down through the same protocol`, is just `core-ingestion/dist` not being built in my worktree; CI builds it, and the job is green.

KageBinary:
Both findings closed, and the new approach is better than the fix I suggested. Re-verified on `2a8926e`.

## The hang is gone

Same reproduction as before — the real `ParsePool` against a worker doing pure-JS CPU work with no native call, `graceMs` 200 and the ceiling injected at 800ms:

```
>>> destroy() outcome=destroyed after 805ms
>>> in-flight parse: settled
```

On `c3d235b` that was `hung`, still pending at a 5s cutoff and re-arming forever. The hard deadline closes it, and settling `this.active` on the way out closes the second half I had not thought to check: `destroy()` returning on schedule while the task's promise stays pending forever is the same hang one frame further out. Good catch, and the comment explaining that `onError` early-returns once `destroyed` is set — so nothing else would ever have settled it — is exactly right.

## Dropping `terminate()` entirely is the right call

This is a better answer than the bounded-terminate I proposed. My suggestion kept a rule about when killing a thread is safe, and you are right that every such rule had a counterexample — the premise I was arguing from (that `terminate()` works on a JS-bound worker) is true but does not generalise, and building the deadline on it would have reintroduced the segfault for exactly the native case the PR exists to fix.

Verified the replacement premise independently rather than taking the measurement: four workers loaded with the real tree-sitter addon by actually parsing, then wedged in a JS loop they can never leave, then `unref()`'d and the main module returned.

```
unref path:      10/10 clean exits
terminate() path: 10/10 clean exits   (control)
```

The control does not reproduce the crash here, which is expected — you measured it on Windows / Node 26 and this is Linux / Node 22. What it does establish is that the new path is safe on this platform: a process exits cleanly with live, addon-loaded, still-spinning threads left unref'd, which is the load-bearing claim.

## The residual is documented honestly, and named the right consumer

Worth saying out loud because the comment gets it right where an earlier draft got it backwards: `ix watch` is safe because it runs each map as a child process, and the leak that matters is `createInProcessRunner` in the MCP server, which is the default. A leaked thread against a wedging grammar is recoverable; a SIGSEGV takes the server down. That is the correct trade and the comment says where to look if threads ever accumulate.

## Gates

`parse-pool.test.ts` 12/12. Full suite 1653 passed / 2 skipped, typecheck clean, lint 0 errors (41 pre-existing warnings). The one local failure is `watch-dedup`'s `check-ignore` assertion against my worktree's symlinked `node_modules` — Ix#585, fixed by #592, not yours.

Merging this ahead of #597, since that harness does ten pool teardowns per vitest process and is what raised the crash rate enough to find this.

PR #600, opened 2026-09-04
docs(api): document release_version on /v1/health (#599)

### LINKED ISSUE #599: docs: /v1/health response omits release_version from the OpenAPI spec
## Summary

`docs/api/openapi.yaml` documents the `/v1/health` response as `HealthResponse` with only `status` and `schema_version`, but the backend also sends `release_version` on every response. A client reading the reference would never learn the field exists.

## Evidence (live backend, 2026-09-04)

```
$ curl -s http://localhost:8090/v1/health
{"status":"ok","schema_version":3,"release_version":"1.0.28"}
```

Spec (docs/api/openapi.yaml:976-980):

```yaml
HealthResponse:
  type: object
  properties:
    status: { type: string }
    schema_version: { type: integer, description: On-disk graph format version }
```

`release_version` does not appear anywhere in the spec.

## Suggested fix

Add the field to `HealthResponse` (observed value format is a semver string):

```yaml
release_version: { type: string, description: Backend release semver }
```

This is exactly the drift class that #586's proposed commit-time parity gate would catch — the concrete instance, filed separately since it is independently actionable.

Alot1z:
Closes #599.

## The gap (measured)

The live backend answers `/v1/health` with three fields, the schema documents two:

```json
{"status":"ok","schema_version":3,"release_version":"1.0.28"}
```

`docs/api/openapi.yaml` → `HealthResponse` documented only `status` + `schema_version`. The client type (`ix-cli/src/client/types.ts`) has carried `release_version?: string` since the release-tracking work; the schema was the only laggard.

## The fix

One optional property added, matching the client type exactly (absent from backends older than Ix-memory#157 and from images not built by the release pipeline):

```yaml
HealthResponse:
  type: object
  properties:
    status: { type: string }
    schema_version: { type: integer, description: On-disk graph format version }
    release_version: { type: string, description: The published release the running container was built as, when it knows (#599) }
```

## Would the commit-time parity gate have caught this? No.

The gate on #587 compares four surfaces: paths+methods called by the client, `$ref` integrity, schema-name ↔ types.ts-name existence, and README endpoint coverage. A property added *inside* an already-known schema is invisible to every pass — the gate cannot see response-shape fields. Verified directly: with the gate's own tree plus this one-line addition, parity stays `42 documented, 42 called, 13 schemas, 13 refs — exit 0`; with main's pre-#587 yaml it reports the pre-existing drift the gate exists to fix. Response-shape drift is a real hole in the gate's surface; naming it here so it can be extended deliberately rather than discovered by accident.

## Coordination

This touches `docs/api/openapi.yaml` in the same region #587 also edits (schema components vs. the decisions path+requestBody) — no textual overlap, but whichever merges second may need a trivial rebase.

### DIFF
--- a/docs/api/openapi.yaml
+++ b/docs/api/openapi.yaml
@@ -977,6 +977,7 @@ components:
       properties:
         status: { type: string }
         schema_version: { type: integer, description: On-disk graph format version }
+        release_version: { type: string, description: The published release the running container was built as, when it knows (#599) }
     CapabilitiesResponse:
       type: object
       properties:


KageBinary:
The gap is real and the fix is the right one line — but as written the line does not parse into the schema you quoted in the body.

## The comma splits the flow mapping

```yaml
release_version: { type: string, description: The published release the running container was built as, when it knows (#599) }
```

Inside a `{ }` flow mapping a comma is an entry separator, and the description is unquoted, so YAML reads that comma as ending the `description` entry and starting a new one. What the file actually parses to:

```
'type': 'string'
'description': 'The published release the running container was built as'
'when it knows (#599)': None
```

`release_version` gains a third key — `when it knows (#599)`, with a null value — which is not an OpenAPI schema keyword. The PR body shows the intended two-key shape, so this is a typo rather than a decision, but nothing in the repo would have caught it: I checked, and no workflow or script parses `docs/api/openapi.yaml` at all.

It is also the only line in the file with this shape — every other flow-mapping `description` is comma-free, including the `schema_version` line directly above it, which is why the pattern held until now.

## Fix

Quote it:

```yaml
release_version: { type: string, description: "The published release the running container was built as, when it knows (#599)" }
```

Parses to exactly the three-key `{type, description}` — sorry, two-key — schema the body describes. Worth confirming with `python3 -c "import yaml;yaml.safe_load(open('docs/api/openapi.yaml'))"` and printing that one node before pushing.

## Otherwise it checks out

- `ix-cli/src/client/types.ts:182` is `release_version?: string`, so `type: string` + optional matches the client exactly.
- The rest of the document parses cleanly, and the addition sits in `components.schemas` where #587 does not touch — the rebase note in the body is right.
- Your point that the #587 parity gate cannot see response-shape drift is correct and worth carrying over to that PR as a stated limit of the gate rather than leaving it to be rediscovered. This one-line miss is a small illustration of the same thing: nothing mechanical is reading this file today, in any shape.

Alot1z:
Closing #600 in #602's favor — it is my earlier attempt at the same #599 fix, and #602 supersedes it with a strictly more complete state of the reference:

- **The YAML defect you caught here is already absent from #602**: the spec line there is `release_version: { type: string, description: Backend release semver }` — short enough to carry no comma, so the flow mapping parses as intended. Your comma-splitting finding was correct for this branch's longer line.
- **#602 also covers two surfaces this branch missed**: the `HealthResponse` TS reference block in the README and the response-example block, so all four doc surfaces agree.

Nothing of value is lost in closing this one. Thank you for the review — the flow-mapping point made the surviving PR better.

PR #601, opened 2026-09-04
fix(ix-cli): stop ambient home-dir markers from vetoing multi-repo detection

Alot1z:
## Summary

Fixes a latent product bug in `ix-cli/src/cli/system.ts`: `detectSystem` misclassified multi-repo collections as "inside one repo" (returning `undefined`) when an ancestor of the mapped path contains a stray workspace marker. Found via four locally-failing tests in `system.test.ts` that CI never sees.

## The two compounding flaws

1. **Existence-only marker check.** `hasWorkspaceMarker` used `fs.existsSync`, so a *directory* named `WORKSPACE` (an app's data folder — e.g. `C:\Users\<user>\WORKSPACE\default\...` on the reporting host) counts as a Bazel workspace root even though a Bazel `WORKSPACE` is a file. The same class applies to every listed marker (`nx.json`, `lerna.json`, …).
2. **Unbounded ancestor walk.** `isInsideSingleRepo` walked to the filesystem root, so an ambient marker at the user's home vetoed *every* path beneath it. Real-world impact: any user whose home holds a stray marker folder or a dotfiles `.git` silently loses multi-repo detection for **all** of their independent clones — real `ix map` runs, not just tests.

## Why CI never saw it

Runner fixture trees (`/tmp`, `/private/var/folders/...`, `D:\a\_temp`) are never beneath a marked home, so the four SYSTEM-expecting tests only fail where the temp dir sits under one. The other 13 tests in the file expect `undefined`, so they were green for the wrong reason on affected hosts.

## The fix (2 files, +39/−3)

- `hasWorkspaceMarker`: marker presence now requires `statSync(...).isFile()` — a directory named like a marker is not a workspace root.
- `isInsideSingleRepo`: the ancestor walk stops at the user's home (`os.homedir()`, case-insensitive compare on win32). Home is still judged when it *is* the mapped path; it just never classifies paths beneath it. Monorepo subdirs (`ws_npm/packages`, `mono/packages`) still veto correctly — their markers sit between the fixture and home.

## Mutation proof

With `system.ts` reverted to the old code: **5 failures** — the 4 originals plus the new regression pin — on *every* platform, because the pin is hermetic (it fails on old code regardless of host). With the fix:

- `system.test.ts`: 18 passed | 2 skipped (POSIX-only tests, skipped by design on win32)
- Full ix-cli suite: 94 files, **1663 passed | 21 skipped, zero failures**
- `tsc --noEmit` clean

## CI-relevant behavior unchanged

Under runner conditions the old code's marker lookup already returned false at every fixture level (no marked ancestor), so the fix is a no-op there — except that the new regression test now *protects* CI against this regression class. The behavior change is confined to the previously-broken case: real multi-repo detection under a home containing ambient markers.

## Sibling-pattern scoping check (result stated honestly)

Searched `ix-cli/src` for sibling unbounded ancestor walks and existence-only marker checks in the same family:

- `system.ts` was the **only** `dirname`-in-loop ancestor walk in `ix-cli/src`; every other `dirname` use is single-level (`__dirname`, temp-file placement, one-level tsconfig resolution). **No sibling to fix.**
- `read.ts:232` already pairs `existsSync` with `isFile()` — the correct pattern, no change needed.
- `docker.ts` compose-file existence pick fails loudly on use rather than silently vetoing — acceptable as-is.

No sibling defects found; nothing deferred.


KageBinary:
Reproduced the bug and confirmed the diagnosis — but the PR contains two independent changes, and only one of them is needed. The other is untested and has a real cost.

## The bug is real, and I reproduced it

Simulated the reporting host — a home holding a stray `WORKSPACE/` **directory**, with `tmpdir` beneath it — and ran `system.test.ts` against `main`:

```
HOME=<fake>  TMPDIR=<fake>/tmp
× detectSystem recognizes all members
× detectSystem builds the ground-truth repoDeps graph (consumer -> lib)
× treats a plain folder of independently-cloned repos as a multi-repo SYSTEM
× still treats a no-.git, no-workspace folder of projects as a SYSTEM
× ignores a stray DIRECTORY named like a workspace marker
  Tests  5 failed | 15 passed (20)
```

Exactly the four you describe, plus your new pin. The `statSync(...).isFile()` fix is right: every marker in that list is a file, and `existsSync` was never the correct test for any of them.

## The home-boundary walk-stop is not needed for this bug

Kept your `isFile()` change and reverted **only** the `sameDir(parent, home)` line, on the same marked-home fixture:

```
  Tests  20 passed (20)
```

All five failures are the `existsSync`-vs-`isFile` bug alone. The reported symptom is `C:\Users\<user>\WORKSPACE\default\...`, a directory — so once a directory stops counting as a marker, there is nothing left at home to veto anything.

That also means the walk-stop ships with **no coverage**: no test in the suite goes red when it is removed. It is the riskier of the two changes and the only one nothing pins.

## And it makes `detectSystem` depend on who is running it

`isInsideSingleRepo` now consults `os.homedir()`, so the same tree can classify two ways. Same fixture both times — two independent clones under a directory whose parent carries a genuine `package.json` with `"workspaces"`:

| | `main` | this PR |
|---|---|---|
| `HOME` = the marked parent | `undefined` (inside one repo) | **`SYSTEM 5286847b` members=repo-a,repo-b** |
| `HOME` = elsewhere | `undefined` (inside one repo) | `undefined` (inside one repo) |

`main` gives one answer for that tree. This PR gives two, chosen by an environment variable that has nothing to do with the tree.

That matters more than a normal classification wobble, because the answer is an identity that gets persisted: `detectSystem` returns `systemId: stableId('system:' + abs)`, and the system either exists or does not. A CI job running as root (`HOME=/root`), a container with `HOME=/`, and a developer mapping the same checkout under their home now disagree about whether that path is a system at all — and the `workspace_id convergence (solo vs co-ingest)` cases in this very file are about those answers agreeing.

The declaration being skipped is not always ambient, either. `hasWorkspaceMarker`'s own doc calls a workspace config *"an explicit 'these subdirs are MY packages, not separate repos' declaration"*. A `package.json` with `workspaces`, a `Cargo.toml` with `[workspace]`, or a `go.work` at home is that declaration; the walk-stop cannot tell it from a dotfiles `.git`, so it ignores both.

## Suggestion

Split it. The `isFile()` half is a clean, well-pinned fix for a real bug — I would land that on its own today.

If the ambient-`.git`-at-home case is worth fixing (it is a fair concern — `git init` in a home directory does happen), it deserves its own PR with its own failing test, and a narrower rule than "never look at home". Options worth weighing there: skip only `hasOwnGit` at home while still honouring an explicit workspace declaration, since `.git` is the ambient one and `workspaces`/`[workspace]`/`go.work` are deliberate; or stop the walk at the git *boundary* (`git rev-parse --show-toplevel`) rather than at `$HOME`.

## Two smaller notes

- `sameDir` compares `path.resolve`d strings, which does not resolve symlinks. A symlinked home — `/home/x` → `/mnt/home/x`, and macOS's `/var` → `/private/var` for anything under `tmpdir()` — never matches, so the stop silently does not apply there. If the line stays, `fs.realpathSync` on both sides would make it mean what it says.
- The new test is genuinely hermetic (it fails on old code regardless of host), which is the right property — worth keeping whichever way the split goes.

KageBinary:
Pushed `a174203` to this branch, taking the split I suggested rather than leaving it for you — the `isFile()` half is a clean fix for a real bug and I would rather it land today than wait on the other half.

## What I kept

`hasWorkspaceMarker` requiring `statSync(...).isFile()`. That is the fix. Reproduced your report by pointing `HOME` at a directory holding a stray `WORKSPACE/` folder with `tmpdir` beneath it — 5 of 20 tests in `system.test.ts` fail on `main`, and requiring markers to be files turns **all five** green on its own, with the home boundary reverted and nothing else changed. Your new hermetic regression case pins exactly that half, which is the right thing for it to pin.

## What I dropped, and why

The `sameDir(parent, home)` walk stop, plus the now-unused `node:os` import and `sameDir` itself.

It is not needed for the reported bug — that is the measurement above, not an opinion. And it carries a cost the report does not justify:

**It makes `detectSystem` depend on who is running.** Same fixture both times — two independent clones under a parent carrying a genuine `package.json` with `"workspaces"`:

| | `main` | with the boundary |
|---|---|---|
| `HOME` = the marked parent | `undefined` | **`SYSTEM 5286847b`** |
| `HOME` = elsewhere | `undefined` | `undefined` |

`main` gives one answer for that tree; the boundary gives two, chosen by an environment variable that has nothing to do with the tree. That matters more than a normal classification wobble because `detectSystem` returns `systemId: stableId('system:' + abs)` and the system either exists or does not — a CI job running as root (`HOME=/root`), a container with `HOME=/`, and a developer mapping the same checkout would disagree about whether that path is a system at all. The `workspace_id convergence` cases in this very file are about those answers agreeing.

**It cannot tell a deliberate declaration from an ambient one.** `hasWorkspaceMarker`'s own doc calls a workspace config *"an explicit 'these subdirs are MY packages, not separate repos' declaration."* A `package.json` with `workspaces`, a `Cargo.toml` with `[workspace]`, or a `go.work` at `$HOME` is that declaration; the boundary ignores it exactly as it ignores a dotfiles `.git`.

**Nothing pinned it.** Removing the line leaves all 20 tests green — including your new one.

One more thing worth knowing if it comes back: `sameDir` compares `path.resolve`d strings, which does not resolve symlinks. A symlinked home — `/home/x` → `/mnt/home/x`, or macOS's `/var` → `/private/var` for anything under `tmpdir()` — never matches, so the stop silently would not have applied there anyway.

## The other half is still worth doing

The ambient-`.git`-at-home case is real — `git init` in a home directory happens. It just wants its own failing test and a narrower rule. Two shapes that would not have the determinism problem: skip only `hasOwnGit` at home while still honouring an explicit workspace declaration (`.git` is the ambient one; `workspaces`/`[workspace]`/`go.work` are deliberate), or stop the walk at the git boundary (`git rev-parse --show-toplevel`) rather than at `$HOME`. I left that as a comment on the walk so the reasoning is not lost.

Verified on the merged tree: 20 passed in `system.test.ts` both with a normal `HOME` and with the marked-home fixture that reproduces your report; typecheck clean.

PR #603, opened 2026-09-05
feat(install): drive harness presence from toolscan discovery (supersedes #591)

KageBinary:
Supersedes #591, carrying every one of @Alot1z's commits plus one merge commit. Opened here only because the fork branch would not accept a push (`remote rejected … permission denied`); #591 stays open for @Alot1z to close or take over.

## Why it needed a merge at all

`main` moved a long way today — #587, #589, #597, #598, #602 and the four parser PRs all landed. This branch conflicted in four files, and **one of the conflicts was consequential rather than textual**:

`ix-cli/scripts/check-api-parity.mjs` — this branch carries a synced copy of #587's gate (`f1cf0b5c`, "sync the carried parity gate with the #587 rebuild"), and that copy predates #587's own follow-up fix `f37fd49`. A naive merge would have **silently reverted** it:

```js
(?:<[^>]*>)?   // this branch's copy — cannot cross a nested `>`
(?:<.*?>)?     // main, after #587's fix
```

The difference is real, not cosmetic: `this.get<Record<string, unknown>>("/v1/x")` walks straight past the guard under the first and is caught under the second. Taken from `main` wholesale — the file is now byte-identical to it, and the carried copy is gone.

The other three:

- `.github/workflows/ci.yml` — #587 added an `api-parity` job to the same aggregate gate this branch extends. Resolved to the **union**: `ci-success` now requires `api-parity` *and* both harness smoke jobs, rather than either replacing the other.
- `README.md` and `scripts/install-skill.sh` — kept this branch's multi-harness wording and installer. #589's narrower "Claude Code / .agents" text describes the surface *before* toolscan discovery; this PR is what makes the broader claim true.

## Verified on the merged tree

```
check-api-parity: 42 documented, 42 called, 13 schemas, 13 refs — exit 0
bash -n scripts/install-skill.sh                                — OK
mcp-install + discovery + skill-harnesses tests                 — 78 passed, 1 skipped
```

## The review that got it here

All findings from the rounds on #591 were verified fixed on `608c986` before this merge, each mutation-checked rather than read:

- **the name-only toolscan entry** (the reopened P0): `hostInstalled` now requires `paths.has` as well as `names.has`. Dropping the conjunct turns *"degrades a name-only toolscan entry to the embedded probes"* red.
- **the unpinned shell-side resolver**: `skill-harnesses.test.ts` now has the `resolveToolscan (the security pin)` case. Re-adding a bare-PATH fallback to the `.mjs` turns it red — which was the whole point, since the duplicated copy was where a PATH fallback could come back green.
- **the new `quoteForCmd` drift guard** holds: I fuzzed the TS and `.mjs` implementations against each other over 20,000 random arguments (space, `"`, `\`, `&`, `(`, `)`, `^`, `|`, `>`, `<`, `%`, `!`), **0 divergences** — including the trailing-backslash-before-a-quote cases where a hand-written CMD quoter usually diverges.
- both nits fixed: EOF newline, and the smoke fixture reads `process.env.HOME` instead of hardcoding a path.

https://claude.ai/code/session_01FnMJopSVeUMFifL74DJ2Gk


KageBinary:
Not needed — the push to @Alot1z's fork succeeded once I used the branch's real name (`feat/skill-install-toolscan`; my first attempt used the wrong ref and GitHub reported that as `permission denied`, which I misread as the fork being locked).

#591 now carries the same merge commit `b652008` directly, so it keeps its own history and authorship. Continuing there.

PR #605, opened 2026-09-05
feat(cli): render the repo logo as the setup-notice banner

Alot1z:
## Status: DRAFT — part of a larger train

This pull request is intentionally a **draft** and one component of a coordinated set the author is landing together (terminal logo banner · installer polish from the merged-surface audit · toolscan output-contract hardening). It will be marked ready only when every part of the train is green and the whole set is released at once.

Feedback and suggestions are welcome at any stage. Each point is addressed on this branch — folded in or answered — and merged without conflict. The content below is accurate; the head may still move.

---

## Rendered preview

Ix logo banner as printed at first run — truecolor, brand backdrop

Width 48 is the CLI default (`--width 48`); a wider render ships alongside it at `output-samples/banner-80-truecolor.png` (same branch). Preview PNGs live on the `meta/logo-previews` branch so the PR tree carries only what ships.

## What this ships

The one-time setup notice (`ix` first run in a workspace) prints the Ix logo as a terminal banner instead of the plain text heading.

- `ix-cli/scripts/render-logo.mjs` renders `ix-cli/assets/logo.png` at print time. Zero dependencies: PNG chunk decode → inflate → unfilter → coverage-thresholded supersampling → flat 5-tone snap → half-block output. The PNG is the only maintained art; the terminal render is derived from it on every run. Both inputs ship **inside the package** (npm pack + release tarball), with a `npm pack` pin in the tests and a staging gate in `release.yml` so a future packaging change cannot silently strand them again.
- Truecolor with ANSI-256 and ASCII fallbacks. `--color auto` honors `NO_COLOR`, `TERM=dumb`, `FORCE_COLOR`; the no-color path emits zero escape bytes.
- `--bg none` paints only the logo's own pixels (unbacked `▀`/`▄` half-blocks), so the mark reads cleanly on light terminal themes. `--bg brand` (default) keeps the current backdrop. `--json` reports `bg` honestly.
- Absent-safe: a layout without the renderer or the asset falls back to the plain text heading — the banner is decoration, never load-bearing. stderr-only, so `--format json|llm` output stays clean.
- Golden fixtures in `output-samples/` (truecolor/256/ascii × brand/none) with output-stability pins.
- `assets/logo.png` already exists on main — no binary asset is added, only the renderer, wiring, tests, and goldens.
- Rebased onto main at the last branch update (`39d0734`); main's workspace-canonicalization change in `bootstrap.ts` is preserved (verified on a fresh clone of main). Upstream main has since advanced to `e8ab1926` (the #547/#559 exit-code wave merged, 2026-09-05); a final rebase onto that head runs before this PR is marked ready.

## Tests

- Focused suite: 21 pins in `bootstrap-notice.test.ts` — stderr-not-stdout, absent-safe renderer/asset, NO_COLOR zero-escapes, exit-code contracts, `--bg none` invariants (backdrop constant never appears as a background; library ≡ CLI byte-for-byte; JSON honesty), and byte-identity against the goldens.
- Mutation-checked locally: breaking the no-color contract turns the NO_COLOR pin red; disabling the `--bg none` emission path turns its pin red. Both restored green.
- Full `ix-cli` suite on the merged head (Windows): 95 files, 1704 passed / 21 skipped; `tsc --noEmit` clean.

## CI state (updated after the hermeticity fix)

Head `17d2da44` (fix round for the review findings): **all checks green except the single documented CodeQL infra-race run** (`101317167251`, started pre-recreation) — the six originally failing jobs (`Test` × ubuntu/macos × node 22/24, `CI Passed`) were fixed by the hermeticity change below and have passed on the head since the recreation re-ran the suite. The fix made the banner pins hermetic across runners — the half-block pin now pins its env (runners can resolve the ascii fallback otherwise), and the two lib≡CLI byte-identity pins got a realistic budget for the in-process full-PNG decode under v8 coverage (measured ~0.7 s → ~6.8 s with coverage instrumentation; assertions unchanged).

One residual, not a finding: the same documented **CodeQL** infra race (run `101317167251`, "The file may have changed since it was checked" — it started before the recreation, and the branch head moved during analysis). It needs a maintainer-side re-run (admin rights; the author's token is contributor-only) or clears on a branch update. **Update 2026-09-05:** the rendered preview PNGs moved the head to `c05c3a77`, so the suite re-runs on that head — including CodeQL.

## Notes

- The aesthetic approach (rendering a repo asset as a first-run terminal banner) takes some inspiration from skills.sh (vercel-labs/skills); the implementation here is original zero-dependency code. Per the author's convention, no attribution appears in any committed file.
- Part of a coordinated train with the installer README-claims fix (#609); this PR will be marked ready together with it when every part is green.


KageBinary:
Reviewed at `c05c3a77`. The renderer itself is good work — I ran it and it produces a clean mark. Three findings, the first blocking.

## 1. The banner can never render for an installed user

`renderBanner()` resolves `dirname(import.meta.url)/../../..` and requires `scripts/render-logo.mjs` and `assets/logo.png` under it. Both live at the **repo root**, outside the `ix-cli` package — and neither delivery path carries them.

**npm:** `@ix/cli` is published from `ix-cli/`, so files one level above it are never packed.

```
$ cd ix-cli && npm pack --dry-run --json
total files: 473
render-logo      -> 0 hit(s)
logo.png         -> 0 hit(s)
assets/          -> 0 hit(s)
output-samples   -> 0 hit(s)
```

**Release tarball:** `.github/workflows/release.yml` stages exactly three trees —

```
mkdir -p "$STAGING/cli" "$STAGING/core-ingestion" "$STAGING/compass"
cp -rL ix-cli/dist ix-cli/node_modules ix-cli/package.json "$STAGING/cli/"
```

So in the installed layout `banner.js` sits at `$IX_HOME/cli/dist/cli/`, resolves `root` to `$IX_HOME`, and looks for `$IX_HOME/scripts/render-logo.mjs` and `$IX_HOME/assets/logo.png`. Neither is ever written there:

```
resolved root = <install>/
  scripts/render-logo.mjs ABSENT -> renderBanner() returns null
  assets/logo.png         ABSENT -> renderBanner() returns null
```

The absent-safe fallback is what makes this hard to catch: every install silently prints the plain `Ix` heading, exactly as it does today, and nothing reports that the feature did not run. The tests all pass because they execute from a repo checkout, which is the one layout where the files are present. Running `node scripts/render-logo.mjs --width 48` from a checkout does render correctly — the gap is delivery, not the renderer.

To land this, the two inputs need to ship: move them under `ix-cli/` (or stage them explicitly in `release.yml`), and pin it in CI the way `compass/.version` is pinned, so a future staging change cannot silently make the banner unreachable again.

## 2. The CodeQL failure is a real high-severity finding, not the infra race

The PR body reads the alert as "the file may have changed since it was checked — it started before the recreation, and the branch head moved during analysis." That sentence is the **rule's own description of a TOCTOU in the code**, not CodeQL reporting a race against the branch head. The alert on the current head:

```
RULE: js/file-system-race  [high]
FILE: scripts/render-logo.mjs:45
```

```js
try { stat = statSync(fileArg); } catch { throw new LogoError(`file not found: ${fileArg}`, 1); }
if (stat.size > MAX_FILE_BYTES) throw new LogoError(...);
try { buf = readFileSync(fileArg); } catch (e) { ... }
```

`statSync(path)` and `readFileSync(path)` resolve the path independently, so `MAX_FILE_BYTES` guards a file that need not be the one read. There is also no `isFile()` check, and `readFileSync` on a FIFO blocks forever — with `--file` being a user-supplied flag, that is reachable.

This repo already has a house pattern for exactly this: open the path once (`O_NONBLOCK`), `fstat` **the handle**, check `isFile()` and size on that stat, then read from the same handle. Reusing it closes both the TOCTOU and the FIFO hang, and clears the required check.

## 3. Two preview PNGs are committed for the description's benefit

`output-samples/banner-48-truecolor.png` and `banner-80-truecolor.png` (6.4 KB) are referenced nowhere in the repo — no test, no script, no doc. The `.ans` files are genuine goldens and are byte-pinned by `bootstrap-notice.test.ts`; these two are not. They exist to render in the PR body, and the body links them through `raw.githubusercontent.com/Alot1z/Ix-remap/...` rather than from the repo anyway. Worth dropping before this leaves draft.

---

One design note, not a finding: `renderBanner()` spawns a Node subprocess on the first-run path. It is cached and bounded at 10 s, and first-run happens once per workspace, so the cost is defensible — but since the renderer is already zero-dependency ESM, importing it in-process would avoid the spawn entirely and remove the `existsSync` probe that #1 turns on.

Alot1z:
Thanks for the review — all three findings are fixed on the branch (head `17d2da44`, rebased on main `8c0e6b00`), plus the design note adopted.

## 1. Banner unreachable in installed layouts — fixed, and pinned so it cannot regress

- `scripts/render-logo.mjs` → `ix-cli/scripts/render-logo.mjs` and `assets/logo.png` → `ix-cli/assets/logo.png`. Both inputs now live inside the package, so `banner.ts` needs no repo-root path math at all: the dynamic import resolves relative to the compiled file in every layout.
- Verified against the two delivery paths from the finding: `npm pack --dry-run --json` now lists the renderer and the asset, and the compiled `dist/cli/banner.js` renders the banner when run from an installed-layout tree (25 banner lines, non-null) instead of silently degrading.
- Two structural pins added so this cannot silently strand again: a test asserts the pack manifest carries both inputs, and `release.yml` stages both files into `$STAGING/cli/` with a gate asserting their presence in the staging tree.
- The absent-safe contract is preserved: the loader probes for the package-relative inputs first and only then imports, so a layout without them still degrades to the plain heading (found and fixed while testing: the in-process rewrite initially made a missing renderer fatal at module load in the watch-child runtime cache — a third layout the probe now covers).

## 2. CodeQL `js/file-system-race` — fixed with the house open-once/fstat pattern

`decodePng` now opens the path once with `O_NONBLOCK`, `fstat`s **the handle**, checks `isFile()` and size on that stat, and reads from the same fd. The stat-then-read TOCTOU is gone, FIFOs can no longer block the read (nonblocking open rejects them via the `isFile()` check), and the size guard now applies to the file actually read.

## 3. Preview PNGs — dropped from the PR tree

Both `output-samples/banner-*.png` files are removed from the branch. The `.ans` goldens (byte-pinned by `bootstrap-notice.test.ts`) stay. The preview PNGs live on a separate `meta/logo-previews` branch and the PR description links them there, so the PR tree carries only what ships.

## Design note — adopted

`renderBanner()` now imports the renderer in-process (dynamic import, no subprocess, no spawn cost), behind the absent-safe probe above.

## Validation

- Full `ix-cli` suite on `17d2da44`: 1755 passed / 0 failed (21 skipped = parser smoke, run separately); `tsc --noEmit` clean.
- New pins: npm-pack manifest carries renderer + asset; `release.yml` staging gate; byte-identity goldens unchanged (the read-path rewrite produces identical render output).
- Rendered preview PNGs were regenerated from the goldens for the PR description after finding the first conversion painted unset backgrounds black — unrelated to the shipped code (the terminal render is unaffected), noted here since the preview images changed with this push.

Alot1z:
Releasing this PR — it ships the first-run terminal logo banner.

Head at mark-ready: `2f9604772c19575ab9207138f6363523634dd44b`. CI on this head at posting time (fresh probe): 28 checks succeeded, 3 skipped, 0 failed. Base = current main (`8c0e6b00`).

What this PR ships: the first-run logo banner — `ix-cli/scripts/render-logo.mjs` renders the package-resident `ix-cli/assets/logo.png` at print time (zero dependencies: PNG decode → inflate → unfilter → supersample → flat 5-tone snap → half-block), truecolor with ANSI-256 and ASCII fallbacks, `--bg brand`/`--bg none`, honest `--json`, and an absent-safe fallback to the plain text heading in every delivery layout — npm pack carries the renderer, types, and asset (pinned by test) and release staging refuses without them. Asset reads are race-free (open-once → fstat → isFile → read the same fd). Rendered preview PNGs live on the `meta/logo-previews` branch, embedded in this PR's description; the branch itself carries only referenced assets. Golden fixtures + hermetic pins throughout.

The docs-only companion (#609, installer README claims) is already merged on main. Feedback landed in the draft period has been folded in; nothing else changes in this PR.

Alot1z:
Cross-linking the two open proposal issues so their timelines point at this PR (they currently reference each other but not this PR; no back-reference needed from this PR — the issues are linked above in the development sidebar):

- #611 — installer output polish (U1 id-labeled install lines, per-harness summary, documented `--dry-run`): directly relevant here because this PR's banner prints at first-run alongside the installer output. The output-contract changes proposed there land independently; the banner stays output-compatible either way.
- #612 — interactive `ix install` proposal (banner + skills/plugin/MCP menu): this PR ships the banner surface that proposal builds on (`ix-cli/scripts/render-logo.mjs` + `renderBanner()`).

Both are proposals only — nothing in this PR changes for or because of them.

KageBinary:
Re-reviewed at `2f960477`. All three findings from my earlier review are closed, and the first two are closed properly rather than papered over.

**1. The delivery gap is fixed at the root.** `render-logo.mjs` and `logo.png` now live under `ix-cli/`, so `npm pack` carries them, and `release.yml` stages them explicitly:

```
mkdir -p "$STAGING/cli" "$STAGING/cli/scripts" "$STAGING/cli/assets" …
cp -rL ix-cli/scripts/render-logo.mjs ix-cli/scripts/render-logo.d.mts "$STAGING/cli/scripts/"
cp -rL ix-cli/assets/logo.png "$STAGING/cli/assets/"
test -f "$STAGING/cli/scripts/render-logo.mjs" && test -f "$STAGING/cli/assets/logo.png" …
```

The `test -f` guard is the part that matters most — it is what stops this silently regressing to the absent-safe fallback the next time someone edits the staging block, which was the whole reason the original was invisible.

**2. The CodeQL finding is genuinely fixed, with the repo's own pattern.** `readPngFile` opens once with `O_RDONLY | O_NONBLOCK`, `fstat`s the **handle**, checks `isFile()` and the size cap against that inode, reads from the same fd, and closes in a `finally`. That closes the TOCTOU and the FIFO hang together, and it is the same shape the rest of this codebase uses for bounded reads of untrusted paths. CodeQL is green.

**3. The two preview PNGs are gone.** Only the `.ans` goldens remain, and those are byte-pinned by the test.

All non-skipped checks green.

**On sequencing:** I am cutting v0.10.6 from `main` first, without this. Not a reservation about the change — the release exists to ship the #556 ingest fix that is currently costing a 2,000-file repo two files and its whole stitch registration, and this PR modifies `release.yml` itself. Exercising a changed release pipeline for the first time during that release adds risk for no benefit to it. This lands right after and rides v0.10.7, where a staging mistake costs a re-tag rather than a blocked fix.

KageBinary:
Final check before merging: the delivery fix is real, not just declared. Re-ran the same `npm pack --dry-run` that found the gap:

```
  scripts/render-logo.mjs    -> 1 hit(s)   (was 0)
  assets/logo.png            -> 1 hit(s)   (was 0)
```

So the banner will actually render for an installed user, which it could not have done in any prior revision of this PR. Together with the explicit `release.yml` staging and its `test -f` guard, both delivery paths now carry the inputs and a future edit to the staging block fails the release rather than silently reverting to the plain-text fallback.

Merging for v0.10.7, as said above — v0.10.6 went out without it so the ingest fix would not be riding a changed release pipeline. That reason has expired now that it has shipped.

PR #613, opened 2026-09-05
fix(ingest): stop the harness timing out under load, and close four review findings

josephismikhail:
Follow-up to #597, which merged before its review came back. Four findings from that review, plus a flake the review could not see because it only appears under the full suite's parallelism.

## The flake (not in the review — found by running the suite)

Every test in `ingest-files.test.ts` runs a **real** ingest: discovery, a worker-thread parse pool, and an HTTP round trip per commit, over 30 files. On vitest's 5s default timeout. About a second each on an idle machine, which is why it went in without an explicit one.

Under load that is not true. Measured on `main`:

| Condition | Ix#560 test |
|---|---|
| full `ix-cli` suite | **failed 1 of 4 runs** |
| heavier repo-root suite running too | **failed 3 of 3** |
| after this change | 0 of 5 |

A test that only times out on whichever CI leg happens to be busiest reads as a real regression, on six legs. `vi.setConfig` now sets the timeout from the work these tests actually do rather than inheriting a default sized for unit tests.

## Findings from the #597 review

**medium — `afterEach` asserted before it cleaned up.** The `unknownPaths` assertion ran *before* `backend.stop()`, the env restore and both `rmSync` calls, so in exactly the case that assertion exists for, the throw skipped all of them: the HTTP server stayed listening (an open handle that can stop the vitest worker exiting, turning one named failure into a job timeout), both temp trees leaked, and `HOME`/`USERPROFILE`/`IX_ENDPOINT` stayed pointed at the fixture. Because `saved` is shared across the describe, the next `beforeEach` then snapshotted those polluted values and the file's final restore wrote the fixture `HOME` back into the process. Teardown is in a `finally` now.

**low — the fake never sent `status`.** `PatchCommitResult` declares it required and the ingest path branches on it, so every commit came back with `result.status === undefined` and the harness could never produce an `Idempotent` or `BaseRevMismatch` answer. Those are the two branches that decide whether a patch lands in `patchesApplied` or `commitErrors`, and therefore whether the mtime baseline is written at all. A regression flipping the applied test from "not `BaseRevMismatch`" to `=== "Ok"` would have counted zero patches applied against the real backend while every test here stayed green.

**low — the loader's vm fallback matched on English.** It now matches `err.code` (`ERR_VM_DYNAMIC_IMPORT_CALLBACK_MISSING`, plus the `..._FLAG` variant by prefix), keeping the message test only for an error that arrives without a code. Verified in both directions rather than assumed:

- message fallback disabled → suite still passes, so the `code` path is what carries it
- code prefix broken *and* fallback disabled → guard correctly stops matching, the vm error surfaces

**low — the original error was discarded** when the fallback `import()` also failed, so a host where neither path works blamed module resolution for what was really a missing import callback. Both are reported now.

**low — the fallback reintroduced a bare `import(variable)`** into the shipped bundle, which is precisely the static-analysis surface the `new Function` indirection exists to avoid: bundlers analyse it whichever branch runs at runtime, and webpack/ncc turn it into a critical-dependency context module over the whole directory. A second `new Function` is not an option — that is the path that just failed — so it carries `webpackIgnore` and `@vite-ignore` pragmas.

## Verification

- `ingest-files` suite: 10 passed
- Full `ix-cli` suite ×5: 1730 passed each; the 3 failures are the `read-containment` symlink tests that need Windows Developer Mode locally and pass on CI's `windows-2022` runner

### DIFF
--- a/ix-cli/src/cli/__tests__/ingest-files.test.ts
+++ b/ix-cli/src/cli/__tests__/ingest-files.test.ts
@@ -1,4 +1,4 @@
-import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { createServer, type Server, type ServerResponse } from "node:http";
 import type { AddressInfo } from "node:net";
 import { mkdirSync, mkdtempSync, realpathSync, rmSync, writeFileSync } from "node:fs";
@@ -147,11 +147,20 @@ class FakeBackend {
       // and fires against a closed server.
       if (refused) return send(500, { error: "500: transaction begin timeout" });
       this.rev += patches.length || 1;
+      // `status` included, because `PatchCommitResult` declares it required and
+      // the ingest path branches on it. Serving 200s without it left
+      // `result.status` undefined everywhere, so this fake could never produce
+      // an `Idempotent` or `BaseRevMismatch` answer -- the two branches that
+      // decide whether a patch lands in `patchesApplied` or `commitErrors`, and
+      // so whether the mtime baseline is written at all. A regression that
+      // flipped the applied test from "not BaseRevMismatch" to `=== "Ok"` would
+      // have counted zero patches applied against the real backend while every
+      // test here stayed green.
       send(
         200,
         path === "/v1/patches/bulk"
-          ? { rev: this.rev, applied: patches.length }
-          : { rev: this.rev },
+          ? { rev: this.rev, applied: patches.length, status: "Ok" }
+          : { rev: this.rev, status: "Ok" },
       );
       return;
     }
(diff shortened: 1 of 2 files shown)


KageBinary:
Reviewed and merging. The product-code half is the part worth commenting on, and it holds up:

**Matching on `code` rather than the message is right, and I verified the codes are real** rather than taking them from the diff:

```
$ node -e "new vm.Script('import(\"fs\")').runInNewContext()"
throw new ERR_VM_DYNAMIC_IMPORT_CALLBACK_MISSING();
```

and `startsWith("ERR_VM_DYNAMIC_IMPORT_CALLBACK_MISSING")` covers both that and the `_FLAG` variant. The point in the comment is the real one: `/dynamic import callback/i` was a substring of English prose, so an upstream reword would have silently stopped matching and a vm-hosted consumer would have regressed to unloadable with nothing pointing at the guard. Keeping the message test as a fallback for a code-less error is the right shape.

The `webpackIgnore` / `@vite-ignore` pragmas are load-bearing exactly as the comment says — a bare `import(variable)` is the static-analysis surface the `new Function` indirection exists to avoid, and reintroducing it in the fallback would hand webpack/ncc a critical-dependency context module over the whole directory. Worth having written down, since it looks removable.

Reporting both errors on double failure is a small thing that saves a real debugging session: a host where neither path works previously blamed module resolution for a missing import callback.

On the flake — the measured table is the convincing part (1 of 4 under the full suite, 3 of 3 under load, 0 of 5 after). A 5s default sized for unit tests, on a test that runs a real worker pool and an HTTP round trip per commit over 30 files, is a CI leg's business rather than a code fault. Local: `ingest-files.test.ts` 10 passed.

All 25 checks green. Landing this before I cut the CLI release so the fix rides along.

Issue #611, opened 2026-09-05
install-skill.sh: label install output with the harness id and print a per-host summary


Alot1z:
From a hermetic demo run of the merged installer (recorded transcript + screen capture, 2026-09-05). A working implementation exists and can be contributed as a PR if this sounds right.

## Problem

`scripts/install-skill.sh` prints one line per harness, and on a machine with
several harnesses the lines are distinguishable only by a substring buried
mid-path:

```
would install: /home/user/.claude/skills/ix
would install: /home/user/.agents/skills/ix
would install: /home/user/.codex/skills/ix
would install: /home/user/.cursor/skills-cursor/ix
```

Reading a dry run (or a CI log of one) means parsing paths by eye. The harness
id — the thing a user would actually pass on the command line — appears
nowhere in the human output, though it is in hand in the install loop and in
every `--json` record.

A second gap: at the end of a run, the footer reports only counts
(`Ix skill installed for 3 harness(es).`). There is no compact per-host
accounting of what happened — including **skips and refusals**, which are the
interesting outcomes (a refusal means a foreign skill was protected and the
user should know where).

## Evidence

Recorded in a hermetic demo run of the merged installer (2026-09-05):

- Four near-identical `would install:` lines; the harness ids appear only in
  the `--json` report, never in human output.
- The refusal path prints its message inline during the loop and is easy to
  miss when several hosts install before it.
- `--json` exists and is honest, but nothing in the human output mentions it
  (that part is already fixed on the branch above via `--help` + a dry-run
  footer hint; kept separate here so this issue stays about the two output
  changes).

## Minimal fix (output text only — no behavior, no flags, no exit-code changes)

**1. Label the install/dry-run lines with the id:**

```
would install [claude]: /home/user/.claude/skills/ix
Installed [codex]: /home/user/.codex/skills/ix
```

Cost: the id is already in scope at both `say` sites.

**2. Print a per-host summary table from the existing `DECISIONS` records
before the footer** (no new state — the array already holds
`id/action/dest/via`):

```
Harness   Action     Destination                    Detected via
claude    installed  ~/.claude/skills/ix            path
agents    installed  ~/.agents/skills/ix            config-dir
codex     skip       —                              none
cursor    refused    ~/.cursor/skills-cursor/ix     config-dir
```

Destinations print `~`-folded to keep lines short; the absolute path stays in
`--json`. Skips and refusals appear with their real action so the table is
honest, not promotional.

## Compatibility notes (checked)

[...]

PR #618, opened 2026-09-06
chore(install): label installer output per host and add --help

Alot1z:
## Status: DRAFT — smallest installer-UX polish, re-derived onto current main

This pull request is intentionally a **draft**. Feedback and suggestions are welcome at any stage — each point gets addressed on this branch, folded in or answered. The change below is accurate and complete; the branch may still move before it is marked ready.

Refs #611 — this delivers the id-label half of that request. The per-host summary table is deliberately held for a follow-up so this PR stays smallest-first (see Scope).

## What this ships

Three output-text changes to `scripts/install-skill.sh` plus the two CI assertions that must move with them, grounded at current main `8c0e6b00eabc`:

1. **Id-labeled install lines.** `Installed: <dest>` becomes `Installed [claude]: <dest>` (same for `would install [claude]: <dest>` in previews) — four long look-alike paths are distinguishable at a glance, and the id is already in hand in the loop. No behavior change.
2. **A real `--help`.** Prints the usage block and the valid harness ids from the registry (`claude agents codex cursor`), exits 0. Unknown options keep their exit-1 error and gain `try: bash scripts/install-skill.sh --help`. Argument parsing moves after the registry read so `--help` lists real ids; nothing between the two points reads the flags.
3. **Dry-run footer names the JSON variant.** One added line: `Add --json for a machine-readable report (same shape as ix mcp install).` — an in-band signal that `--json` exists, on a shape-parity claim that is already true.

**CI updated in the same change** (both the posix and windows install-smoke jobs): the `would install:` grep assertions become `would install [claude]:`, and each job additionally asserts the new footer line and the `--help` ids line. The old assertions break without this — verified, not assumed (see Validation).

## Deliberately not in this PR

- The per-host summary table (the second half of #611): larger output change, deserves its own review; happy to follow with it.
- No `--prefix`/`--root` flag, no progress bars, no color, no change to guard order, exit codes, or the JSON shape — the audit verified those as sound.

## Scope

- `scripts/install-skill.sh` and `.github/workflows/ci.yml` only. No product code, no README, no registry change.

## Validation

- Hermetic `HOME`-redirected runs on this machine: `--dry-run` shows the four labeled targets and the new footer; `--dry-run --json` parses and carries the same `id/action/dest/detectedVia` shape as before; the foreign-skill refusal still exits 1 with the destination untouched while other hosts install; `--force` bypasses the ix-ownership guard for any existing target (its documented purpose is overwriting a same-name foreign skill), while a bare re-run overwrites a previous Ix install and still refuses a foreign skill.
- CI-assertion simulation at the CI's `/tmp/ix-harness-home` fixture: every new grep passes against the actual job output, and the old `would install:` assertion fails against it — confirming the CI lines must move in the same commit, which they do here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -373,7 +373,9 @@ jobs:
 
           install_out=$(bash scripts/install-skill.sh --dry-run)
           printf '%s\n' "$install_out"
-          grep -Fq 'would install: /tmp/ix-harness-home/.claude/skills/ix' <<<"$install_out"
+          grep -Fq 'would install [claude]: /tmp/ix-harness-home/.claude/skills/ix' <<<"$install_out"
+          grep -Fq 'Add --json for a machine-readable report' <<<"$install_out"
+          bash scripts/install-skill.sh --help | grep -Fq 'Valid harness ids:'
           test ! -e "$HOME/.claude/skills/ix"
 
           shell_json=$(bash scripts/install-skill.sh --dry-run --json)
@@ -479,7 +481,9 @@ jobs:
           install_out=$(bash scripts/install-skill.sh --dry-run)
           printf '%s\n' "$install_out"
           dest="$HOME/.claude/skills/ix"
-          grep -Fq "would install: $dest" <<<"$install_out"
+          grep -Fq "would install [claude]: $dest" <<<"$install_out"
+          grep -Fq 'Add --json for a machine-readable report' <<<"$install_out"
+          bash scripts/install-skill.sh --help | grep -Fq 'Valid harness ids:'
           test ! -e "$dest"
 
           shell_json=$(bash scripts/install-skill.sh --dry-run --json)
--- a/scripts/install-skill.sh
+++ b/scripts/install-skill.sh
@@ -36,8 +36,6 @@ ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 SRC="$ROOT/skills/ix"
 [ -f "$SRC/SKILL.md" ] || { echo "error: $SRC/SKILL.md not found" >&2; exit 1; }
 
-# Argument parsing runs after the harness registry is read (below) so --help
-# can list the real registry ids; nothing between here and the read uses flags.
 
 # A machine-readable report replaces the human output: one object per harness
 # with the action this run takes (would-install | would-refuse | installed |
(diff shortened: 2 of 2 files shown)


KageBinary:
Reviewed the draft. The happy path is good — with node present, `--help` prints usage and the real ids (`claude agents codex cursor`), and an unknown option now points at it:

```
$ bash scripts/install-skill.sh --bogus
error: unknown option --bogus
       try: bash scripts/install-skill.sh --help
```

The id-labelled install lines are a clear win too; four look-alike home-dir paths really are hard to tell apart.

## One finding, on the move that makes `--help` possible

The PR's own comment states the condition it depends on:

> *Argument parsing runs after the harness registry is read (below) so `--help` can list the real registry ids; **nothing between here and the read uses flags**.*

Nothing between them uses flags — that part is true. But two things between them can **exit**, and they now run before any flag is looked at:

```sh
if ! command -v node >/dev/null 2>&1; then
  echo "error: node is required to read the harness registry ..." >&2; exit 1
fi
HELPER_OUT="$(node "$HELPER" --probe)" || { ...; exit 1; }
```

So `--help` now requires node and a successful registry probe. Measured with node off `PATH`:

```
$ bash scripts/install-skill.sh --help
error: node is required to read the harness registry (ix-cli/src/mcp/hosts.ts)
exit=1
```

and the same for `--bogus`, which reports the node error instead of the unknown option.

To be fair about the baseline: this is **not a regression**, since `main` has no `--help` at all (it answers `error: unknown option --help`). It is a flaw in the new feature rather than something the PR broke. But it lands on the one command that should work when nothing else does — someone reaching for `--help` is quite often someone who has not got the toolchain set up yet, and telling them to install node before they can read usage inverts that.

There is a second, smaller cost: `--help` now spawns a subprocess that probes the filesystem for installed harnesses. A help request should not need to scan the machine.

## Suggestion

A pre-scan for `--help`/`-h` before the node check, printing the static usage block, and appending the ids only when the registry happens to be readable. Something like: emit the four option lines unconditionally, then `Valid harness ids: …` if `IDS` is populated, else a fixed fallback or nothing at all. That keeps the real-ids benefit you are after on every machine that can produce them, without making usage itself conditional on a toolchain.

Either way, worth deciding deliberately rather than leaving it to the ordering — the current comment asserts the weaker property (`uses flags`) than the one that actually matters here (`can exit`).

KageBinary:
Pushed the fix for the `--help` finding to this branch (`9d06d9b`), since maintainer edits are enabled and this had been sitting a day. The feature this PR is actually for — the id-labelled output — is untouched and still correct.

**What I changed and why.** The stated justification for moving argument parsing after the registry read was that "nothing between here and the read uses flags". Nothing does — but two things there `exit 1`:

```sh
if ! command -v node ...; then echo "error: node is required ..."; exit 1; fi
HELPER_OUT="$(node "$HELPER" --probe)" || { ...; exit 1; }
```

so the guard was asserting the weaker property. Parsing moves back above the registry read, and the real ids — worth having, and the whole point of your change — are fetched separately inside `usage()` and treated as an enrichment: a missing node, a broken helper or an empty probe all still leave `--help` printing usage and exiting 0. `|| true` there because the script runs under `set -euo pipefail`.

Verified in both environments rather than reasoned about:

```
node absent    --help  usage, exit 0              --bogus  "unknown option", exit 1
node present   --help  usage + real ids, exit 0   --bogus  "unknown option", exit 1
```

That second column is the half I flagged and did not want to leave: `--bogus` was reporting the toolchain instead of the option.

**Your CI assertions still hold**, run locally against the new arrangement:

```
PASS  would install [claude]:
PASS  Add --json for a machine-readable report
PASS  --help | grep -Fq 'Valid harness ids:'
PASS  unknown harness id still exits 1
```

One process note, my error rather than yours: I first pushed to a mistyped branch name and created a stray `chore/installer-labels-and-help` on the upstream repo. It is deleted; nothing else was touched.

Marking ready and merging once CI is green — say the word if you would rather take it from here instead.

PR #621, opened 2026-09-06
fix(ingest): widen the vm dynamic-import guard, and decouple the harness teardown

josephismikhail:
Follow-up to #613, which merged at its first commit — so the review fixes below are not on `main`.

**The first one matters most: `main` currently carries a guard that is narrower than what it replaced.**

## The regression now on main

#613 changed the loader's vm fallback from a message match to an `err.code` match, described as hardening. Written as a **ternary**, it made the two checks mutually exclusive — the message test is reachable only when `code` is absent:

```ts
const isVmCallbackMissing = code
  ? code.startsWith("ERR_VM_DYNAMIC_IMPORT_CALLBACK_MISSING")
  : /dynamic import callback/i.test(String(err));
```

So any layer that rethrows the vm failure with its own code defeats it. That is not hypothetical — Vite's module runner (`reviveInvokeError`) rethrows with `code: 'ERR_LOAD_URL'` and the message preserved, which is exactly the host this fallback exists for. On such a host the wrapped error has a code, fails the prefix test, never reaches the message test, and `ingestFiles` is unloadable.

It is now `||`: either signal, neither excluding the other.

## Also fixed

**`code.startsWith` assumed a string.** `NodeJS.ErrnoException` types it that way, but numeric codes are common (libuv, zlib, OpenSSL). A truthy non-string code threw a `TypeError` from *inside* the catch block, replacing the real load failure with an unrelated one and discarding the original — the precise failure the rest of that block exists to prevent.

**The harness teardown steps still shared fate.** `backend.stop()` ran first inside the `finally`, so a rejection or hook timeout there skipped the env restore and both `rmSync` calls — the same leak the `finally` was added to fix, moved one level in. Env is restored first (it is the only part that leaks out of the file), `stop()` is optional-chained because a `beforeEach` that throws before assigning `backend` would otherwise bury the real failure in a `TypeError`, and the temp trees are removed in a nested `finally`.

## Tests

The predicate is extracted as `isVmDynamicImportUnavailable` and tested. Every error it is given is **derived from a real failure** rather than written by hand:

- the genuine vm failure, from running the loader's own `new Function` indirection under vitest
- the wrapped case, by re-throwing that real error the way Vite's module runner does
- the negative case, from an actual failed import of a missing module

A hand-built `Error` with a plausible message and code would only prove the predicate matches the shape its author already had in mind — which is the shape it was written against.

Mutation-validated: the ternary form fails the wrapped case; dropping the `typeof` guard fails the numeric case.

14 tests pass across the two files.

## Not in this PR, deliberately

The #613 review also found that the Ix#571 **cutoff drain lacks the `BaseRevMismatch` guard** its two sibling commit sites carry, so a drain answering "200, wrote nothing" is counted as applied, `commitErrors` stays 0, and the mtime baseline is written for files the graph never received — silent until someone notices missing symbols, recoverable only with `--force`.

That is a real pre-existing data-loss bug and it gets its own PR. I have the fix, but three attempts at a test were each measured and each shown to exercise the wrong path:

| Attempt | Why it measured nothing |
|---|---|
| poison files to trip the cutoff | the cutoff holds the *failed* patches too, so the drain carries the poison and is refused |
| refuse by request count | the cutoff trips on the fifth failure and the drain follows immediately, so any threshold high enough to produce five failures still refuses the drain |
| refuse by request kind | the run goes fatal before it returns a summary |

Zero accepted bulks in the whole run, measured. Shipping the fix with a test that passes either way would be worse than shipping neither.

### DIFF
--- a/ix-cli/src/cli/__tests__/ingest-files.test.ts
+++ b/ix-cli/src/cli/__tests__/ingest-files.test.ts
@@ -308,13 +308,22 @@ describe("ingestFiles against a fake backend", () => {
       // `saved` is shared across the describe, so the next `beforeEach`
       // snapshotted those polluted values and the file's final restore wrote
       // the fixture `HOME` back into the process.
-      await backend.stop();
+      // And the steps do not share fate. Env first, because it is the one that
+      // leaks OUT of this file: a `stop()` that rejects or outruns the hook
+      // timeout would otherwise skip it and leave the whole process pointed at
+      // a deleted fixture home. `backend` is optional-chained because a
+      // `beforeEach` that throws in `mkdtempSync` leaves it unassigned, and a
+      // TypeError here would bury that failure.
       for (const [k, v] of Object.entries(saved)) {
         if (v === undefined) delete process.env[k];
         else process.env[k] = v;
       }
-      rmSync(home, { recursive: true, force: true });
-      rmSync(repo, { recursive: true, force: true });
+      try {
+        await backend?.stop();
+      } finally {
+        rmSync(home, { recursive: true, force: true });
+        rmSync(repo, { recursive: true, force: true });
+      }
     }
   });
 
--- a/ix-cli/src/cli/__tests__/ingestion-loader.test.ts
+++ b/ix-cli/src/cli/__tests__/ingestion-loader.test.ts
@@ -1,0 +1,86 @@
+import { describe, expect, it } from "vitest";
+
+import { isVmDynamicImportUnavailable } from "../commands/ingestion-loader.js";
+
+/**
+ * The loader reaches `ingestFiles` through `new Function("return import(...)")`
+ * so the specifier stays invisible to bundlers. That indirection cannot run in
+ * a `vm` context without an `importModuleDynamically` hook -- vitest's
+ * vite-node being the one that matters -- so there is a fallback, and this
+ * predicate is the gate on it. Get the gate wrong in either direction and the
+ * damage is quiet: too narrow and `ingestFiles` is unloadable under vite-node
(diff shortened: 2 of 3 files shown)


KageBinary:
Both findings confirmed, and the first one is mine — I reviewed #613, checked that the error codes were real, and did not notice the ternary had made the two signals mutually exclusive. Verified against `main`'s exact expression rather than by reading:

```
  wrapped code + real message    old(main)=false   new(#621)=true
  numeric code + real message    old(main)=THREW   new(#621)=true
  raw node vm error              old(main)=true    new(#621)=true
```

**The narrowing is real.** Any layer that rethrows the vm failure with its own code takes the `code` branch, fails the prefix test, and never reaches the message test — so `main` today answers `false` for a host the fallback exists to serve. That is strictly narrower than the message-only match it replaced, which would have caught it. `||` is the right shape; neither signal should exclude the other.

**The numeric-code path is worse than "returns false".** `main` does not `String()` the code, so:

```
TypeError: code.startsWith is not a function
```

thrown from *inside* the catch block, replacing the real load failure with an unrelated one and discarding the original. `NodeJS.ErrnoException` types `code` as a string, but libuv/zlib/OpenSSL hand back numbers, so the type is a claim about the declaration rather than about the value. `typeof code === "string"` is the fix.

Extracting `isVmDynamicImportUnavailable` is the right call too — this is exactly the sort of predicate whose accepted shapes are not guessable, and deriving the test's fixtures from a real failure rather than writing plausible ones by hand is what would have caught the ternary in the first place.

Local: `tsc --noEmit` clean, `ingestion-loader.test.ts` + `ingest-files.test.ts` 14 passed. All checks green. Merging.

PR #628, opened 2026-09-06
fix(cli): support path disambiguation in graph queries

### LINKED ISSUE #627: bug: five graph commands recommend an unsupported --path option
## Description

`impact`, `callers`, `callees`, `imports`, and `imported-by` recommend `--path` when a symbol is ambiguous, but reject that option. The shared resolver already supports path filtering; these five command registrations do not expose or forward it.

## Steps to reproduce

With two indexed functions named `Duplicate`, one in `src/first.ts` and one in `src/second.ts`, run each of the following:

```sh
ix impact Duplicate --format json
ix callers Duplicate --format json
ix callees Duplicate --format json
ix imports Duplicate --format json
ix imported-by Duplicate --format json
```

Follow the ambiguity diagnostic by adding `--path src/second.ts` to each command.

## Expected behavior

The advertised filter is accepted and the command resolves only the matching candidate. A path matching no candidate must not fall back to an entity in another file.

## Actual behavior

The ambiguity diagnostic says `Use --pick <n> or --path to disambiguate.`, but following it produces `error: unknown option '--path'` instead of a query result.

## CLI output

```text
Use --pick <n> or --path to disambiguate.
error: unknown option '--path'
```

## Environment

Reproduced on CLI v0.10.6 and v0.10.8 and in command-registration regression tests against main at `1ac30fd`, on macOS arm64 with Node.js 26.8.1. The option rejection itself occurs before backend access. The synthetic regression tests mock search results, not command parsing or target resolution.

## Acceptance criteria

- All five commands accept and forward `--path` to the existing resolver.
- Only the matching candidate is expanded; an unmatched path remains an unresolved error.
- Unfiltered ambiguity retains its existing structured output and exit status.
- Help/flag documentation and regression tests cover the change.

Hiro-Chiba:
Fixes #627

## Summary

Make the existing `--path` ambiguity guidance actionable for `impact`, `callers`, `callees`, `imports`, and `imported-by`.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Register `--path <path>` on all five commands and forward it to the existing shared resolver.
- Document the substring filter in the flag reference.
- Add fifteen regression tests using synthetic same-name candidates. Tests cover selecting only the matching candidate, refusing an unmatched path, and preserving unfiltered ambiguity for each command.

The resolver algorithm, existing exit statuses, backend API, and MCP tool schemas are unchanged. This PR does not address architecture-map language coverage or reverse-proxy configuration.

## Validation

- The original three-command fix had six failing tests before the fix. The follow-up extends the same regression coverage to imports/imported-by; all fifteen tests now pass.
- CLI full suite: 97 files passed, 1,797 tests passed and 3 skipped. Parser smoke passed.
- Core ingestion full suite: 38 files passed, 389 tests passed and 1 skipped on rerun. The initial concurrent run failed the pre-existing CUDA adversarial-input timing assertion; no core code was changed.
- Build, typecheck, lint (0 errors, 41 existing warnings), and knip passed.
- Documentation parity passed: 54 commands and 201 flags.
- Live-backend checks confirmed all five patched commands accept a matching path and return success, while a missing path returns structured `unresolved_target` JSON and exit 1.
- `git diff --check` passed.

The live checks verify this specific fix, not universal repository stability. The regression fixtures and public report contain only synthetic names and paths.

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

Release versioning is left to the maintainer. No backend or Compose changes are included.

### DIFF
--- a/ix-cli/src/cli/__tests__/path-disambiguation.test.ts
+++ b/ix-cli/src/cli/__tests__/path-disambiguation.test.ts
@@ -1,0 +1,80 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { Command } from "commander";
+import { registerCallersCommand } from "../commands/callers.js";
+import { registerImpactCommand } from "../commands/impact.js";
+
+const { search, expand } = vi.hoisted(() => ({ search: vi.fn(), expand: vi.fn() }));
+
+vi.mock("../../client/api.js", () => ({
+  IxClient: class {
+    async workspaceSystem() { return { systemId: null }; }
+    async search(...args: unknown[]) { return search(...args); }
+    async expand(...args: unknown[]) { return expand(...args); }
+  },
+}));
+
+vi.mock("../config.js", async (importOriginal) => ({
+  ...await importOriginal<typeof import("../config.js")>(),
+  readStitchScope: () => undefined,
+  writeStitchScope: vi.fn(),
+}));
+
+vi.mock("../hierarchy.js", async (importOriginal) => ({
+  ...await importOriginal<typeof import("../hierarchy.js")>(),
+  getSystemPath: async () => [],
+  bucketByHierarchy: async () => [],
+}));
+
+async function run(command: string, args: string[]) {
+  const program = new Command().name("ix").exitOverride();
+  registerCallersCommand(program);
+  registerImpactCommand(program);
+  const output: string[] = [];
+  vi.spyOn(console, "log").mockImplementation((...parts) => { output.push(parts.join(" ")); });
+  await program.parseAsync([command, "Duplicate", ...args, "--format", "json"], { from: "user" });
+  return JSON.parse(output.join("\n"));
+}
+
+describe.each(["impact", "callers", "callees"])("%s path disambiguation", (command) => {
+  let savedExitCode: typeof process.exitCode;
+
+  beforeEach(() => {
+    savedExitCode = process.exitCode;
+    process.exitCode = undefined;
+    search.mockReset().mockResolvedValue([
+      { id: "first-id", kind: "function", name: "Duplicate", provenance: { sourceUri: "src/first.ts" } },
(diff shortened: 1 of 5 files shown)


KageBinary:
Reviewed the draft. The change is correct and the tests are real; one note about wording that turned into its own PR.

**It is pure wiring of a capability that already exists**, which is the right shape for this. `resolveEntityFull` already accepts `opts.path` and applies a hard filter; these five commands were simply the ones not offering it. Every one of the five does the identical two-line change (register the option, add `path: opts.path` to `resolveOpts`), so there is no per-command behaviour to get wrong.

**Tests are non-vacuous** — I checked rather than counting them. Dropping the forwarding in `impact.ts` alone turns exactly that command's two cases red and leaves the other thirteen green:

```
× selects only the candidate matching --path
× does not silently select a candidate outside --path
  Tests  2 failed | 13 passed (15)
```

`tsc --noEmit` clean, 15/15 pass locally.

**On the wording, and this is in your favour:** "Filter target entity by file path (substring match)" is the accurate description — the resolver hard-filters and returns not-found when nothing survives. What I found while checking that you matched the existing flags is that **nine commands already had `--path` and describe it as "Prefer symbols from files matching this path substring"**, which is wrong for the same code path. A preference cannot fail; this can, and since #547/#559 it also exits non-zero.

So your five would have been the accurate ones and the older nine the misleading ones. Rather than ask you to match the wrong text, I have opened #635 to fix those nine. Nothing for you to change here — worth knowing only so the two PRs do not look like they disagree.

One thing I did not verify, since it needs a graph I do not have handy: the PR says live checks confirmed a missing path returns structured `unresolved_target` JSON and exit 1. That matches what `resolveEntityFull` does on an empty filter result, so I have no reason to doubt it.

Hiro-Chiba:
Marked as ready for review.

PR #632, opened 2026-09-06
fix(rank): scope candidates by path before the fetch limit

### LINKED ISSUE #631: [bug] rank applies --path after truncating the candidate list
## Description

`ix rank --path` can report no matching entities even when the requested class exists and can be listed and expanded. The CLI fetches the first 2,000 entities of the kind without a path condition and only then applies the path filter. Matching entities outside that initial set never reach ranking.

## Steps to reproduce

With more than 2,000 indexed classes, choose a class outside the first unscoped `/v1/list` result set. Using generic example names:

```sh
ix inventory --kind class --path src/target.ts --format json
ix rank --by members --kind class --path src/target.ts --format json
```

The first command can find the class while rank returns no matches. Reproduced with three existing classes on a live backend. Their source and paths are not included here.

A deterministic regression fixture models 2,000 unrelated classes followed by `Wanted` in `src/target.ts`, with the list implementation filtering by `scope` before its limit. It reproduces the command failure without requiring a particular database ordering.

## Expected behavior

Apply the path inclusion condition before the candidate limit so unrelated files cannot crowd the requested path out of the ranking input.

## Actual behavior

Rank exits 0 with `results: []`, `evaluated: 0`, and a diagnostic saying no entities matched the path. Directly listing the same kind with the same workspace scope and a path `scope` finds the target. Its member expansion also succeeds.

## Cause and scope

`registerRankCommand` calls `listByKind(kind, { limit: 2000, ...activeReadScope() })` without `scope`. The existing `/v1/list` API already supports that field and `inventory --path` uses it.

This report concerns path filtering before truncation. It does not claim that the existing 2,000-candidate ranking limit provides an exhaustive global ranking, or that exclusions are server-side.

## Environment

- macOS arm64, Node.js 26.8.1
- Official CLI 0.10.8, backend 1.0.30
- Reproduced in a command regression test based on main `105cd36`

## Acceptance criteria

- Forward a nonempty `--path` as the existing list API's scope while preserving workspace/system scope.
- Keep existing local inclusion/exclusion behavior and scoring.
- A target after 2,000 unrelated candidates must be returned by the scoped command test.
- Preserve the existing candidate bound and avoid adding an unbounded graph scan.

Hiro-Chiba:
Fixes #631

## Summary
Apply `rank --path` through the existing list API before the candidate limit, so unrelated files cannot crowd a requested path out of the ranking input.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Forward a nonempty path as `scope` to `listByKind`, preserving the active workspace/system scope and the 2,000-candidate limit.
- Retain existing client-side path/kind exclusions and scoring.
- Add three command tests, including a target after 2,000 unrelated classes.

This is a scoped retrieval fix, not exhaustive global ranking or pagination. More than 2,000 matches within the requested scope can still be truncated; that existing bound is unchanged.

## Validation
- The new positive regression fails on main before the fix and passes afterward.
- Targeted rank tests: 21 passed.
- Full CLI suite: 97 files passed, 1,785 tests passed and 3 skipped; parser smoke passed.
- Build, typecheck, lint (0 errors, 41 existing warnings), knip, documentation parity, and git diff checks passed.
- Live backend 1.0.30: all three previously missed classes were returned by the scoped rank query, with member counts 3, 2, and 4. The released 0.10.8 CLI had returned empty rankings for those paths.

All public examples and regression fixtures use generic names. No private source or paths are included.

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

Release versioning is left to the maintainer. No backend or Compose changes.

### DIFF
--- a/ix-cli/src/cli/__tests__/rank-path-scope.test.ts
+++ b/ix-cli/src/cli/__tests__/rank-path-scope.test.ts
@@ -1,0 +1,58 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { Command } from "commander";
+import { registerRankCommand } from "../commands/rank.js";
+
+const { listByKind, expand } = vi.hoisted(() => ({ listByKind: vi.fn(), expand: vi.fn() }));
+vi.mock("../../client/api.js", () => ({
+  IxClient: class {
+    async listByKind(...args: unknown[]) { return listByKind(...args); }
+    async expand(...args: unknown[]) { return expand(...args); }
+  },
+}));
+vi.mock("../resolve.js", () => ({
+  ensureReadScope: async () => {},
+  activeReadScope: () => ({ workspaceId: "test-workspace" }),
+}));
+
+async function run(args: string[]) {
+  const program = new Command().name("ix").exitOverride();
+  registerRankCommand(program);
+  const output: string[] = [];
+  vi.spyOn(console, "log").mockImplementation((value) => { output.push(String(value)); });
+  await program.parseAsync(["rank", "--by", "members", "--kind", "class", "--format", "json", ...args], { from: "user" });
+  return JSON.parse(output.join("\n"));
+}
+
+describe("rank path filtering before the candidate limit", () => {
+  beforeEach(() => {
+    const nodes = [
+      ...Array.from({ length: 2000 }, (_, i) => ({ id: `other-${i}`, name: `Other${i}`, kind: "class", provenance: { sourceUri: "src/other.ts" } })),
+      { id: "wanted", name: "Wanted", kind: "class", provenance: { sourceUri: "src/target.ts" } },
+    ];
+    listByKind.mockReset().mockImplementation(async (_kind, opts) =>
+      nodes.filter(n => !opts.scope || n.provenance.sourceUri.includes(opts.scope)).slice(0, opts.limit),
+    );
+    expand.mockReset().mockResolvedValue({ nodes: [{ id: "member" }], edges: [] });
+  });
+  afterEach(() => vi.restoreAllMocks());
+
+  it("ranks a matching entity beyond the first 2000 unfiltered candidates", async () => {
+    const output = await run(["--path", "src/target.ts"]);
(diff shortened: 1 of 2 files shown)


KageBinary:
Reviewed the draft. Correct, and the one-line shape is the right fix for the bug — but I checked the two things that could have quietly broken it, since a one-liner is where those hide.

**The spread order is safe.** `scope` is set before `...activeReadScope()`, so if that returned a `scope` key it would silently clobber the fix. It does not — it returns `{ workspaceId?, systemId? }` only, and `listByKind` accepts `scope` as a separate field. (It actually spreads `cwd` and `stitchChecked` too, since `activeScope()` hands back the whole `_scopeCache` while its declared type hides them. Harmless — `listByKind` reads four named fields — and pre-existing, not yours.)

**The server-side filter really is the same predicate.** `/v1/list` threads `scope` to `scopeNodeFilter`, which is

```
AND CONTAINS(n.provenance.source_uri, @scope)
```

so it is a substring match on `source_uri` — exactly what the client-side `getSourceUri(node).includes(includePath)` does. Both are **case-sensitive**, so pushing the filter down does not change which entities match, only when they are dropped. That is the property that makes this safe, and it is worth stating because it is not automatic: `resolveEntityFull`'s `--path` lowercases via `normalizeForPathMatch` and would NOT have matched the backend. `rank` happens to be the command where the two agree.

**Tests are non-vacuous.** Reverting the push-down turns exactly the case that motivated the PR red and leaves the other two green:

```
× ranks a matching entity beyond the first 2000 unfiltered candidates
  Tests  1 failed | 2 passed (3)
```

`tsc --noEmit` clean, 3/3 pass locally.

The scoping note in the description is honest and worth keeping: this narrows *what* fills the 2,000 slots, it does not remove the bound, so more than 2,000 matches inside the requested path can still truncate.

Hiro-Chiba:
Marked as ready for review.

PR #634, opened 2026-09-06
fix(trace): honor depth and node budgets in path searches

### LINKED ISSUE #633: [bug] trace --to ignores requested depth and node budgets
## Description

`ix trace <from> --to <to>` does not honor the depth or node budgets accepted by the command. A two-edge route is returned even with `--depth 0`, `--depth 1`, or a node cap smaller than the route.

## Steps to reproduce

For an indexed chain `A -> B -> C`, with no direct connection between A and C:

```sh
ix trace A --to C --kind calls --depth 1 --format json
ix trace A --to C --kind calls --depth 0 --format json
ix trace A --to C --kind calls --cap 1 --format json
```

The generic chain above is reproduced in command-registration tests with a synthetic graph. On a live backend, the equivalent two-edge file-to-class-to-method `CONTAINS` chain also returned all three nodes for depths 0, 1, and 2 with cap 1. Exact IDs are omitted.

## Expected behavior

Depth should count graph edges consistently with traversal mode. A depth of one must not return a two-edge route. The path search should also honor the supplied node cap and explain that a bounded miss does not establish that no route exists outside those limits.

## Actual behavior / CLI output

All live combinations above exited 0 with a three-node route and:

```json
{"summary":{"path_length":3}}
```

The `--to` branch passes `maxDepth + 7` to `findPath` and does not pass `maxNodes` at all. `findPath` also compares a path's node count rather than its edge count to the depth bound.

## Environment

- macOS arm64, Node.js 26.8.1
- Official CLI 0.10.8, backend 1.0.30
- Reproduced in command regression tests based on main `105cd36`

## Acceptance criteria

- Forward the requested depth and node budget into path search.
- Enforce depth in edges and a shared unique-node budget for the bidirectional path search, with the source counted once.
- Cover zero budgets, exact boundaries, self paths, and the unrestricted path behavior.
- Qualify bounded misses in text, JSON, and LLM output without changing existing command exit statuses.
- Document the path-mode budget semantics.

Hiro-Chiba:
Fixes #633

## Summary
Honor explicit depth and node budgets when `trace --to` searches for a route.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Test
- [ ] CI

## Changes
- Remove the hidden `+ 7` depth allowance and count depth in graph edges.
- Forward `--cap` to the bidirectional BFS and count unique discovered nodes, including the source, across that search.
- Handle zero budgets and a zero-edge self path without expanding the graph.
- Qualify bounded misses in text, JSON, and LLM output.
- Document the path-mode budget semantics in help and the flag reference.

Successful output shapes and exit statuses are unchanged. Bounded misses retain the existing `no_path` code but explicitly say the route was not found within the requested limits. This node budget bounds graph traversal, not the size of a single backend expand response. No default traversal limits or backend API changes are added.

## Validation
- Eight of eleven new command-registration tests fail before the fix; all eleven pass after it.
- Targeted path tests: 17 passed, including existing inbound traversal cases.
- Full CLI suite: 97 files passed, 1,793 tests passed and 3 skipped; parser smoke passed.
- Build, typecheck, lint (0 errors, 41 existing warnings), knip, documentation parity, and git diff checks passed.
- Live backend 1.0.30: a known two-edge route is no longer returned at depth 0/1, or with a node cap of 2; depth 2 and a sufficient cap return the original three-node route.
- Synthetic A/B/C tests cover exact depth/cap boundaries, a self path, all three output formats, and an unrestricted successful route.

The regression fixtures contain only generic symbols. No private graph identifiers, paths, or source are included.

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

Release versioning is left to the maintainer. No backend or Compose changes.

### DIFF
--- a/ix-cli/src/cli/__tests__/trace-path-budgets.test.ts
+++ b/ix-cli/src/cli/__tests__/trace-path-budgets.test.ts
@@ -1,0 +1,77 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { Command } from "commander";
+import { registerTraceCommand } from "../commands/trace.js";
+
+const expand = vi.hoisted(() => vi.fn());
+vi.mock("../../client/api.js", () => ({
+  IxClient: class { async expand(...args: unknown[]) { return expand(...args); } },
+}));
+vi.mock("../resolve.js", async (importOriginal) => ({
+  ...await importOriginal<typeof import("../resolve.js")>(),
+  resolveFileOrEntityFull: async (_client: unknown, name: string) => ({
+    resolved: true, entity: { id: name, name, kind: "function" },
+  }),
+}));
+
+async function run(args: string[], format = "json", to = "C") {
+  const program = new Command().name("ix").exitOverride();
+  registerTraceCommand(program);
+  const output: string[] = [];
+  vi.spyOn(console, "log").mockImplementation((...parts) => { output.push(parts.join(" ")); });
+  await program.parseAsync(["trace", "A", "--to", to, "--kind", "calls", ...args, "--format", format], { from: "user" });
+  return output.join("\n");
+}
+
+describe("trace --to search budgets", () => {
+  beforeEach(() => {
+    expand.mockReset().mockImplementation(async (id: string, opts: { direction: string }) => {
+      const adjacent: Record<string, string[]> = opts.direction === "out"
+        ? { A: ["B"], B: ["C"] } : { B: ["A"], C: ["B"] };
+      return { nodes: (adjacent[id] ?? []).map(name => ({ id: name, name, kind: "function" })), edges: [] };
+    });
+  });
+  afterEach(() => vi.restoreAllMocks());
+
+  it("does not return a two-edge path at depth one", async () => {
+    const output = JSON.parse(await run(["--depth", "1"]));
+    expect(output.path).toBeNull();
+    expect(expand.mock.calls.map(([id]) => id)).toEqual(["A", "A"]);
+  });
+
+  it("finds a path exactly at the depth boundary", async () => {
+    const output = JSON.parse(await run(["--depth", "2"]));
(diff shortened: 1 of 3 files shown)


KageBinary:
Reviewed the draft. This is the largest of the three and the one with the most room to go wrong; it holds up. Three things I checked rather than took on trust.

**The `+ 7` removal does not narrow the default.** This was my main concern — removing a hidden allowance can turn "found a route" into "no route" for people who never passed a flag. It cannot here:

```ts
const DEFAULT_MAX_DEPTH = Infinity;
const MAX_NODES = Infinity;
```

`Infinity + 7` is `Infinity`, so the allowance only ever applied when someone explicitly passed `--depth` — precisely the case where it was searching seven levels deeper than asked. Removing it changes behaviour only for users who had asked for a bound and silently were not given one.

**Depth now counts edges, and the tests pin it.** `path.length >= maxDepth` → `path.length - 1 >= maxDepth`, matching the new help text ("Cap traversal depth in edges"). Reverting just that turns exactly the two boundary cases red:

```
× does not return a two-edge path at depth one
× finds a path exactly at the depth boundary
  Tests  2 failed | 9 passed (11)
```

**The self-path does not render blank.** `if (fromId === toId) return [{ id: fromId, name: "", kind: "" }]` looked like it would print an unnamed node, but the caller already backfills index 0 from `fromTarget` — the `// Fill in the from-node name (was left blank above)` map — so the empty strings never reach output. Consistent with how the existing code builds the first node.

One design note, not a request. `visited.size >= maxNodes` returns `null` from the whole search rather than stopping expansion and returning the best route found so far. For a shortest-path BFS that is the right call — a partial answer would not be shortest and would be worse than saying so — and the qualified message ("within the requested search limits") is what makes it honest rather than looking like a genuine absence. Worth keeping that message wired through all three formats, which you did.

`tsc --noEmit` clean, 11/11 pass locally.

Hiro-Chiba:
Marked as ready for review.

PR #637, opened 2026-09-06
fix(ingest): count a lost base-rev race on the cutoff drain as an error

josephismikhail:
Found by the reviewer on #613, reading the third commit site against the other two.

## The bug

A backend can answer **200 and still have written nothing**. `BaseRevMismatch` means it read the latest rev outside the transaction and the rev moved before the commit ran, so the patches were not applied.

`onBulkCommitted` and the per-file path both check for it. The cutoff drain added for Ix#571 did not — it counted the whole chunk as applied.

The consequence is worse than a wrong number in the summary. `commitErrors` staying at zero is exactly the condition `persistIngestBaselineIfClean` requires, so the run **writes an mtime baseline for files the graph never received**. Every later incremental map then skips them as unchanged. The symbols are missing, nothing reports it, and only `ix ingest --force` recovers.

The fix is the same guard its two siblings carry.

## The test was the hard part

Three attempts each measured something other than the guard, and each looked reasonable until checked. That is why the harness now records the **response code** of every request and exposes `acceptedBulks()` — checking the premise is what exposed them:

| Attempt | What actually happened |
|---|---|
| poison files to trip the cutoff | the cutoff holds the patches that **failed** as well as the untried ones, so the drain carries the poison and is refused — **zero accepted bulks in the whole run** |
| refuse by request count | the cutoff trips on the fifth failure and the drain follows immediately, so any threshold high enough to produce five failures is still in force when the drain arrives — zero again |
| refuse by request **kind** | works; the sequence is now pinned in the test |

The working sequence, asserted rather than assumed:

```
BULK:30 => 500     the opening bulk, refused
one:1   => 500     x5, refused, which trips the cutoff
BULK:25 => 200     the drain, ACCEPTED — the request under test
```

## The assertion is the user-visible consequence

The same fixture is ingested twice and the only difference is the status the accepted drain answers with:

- `"Ok"` → places 25 patches, run succeeds
- `BaseRevMismatch` → places nothing, so the run must report **30 of 30 missing** and end fatally

Without the guard the second run succeeds identically to the first: exit 0, same error count, 25 files silently absent from the graph. Mutation-validated — with the guard removed the test fails with `expected the ingest to end fatally`, which is the bug stated as plainly as it can be.

## Verification

- `ingest-files` suite: 11 passed
- Full `ix-cli` suite: 1732 passed; the 3 failures are the `read-containment` symlink tests that need Windows Developer Mode locally and pass on CI's `windows-2022` runner

### DIFF
--- a/ix-cli/src/cli/__tests__/ingest-files.test.ts
+++ b/ix-cli/src/cli/__tests__/ingest-files.test.ts
@@ -42,7 +42,7 @@ import { ingestFiles } from "../commands/ingest.js";
 
 /** A backend that answers the endpoints an ingest touches, and records them. */
 class FakeBackend {
-  readonly requests: Array<{ path: string; patches: number }> = [];
+  readonly requests: Array<{ path: string; patches: number; code?: number }> = [];
   /** Paths this fake does not implement. Asserted empty after every test. */
   readonly unknownPaths: string[] = [];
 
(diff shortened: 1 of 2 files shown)


KageBinary:
Verified and merging. I rebased it onto main first — it was ten commits behind and conflicting.

**The bug is real and the consequence chain checks out.** I confirmed each link rather than reading it:

- There were exactly **two** `COMMIT_STATUS_BASE_REV_MISMATCH` guards on main (lines 2206 and 2533); the cutoff drain was genuinely the third site without one.
- `persistIngestBaselineIfClean` opens with `if (!ingestCompletedCleanly(parseErrors, commitErrors)) return false;` — so `commitErrors` staying at zero is exactly what lets the baseline be written.

That is the part worth spelling out: the visible symptom is a wrong number in the summary, but the actual cost is an mtime baseline recorded for files the graph never received. Every later incremental map then skips them as unchanged, so the symbols are simply absent and nothing ever says so. Only `--force` recovers it. That is the same silent-until-someone-notices shape as Ix#620, and it is worse here because the graph is wrong rather than merely under-registered.

**Mutation-checked** — disabling only the new guard (line 2369) turns exactly its test red:

```
× counts a lost base-rev race as an error on the CUTOFF DRAIN too
  Tests  1 failed | 10 passed (11)
```

**Conflict resolution, for the record:** one hunk, in `ingest-files.test.ts`, and it was purely comment wording — both sides describe the same `await backend?.stop()`. I kept main's phrasing ("`backend` is optional-chained") since it names what the code actually does; nothing executable differed. `ingest.ts` merged clean. After rebase: `tsc --noEmit` clean, 11/11 pass, guard count 2 → 3.

The write-up of the three failed test attempts is worth keeping in the history. "The cutoff holds the patches that failed as well as the untried ones, so the drain carries the poison and is refused — zero accepted bulks in the whole run" is exactly the kind of thing that makes a test look right and measure nothing, and `acceptedBulks()` recording response codes is the right answer to it.

PR #649, opened 2026-09-07
fix(ingest): make the code clause of the vm guard load-bearing, and guard the teardown properly

josephismikhail:
Follow-up to #621, which merged at its first commit — these six review findings are not on `main`.

## The one that matters: my `code` clause had zero coverage

#621 widened the vm guard to `code || message`, with tests. But the real vm error carries **both** a matching code and a matching message, so `||` short-circuits on the message and the code branch is never reached.

Mutation-verified: replacing the whole predicate with the message-only test left the file **4/4 green**.

That is exactly the pre-#613 form. So a later "simplification" back to it ships green — and then Node rewords the message, which is the precise hazard the code check was added for, and the fallback silently stops firing with nothing pointing at the guard. The test file claimed the gate was protected "in either direction". It was protected in one.

There is now a case only the code clause can satisfy: the real error, its real code, and the message replaced with something that does not match. Derived from a genuine failure like the others. **Both clauses are mutation-validated now** — deleting either fails a different test.

## The teardown guards, which were in the wrong places

| Finding | Why the previous fix did not work |
|---|---|
| `backend?.` optional chain was inert | the scenario it names is a `beforeEach` that throws before assigning `backend` — but the unguarded `backend.unknownPaths` dereference is two lines **above** it and throws first |
| `home`/`repo` unguarded | `force: true` suppresses ENOENT, **not** the argument-type check: `rmSync(undefined, …)` throws `ERR_INVALID_ARG_TYPE`, raising a second unrelated error and skipping the `repo` removal entirely — leaking the tree in exactly the case cleanup exists for |
| the comment overclaimed | it said the decoupling protects against a `stop()` that "outruns the hook timeout". It does not — on a hook timeout the promise is rejected from outside while the `await` is still pending, so no `finally` runs at all |

The last one is now fixed where it has to be: `stop()` can no longer hang. It calls `closeAllConnections()` before `close()` and races a 2s deadline, because `close()` waits for every open socket and a single stray keep-alive was enough to leave the hook pending. The comment now claims only what it delivers.

## Two smaller ones

- The loader's JSDoc had drifted onto the wrong function — the new block was inserted *after* the existing one instead of replacing it, so both attached to the predicate and `importModule` was left undocumented.
- `(await realVmFailure()) as Error` then `original.message` would crash with an opaque `TypeError` in the scenario the file explicitly anticipates (a host that gains a dynamic-import hook returns `null`). Both sites assert it first.

## Verification

- 16 tests pass across the two files
- Full `ix-cli` suite: 1736 passed; the 3 failures are the `read-containment` symlink tests that need Windows Developer Mode locally and pass on CI's `windows-2022` runner

### DIFF
--- a/ix-cli/src/cli/__tests__/ingest-files.test.ts
+++ b/ix-cli/src/cli/__tests__/ingest-files.test.ts
@@ -150,7 +150,19 @@ class FakeBackend {
   }
 
   async stop(): Promise<void> {
-    if (this.server) await new Promise<void>((resolve) => this.server!.close(() => resolve()));
+    if (!this.server) return;
+    const server = this.server;
+    // `closeAllConnections` first, and a deadline behind it. `close()` waits for
+    // every open socket, so one keep-alive connection the run did not finish
+    // with leaves this pending -- and a hook that never settles takes its
+    // timeout, at which point NEITHER `finally` in the teardown runs and both
+    // temp trees leak anyway. The teardown can only be made safe if this cannot
+    // hang, so the fix belongs here rather than in another `try`.
+    server.closeAllConnections?.();
+    await Promise.race([
+      new Promise<void>((resolve) => server.close(() => resolve())),
+      new Promise<void>((resolve) => setTimeout(resolve, 2000).unref?.()),
+    ]);
   }
 
   private route(url: string, body: string, res: ServerResponse): void {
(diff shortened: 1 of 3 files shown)


josephismikhail:
## The red check is a known flake, not this PR

`Test (windows-2022 · node 22)` failed on `4d8046b` at `parse-pool.test.ts:137` — `b2.ts` returning `null` in `takes a worker that faults while IDLE out of the free list`.

**This PR does not touch `parse-pool.test.ts`.** Its diff is limited to `ingest-files.test.ts`, `ingestion-loader.test.ts` and `ingestion-loader.ts`.

That test waits a flat 120ms for a worker `'error'` event to be delivered before dispatching two parses; when the machine is busy the event arrives late and the second parse reaches an already-dying worker. Measured locally: 0 of 10 runs failed idle, 1 of 8 under saturating CPU load.

Fixed in #651, which waits on the pool's own signal that `onError` has run instead of on a duration. Once that merges this leg should go green with no change here.

PR #651, opened 2026-09-08
test(parse-pool): wait for the idle fault to land, not for 120ms

josephismikhail:
## What

`takes a worker that faults while IDLE out of the free list` is load-flaky. It failed a full-suite run on a busy machine with `b2.ts` returning `null` — **indistinguishable from the Ix#567 bug it guards**, which is the worst way for a flake to present: whoever sees it red next has no reason to suspect the harness.

## Reproduction

The fixture throws 20ms after serving the first task; the test slept a flat 120ms waiting for the `'error'` event before dispatching two parses.

| condition | result |
|---|---|
| idle | 0 of 10 runs failed |
| saturating CPU load | **1 of 8 runs failed** |

Under load the event had not been delivered, so the second parse reached an already-dying worker and resolved `null`.

## Fix

Wait on the condition, not a duration. `onError` increments `respawns` and pushes the replacement in the same synchronous handler, so a non-zero count means the pool has *finished* reacting — including to a fault with no task in flight, which `crashedTasks()` deliberately does not count because no file was lost.

That distinction is precisely why there was nothing to wait on, and why this adds `respawnCount()` rather than reusing a counter. It earns its place outside the test as well: a run that quietly respawned a dozen workers looks identical to a healthy one in every other counter the pool exposes.

## What I am *not* claiming

Post-fix loaded runs were 0 of 8. **That is not strong evidence** — under the observed 1-in-8 rate, zero in eight happens 34% of the time. The argument is structural: the test waits on the pool's own state instead of on an event arriving inside a fixed window.

**Correction (round 4):** an earlier revision of this section claimed the test is *faster* than the sleep it replaces. That was true of the original 20ms fault and stopped being true when round 3 widened it to 250ms to close a second ordering window — the test now waits on a 250ms timer where it previously slept 120ms, so it is somewhat slower, not faster. The wait-on-condition change is still the right one; it is not a speedup.

The same round also has to qualify "removed rather than made less likely": the *first* window (waiting for the fault to land) is genuinely removed, but the ordering dependency cannot be eliminated — the worker has no way to learn its reply was consumed, and a fault raised while a task is in flight never leaves a stale entry in `idle`, which is the whole bug. So the premise needs an idle fault, an idle fault needs a delay, and 250ms only makes the required parent stall implausible.

## The test still catches the bug

With the `idle.splice` in `onError` deleted — Ix#567 restored — the test fails, hanging until the 20s timeout, which is the "promise never settles" symptom exactly. The poll signal increments *after* that splice, so it does not depend on the fix being present and cannot mask its absence.

The other five fixed sleeps in this file wait for a parse to be *dispatched* rather than for an error to be *delivered*, and none has been observed to fail. Left alone rather than rewritten on speculation.

## Verification

- 13 pool tests pass; full suite 1791 (3 known Windows symlink failures, unrelated, pass on CI's `windows-2022`)
- typecheck and lint clean
- mutation-tested as above

### DIFF
--- a/ix-cli/src/cli/__tests__/parse-pool.test.ts
+++ b/ix-cli/src/cli/__tests__/parse-pool.test.ts
@@ -125,7 +125,17 @@ describe("ParsePool", () => {
       try {
         writeFileSync(marker, threadId + '\\n', { flag: 'wx' });
         armed = true;
-      } catch {
+      } catch (err) {
+        // EEXIST is the expected answer: another thread holds the claim.
+        // Anything else -- EPERM or EBUSY from an indexer touching the fresh
+        // mkdtemp directory, ENOSPC -- means the FIRST worker failed to arm,
+        // and a bare catch would swallow that and leave the test to time out
+        // 10s later saying the pool never reacted to a fault nobody ever
+        // scheduled. Accusing the pool of a harness failure is the exact
+        // misdiagnosis this test exists to prevent, so rethrow: the worker
+        // faults, the marker stays empty, and the arming assertion fails
+        // naming the real problem.
+        if (err.code !== 'EEXIST') throw err;
         armed = false;
       }
       if (armed) {
@@ -202,6 +212,12 @@ describe("ParsePool", () => {
     // `respawnCount()` is the pool's own signal that `onError` ran to
     // completion, so this waits on the state the test actually depends on and
     // is done as soon as it holds.
+    // `> 0` is only correct because this is the FIRST death of the run. The
+    // counter is monotonic and never resets, so a second wait written this
+    // way returns immediately -- a silent no-op, which is the same class of
+    // bug this test was fixed for. A later fault must capture a baseline
+    // first: `const before = pool.respawnCount()` then `() => pool
+    // .respawnCount() > before`.
     await waitUntil(
       () => pool.respawnCount() > 0,
       "the idle fault never reached the pool",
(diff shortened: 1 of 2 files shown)


josephismikhail:
## Now reproduced on CI, not just locally

`Test (windows-2022 · node 22)` on PR #649 (`4d8046b`) failed with exactly this:

```
AssertionError: expected [ { filePath: 'a2.ts' }, null ] to deeply equal [ { filePath: 'a2.ts' }, …(1) ]
-   { "filePath": "b2.ts" },
+   null,
 ❯ src/cli/__tests__/parse-pool.test.ts:137:18
```

[run 34179642182](https://github.com/ix-infrastructure/Ix/actions/runs/34179642182/job/101915931370) — same test, same assertion, same line.

**#649 does not touch `parse-pool.test.ts`.** Its diff is `ingest-files.test.ts`, `ingestion-loader.test.ts`, `ingestion-loader.ts`. So this is not that PR's regression — it is this flake, and the windows-2022 runner is loaded enough to hit it.

That upgrades the evidence here in two ways:

1. The failure mode is **real on CI**, not an artifact of my synthetic load. A green history simply means the runner usually delivers the `'error'` inside 120ms.
2. It manifests as `b2.ts → null`, which is the Ix#567 signature. Anyone triaging that red on an unrelated PR would start by suspecting their own change, which is the cost this fix removes.

Merging this first would clear #649's red without touching #649.

PR #658, opened 2026-09-08
fix(readme): unclip the architecture diagram's backend caption

KageBinary:
The architecture diagram shipped in #656 with its "Ix backend" caption cut off
mid-word:

> Runs locally in Docker and persists system **knowledg**

The sentence overflowed its panel and the tail ran underneath the Backend API
card, so the front page of the repo showed a truncated word.

## Fix

Wrapped it onto two lines rather than shortening it — the panel had ~90px of
unused space below the caption and nothing else occupies it:

```
Runs locally in Docker and
persists system knowledge.
```

Redrawn in place at the original baseline (ink top row 829, ink left x=146), so
the first line sits exactly where the old one did. Liberation Sans 15px matches
the original run's metrics to within 1px measured over 51 characters, so the
replacement is visually consistent with the rest of the panel.

The caption now ends at **x=327** against a card edge at **x=501** — 174px of
clearance, so it cannot clip again at this width.

Verified nothing else moved: the only pixels touched are the caption band
(rows 820–850, x 140–498). The card, its border and every other element are
byte-identical.


KageBinary:
Superseded by #672. #665 re-cropped the diagram to 1618×741 but kept the clipped caption, so this fix (drawn on the 1800×1050 canvas) no longer applies. #672 is the identical two-line wrap on the new framing; every pixel outside the caption band is byte-identical to `main`.

PR #660, opened 2026-09-09
fix(text): stream ripgrep output within the result limit

### LINKED ISSUE #659: [bug] ix text fails above 10 MiB of aggregate ripgrep output despite a small result limit
## Description

`ix text` can fail before returning even its default 20 results when ripgrep's aggregate JSON output exceeds 10 MiB. This depends on total matching output, not the number of results requested by the caller. No backend is involved.

## Reproduction

The following standalone fixture uses only generated content. It requires Node.js, ripgrep and an `ix` executable on PATH, and removes its temporary files afterward.

```sh
node --input-type=module <<'JS'
import { mkdtempSync, writeFileSync, rmSync } from 'node:fs';
import { tmpdir } from 'node:os';
import path from 'node:path';
import { execFileSync, spawnSync } from 'node:child_process';
const root = mkdtempSync(path.join(tmpdir(), 'ix-text-repro-'));
try {
  for (let i = 0; i < 128; i++) {
    writeFileSync(path.join(root, `${i}.ts`), `needle ${'x'.repeat(100_000)}\n`);
  }
  const raw = execFileSync('rg', ['--json', '--max-count', '20', '--', 'needle', root], {
    maxBuffer: 32 * 1024 * 1024,
  });
  console.log('rg JSON exceeds 10 MiB:', raw.length > 10 * 1024 * 1024);
  const result = spawnSync('ix', ['text', 'needle', '--root', root, '--format', 'json'], {
    encoding: 'utf8', maxBuffer: 32 * 1024 * 1024,
    env: { ...process.env, IX_MCP_CHILD: '1' },
  });
  console.log({ exit: result.status, stdoutBytes: Buffer.byteLength(result.stdout ?? ''), stderr: result.stderr });
} finally {
  rmSync(root, { recursive: true, force: true });
}
JS
```

Before the fix, the command exits 1, writes no JSON, and prints `Error: ripgrep failed` to stderr. Expected behavior is a valid JSON array containing 20 matches and exit 0.

## Cause and verification

On main `6f72257bb1156c1bb70bd66356ecc7b5914e963d`, `runRipgrep` uses `execFile` with `maxBuffer: 10 * 1024 * 1024`. `rg --max-count 20` limits matches per file, not across the workspace. The command collects all stdout before applying `results.slice(0, limit)`.

The generated fixture was tested through the real command registration and real ripgrep. The regression test failed on the unchanged implementation and passes with streaming retention of the requested matches. Independently, two common search terms in a large local workspace produced `ERR_CHILD_PROCESS_STDIO_MAXBUFFER` with 10,485,760 bytes collected by the same underlying invocation. No private workspace data is needed to reproduce the issue.

Environment: macOS arm64, Node.js 26.8.1, ripgrep 15.2.0. The test does not start or query a backend.

## Resolution criteria

- Return the requested results without buffering all aggregate ripgrep output.
- Preserve workspace/path/language filtering, the pattern argument separator, and existing output formats.
- Preserve no-match success and actual ripgrep failures, including errors after the requested matches have been encountered.
- Include a regression test whose aggregate rg JSON output exceeds 10 MiB.

Hiro-Chiba:
Fixes #659

## Summary

Stream `ix text` ripgrep output and retain only the first requested match records, removing the aggregate 10 MiB stdout-buffer failure.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [x] Test
- [ ] CI

## Changes
- Replace buffered `execFile` with `spawn` and line-by-line JSON consumption. Keep only the first N match records for the existing formatter.
- Drain the remaining output and wait for the child's natural exit. Do not terminate at N, since that could hide later read errors. Successful diagnostic stderr remains compatible with rg debug output.
- Retain up to 64 Ki characters of stderr for existing error rendering. Preserve missing-rg handling, no-match success, and nonzero failure behavior.
- Keep the existing rg arguments, path guard, language filters and JSON/text/LLM formatting. Adapt the argument-injection tests to the existing command runner injection seam.
- Add real-rg overflow regression coverage and stream lifecycle tests. Declare the existing external `rg` prerequisite in Knip's binary allowlist, now detected at the direct spawn call sites.

This does not promise faster scanning or a fixed memory ceiling. Ripgrep still finishes its scan; retained memory depends on the requested matches and their line lengths, plus the current input line and stream buffers. No backend or plugin change is required, and no new command exit-code policy is introduced.

## Validation

- Red/green regression with 128 generated files, each containing a 100,000-character matching line. The fixture verifies aggregate rg JSON exceeds 10 MiB. The original command fails; the patched command returns exactly 20 JSON matches.
- Eleven lifecycle tests cover split UTF-8 records, metadata, retained-result limits, waiting for child close, natural exits 0/1, diagnostic stderr on success, late errors, abnormal termination, missing rg, and text/LLM formatting.
- `npm test`: 104 files passed, 1,850 tests passed and 3 skipped; parser smoke passed. The real-rg regression ran, not skipped.
- `npm run build`, `npm run typecheck`, and `npm run knip` passed. `npm run lint` passed with 0 errors and 41 existing warnings. `git diff --check` passed.
- Read-only CLI checks in a large local workspace: two formerly failing default-limit searches now each return 20 results with exit 0. Also checked a one-result limit, all three output formats, no matches, and invalid regex failure. No private source or identifiers are included here.
- Backend remained stopped. Graph remapping is not needed to exercise this backend-independent command and was not performed.

## Checklist
- [x] Tests pass
- [x] Smoke tests pass

Release version/tagging is left to maintainers. No backend or compose changes.

### DIFF
--- a/ix-cli/knip.json
+++ b/ix-cli/knip.json
@@ -3,5 +3,5 @@
   "entry": ["src/cli/main.ts", "src/index.ts", "test/parser.smoke.mjs"],
   "project": ["src/**/*.ts"],
   "ignore": ["test/fixtures/**", "test/parser.stress.mjs"],
-  "ignoreBinaries": ["ix", "cygpath", "tar", "start", "xdg-open", "mkfifo"]
+  "ignoreBinaries": ["ix", "rg", "cygpath", "tar", "start", "xdg-open", "mkfifo"]
 }
--- a/ix-cli/src/cli/__tests__/text-arg-injection.test.ts
+++ b/ix-cli/src/cli/__tests__/text-arg-injection.test.ts
@@ -1,18 +1,7 @@
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import { Command } from "commander";
 
-const execFile = vi.hoisted(() =>
-  vi.fn((_cmd: string, _args: string[], _opts: unknown, cb: (e: unknown, r: unknown) => void) =>
-    cb(null, { stdout: "", stderr: "" }),
-  ),
-);
-
-// Spread the real module: `config.ts`, reached through this command, imports
-// `execSync` from it.
-vi.mock("node:child_process", async (orig) => ({
-  ...(await orig<Record<string, unknown>>()),
-  execFile,
-}));
+const runRipgrep = vi.fn(async (_args: string[], _limit: number) => ({ stdout: "" }));
 
 import { registerTextCommand } from "../commands/text.js";
 
@@ -25,14 +14,14 @@ import { registerTextCommand } from "../commands/text.js";
  */
 describe("ix text does not let the search term become an rg flag", () => {
   beforeEach(() => {
-    execFile.mockClear();
+    runRipgrep.mockClear();
   });
 
   async function argvFor(term: string, extra: string[] = []): Promise<string[]> {
     const program = new Command();
-    registerTextCommand(program);
+    registerTextCommand(program, runRipgrep);
     await program.parseAsync(["text", ...extra, "--", term], { from: "user" });
-    return execFile.mock.calls[0]?.[1] as string[];
+    return runRipgrep.mock.calls[0]?.[0] as string[];
   }
 
   it.each(["--files", "--pre=/bin/sh", "--file=/etc/passwd", "-l", "--no-ignore"])(
--- a/ix-cli/src/cli/__tests__/text-stream-integration.test.ts
+++ b/ix-cli/src/cli/__tests__/text-stream-integration.test.ts
@@ -1,0 +1,37 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { Command } from "commander";
+import { execFileSync, spawnSync } from "node:child_process";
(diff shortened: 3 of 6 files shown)


KageBinary:
Reviewed and verified locally on the head (2ed5ce2): typecheck, lint (0 errors) and knip clean; the four text test files pass 21/21 including the real-ripgrep 128 x 100 KB reproduction of #659.

Traced the paths I was worried about and found no defect: every readline `line` event is delivered before the child's `close` (including an unterminated final record), a real ENOENT emits `error` before `close` so the not-installed message survives, exit 1 still renders an empty result set, and exit 2 after matches still fails with rg's first stderr line. Windows resolution is unchanged because `execFile` is `spawn` underneath.

One non-blocking follow-up: none of the GitHub-hosted runner images ship ripgrep and `ci.yml` installs none, so `text-stream-integration.test.ts` is skipped on all five legs. The mocked lifecycle tests still cover the streaming logic in CI. Installing ripgrep on the ubuntu leg would make the regression test real there.

Merging.

PR #663, opened 2026-09-10
chore: copyright header on every source file + GitHub Sponsors button

KageBinary:
Tanner, in Slack: every source file in the open source needs a copyright line at the top, using the format already in the repo with `LLC` replaced by `Inc.`. Plus a Sponsor button, since the org has GitHub Sponsorship.

## Copyright headers

The only `Copyright` strings in the repo today were `LICENSE` and `NOTICE`, both reading `Copyright 2026 Ix Infrastructure LLC`. That is the format matched here:

```ts
// Copyright 2026 Ix Infrastructure Inc.

import type { Command } from "commander";
```

- **272 source files** — `.ts`, `.mjs`, `.js`, `.scala`, `.sh`, `.ps1`, `.py`, `.rb`, `.cmd` — each language in its own comment syntax (`//`, `#`, `@REM`).
- **Shebangs stay on line 1.** The header goes after `#!/usr/bin/env bash`, after `@echo off` in `install.cmd` (a bare `REM` on line 1 would echo to the console), and after a Python encoding declaration.
- **`LICENSE` and `NOTICE` move to `INC` too** — otherwise the new per-file headers would name a different entity than the repo's own legal notices.

### What is deliberately excluded

The 11 files under `core-ingestion/test-fixtures/`, `ix-cli/test/fixtures/` and `.github/fixtures/`. The ingestion tests parse these and snapshot the result, including line numbers — a header shifts every symbol down a line and breaks the snapshots, for no benefit. They are test *data*, not Ix source.

## Sponsor button

The org already has a public sponsors listing at `github.com/sponsors/ix-infrastructure`, but **no repo had a `.github/FUNDING.yml`** — that file, not a README edit, is what renders GitHub's Sponsor button. Both are added:

- `.github/FUNDING.yml` → `github: [ix-infrastructure]`, which puts the Sponsor button in the repo header.
- README gets a sponsor badge in the badge row and a `Sponsor` link in both nav rows.

## Verification

| Gate | Result |
|---|---|
| `ix-cli` vitest + parser smoke | 1,839 passed, 2 skipped (102 files) |
| `core-ingestion` vitest | 389 passed, 1 skipped (38 files) |
| `npm run typecheck` | clean |
| `npm run lint` | 0 errors (41 pre-existing warnings) |
| `npm run knip` | clean |
| `npm run format:check` | 168 files flagged — identical count before and after, pre-existing drift |

Shebang integrity was checked mechanically: every file whose first line was a shebang before still has it first.

This is the first of 10 repos — the other 9 OSS repos get the same treatment next.

https://claude.ai/code/session_016WMzUgkHNBXpThNT9c9wWw

## CI gate

Headers decay the moment someone adds a file, so this PR also ships the gate. `.github/workflows/copyright.yml` runs on every PR and push to `main` and fails the build listing each offender:

```
2 source file(s) are missing the copyright header:

  ix-cli/src/cli/brand-new.ts
  scripts/brand-new.sh

Fix them all with:

    python3 .github/scripts/copyright-headers.py --fix
```

The same script carries `--fix`, so a contributor repairs every offender in one command rather than hand-editing files.

It reuses **the exact exclusion rules the headers were applied under** — fixtures, `linguist-generated` paths, build output, dependencies — so the gate agrees with the tree instead of fighting it. Verified by adding a throwaway `.ts`, a throwaway `.sh` and a throwaway *fixture*: it flagged the first two, correctly ignored the fixture, `--fix` repaired both with the shebang left on line 1, and the re-check went green.

No `setup-python` step — `python3` is preinstalled on `ubuntu-latest` and the script is stdlib plus `git ls-files` / `git check-attr`.

**To make this binding, the check needs adding to branch protection** — otherwise it reports but cannot block.


KageBinary:
Refreshed before merging (e987e01): merged `main` into the branch and ran `copyright-headers.py --fix`, which headered the five source files that landed on `main` after this branched (`ix-home.ts`, `vitest.setup.ts`, and the `ix-home`, `text-stream` and `text-stream-integration` tests). The two merge conflicts, `bootstrap.ts` and `commands/text.ts`, were only at the top of the file, where this branch added the header and `main` rewrote imports; both were resolved to `main`'s side and re-headered.

Checked against `main`: every changed file outside FUNDING.yml, the script, the CI step, CONTRIBUTING, LICENSE, NOTICE and README is a pure two-line header insertion. Also reviewed the script itself (shebang, `@echo off` and coding-line placement, CRLF preservation, linguist-generated and fixture exclusions, idempotence) and scanned for BOM-prefixed sources that an insertion at byte 0 would damage: none. The gate is a step in `static`, which `CI Passed` needs, so it binds. CI 31/31.

PR #666, opened 2026-09-13
fix(config): honour IX_HOME for config, the workspace registry and caches

KageBinary:
## The bug

`~/.ix` is resolved in nine places across three idioms:

| How | Where |
|---|---|
| `process.env.IX_HOME` bound at import | `backend-status.ts`, `backend-version.ts`, `commands/docker.ts`, `commands/upgrade.ts`, `commands/view.ts` |
| `process.env.IX_HOME` read per call | `commands/doctor.ts`, `commands/context.ts` |
| **IX_HOME ignored — bare `homedir()`** | **`config.ts` (×6), `bootstrap.ts`, `single-flight.ts`** |

The third row owns nearly all the state a run touches: `config.yaml`, the workspace registry, ingest mtimes, map baselines and stitch-scope caches. So `IX_HOME` isolated the CLI *install* and almost nothing else.

Concretely, a harness that sets `IX_HOME` to keep off a developer's machine still:

- reads its **endpoint** from that developer's `~/.ix/config.yaml`, and
- registers its throwaway worktrees there as **workspaces**.

## How it surfaced

A stale `endpoint: http://new.example` left in a real `~/.ix/config.yaml` failed **17 Ix-mode runs** of an outside benchmark with `Ix backend not reachable`, while every baseline run in the same experiment passed. One arm failing and the other not looks like an Ix regression. It was a leaked environment.

Where that stale value came from is the second half of this PR. `config-file-mode.test.ts` isolated itself by overriding `HOME`/`USERPROFILE`, which works **only** under vitest's default forks pool: `os.homedir()` answers from the environment libuv sees, and a worker thread's `process.env` write never reaches it. Run that file with `--pool=threads` and `homedir()` returns the real home, so every `saveConfig` in it rewrites the developer's actual config — and because `workspaces` is in `OSS_OWNED_KEYS` and the config object those tests pass has none, it **deletes their entire workspace registry**. The six failing assertions are the only warning you get.

## The change

- **`ixHome()`** (new, `src/cli/ix-home.ts`) is the single source of truth: `$IX_HOME`, else `~/.ix`.
- Resolved **per call**, deliberately not a module-level const. `const IX_HOME = process.env.IX_HOME || …` makes isolation a module-registry problem rather than an environment one — a test that sets IX_HOME in `beforeEach` is isolated when its file runs alone and not when another file imported the module first. That is exactly how #498 wrote a bogus `99.0.0` upgrade notice into a real `~/.ix`.
- `config.ts`, `bootstrap.ts` and `single-flight.ts` route through it.
- `config-file-mode.test.ts` moves from the `HOME` override to `IX_HOME`, which is read off the object and therefore isolates under either pool.

Not touched: the five modules that bind IX_HOME at import. Each derives further path constants from it (`COMPOSE_DIR`, `PID_FILE`, `STANDARD_BACKEND_DIR`…), so converting them ripples through every use site. Worth doing separately.

## Verification

- `config-file-mode.test.ts`: passes under **both** `--pool=threads` and the default pool, and leaves `~/.ix/config.yaml` byte-identical under both. Before this PR, threads left it rewritten with the registry gone.
- New `ix-home.test.ts` covers the fallback, the per-call re-read, and that config.yaml, the 0700/0600 mode guarantees and the three per-project caches all land under IX_HOME.
- Gates: `typecheck` 0, `lint` 0 errors, `knip` 0, full suite green.

## Behaviour change — worth a reviewer's opinion

With `IX_HOME` set, the CLI now reads and writes `$IX_HOME/config.yaml` instead of `~/.ix/config.yaml`.

Neither installer exports `IX_HOME` (`install.sh` and `install.ps1` persist only `PATH`), so this reaches only people who set it deliberately — for whom following it is the point. But someone who set it as a custom *install root* and has an existing `~/.ix/config.yaml` will find the CLI looking at an empty directory: default endpoint, no workspaces, until they move the file or unset the variable.

I documented it in `docs/prerequisites.md` rather than adding a read-time fallback to `~/.ix`, which would split config across two locations. Happy to add a one-time migration instead if you'd prefer that.


KageBinary:
Review follow-ups, pushed as 3971206:

1. **Tests could still write to a developer's real state directory.** Eight test files isolate by overriding `HOME`. Now that `config.ts` honours `IX_HOME`, anyone with `IX_HOME` exported (every custom-install-root user) would have had those files write their fixture config into `$IX_HOME`, because a HOME override relocates `~/.ix` but not `$IX_HOME`. A vitest `setupFiles` entry (`vitest.setup.ts`) now deletes the ambient variable before every file; files that set `IX_HOME` do it in `beforeEach`, which runs after. Verified: running three of those files with `IX_HOME=<tmpdir>` exported leaves the directory empty.
2. **A silently recreated config.** Someone who exported `IX_HOME` as an install root before this change finds a fresh default config (default endpoint, no workspaces) with nothing saying why. `ensureLocalConfig` now prints one stderr line naming the untouched `~/.ix/config.yaml`, only at the moment it creates the new file. Nothing is migrated automatically.
3. The "nothing landed in the real home" assertion now compares the real `config.yaml` before and after, instead of checking a `.tmp` filename that is never written. `docs/api/README.md` gives `IX_LOCK_DIR`'s new default.

Gates: typecheck clean, lint 0 errors, knip clean, 105/105 across the ten affected test files, `ix-home.test.ts` also green under `--pool=threads`, and the real `~/.ix/config.yaml` byte-identical afterwards. CI 25/25.

PR #674, opened 2026-09-15
fix(cli): reconcile reset outcomes and require Pro auth initialization

josephismikhail:
A lost reset status previously advised rerunning an operation that might already have deleted state. The client now requires a 202 UUID acknowledgment and status for that operation. Missing, denied, invalid or unavailable status, partial failure and observation deadlines stop with the operation ID and reconciliation guidance; none submits another reset. Local synchronous behavior and fallback for a missing async start route remain. Missing status never triggers that fallback, and rejected starts remain errors.

An installed Pro could fail silently and leave commands without its credential guard. The loader distinguishes an absent optional plugin from an installed broken plugin, awaits registration and rejects missing exports/dependencies or initialization failures before parsing commands. Knip permits the optional private plugin; OSS does not fetch it as a dependency. No authentication bypass or tenant-header override is added.

Validation on reviewed Node 22.23.2/npm 10.9.8: 1,057 Pro tests across 103 files, 28 CLI security tests, 27 registry checks, 11 release-input checks, two registry-helper checks, lint/typecheck/build and packed installation. The packed gate verifies 45 commands, six synthetic credential boundaries, shared CLI module identity, pruning recovery and damaged-peer rejection. Full native parsing and installed help/version pass before and after production-dependency pruning on local Linux x64, Windows x64, and now Linux ARM64 under QEMU/PRoot. ARM64 also passes real loopback TLS tests for credential injection, 307/308 redirect rejection, safe manual redirects and certificate rejection. Its 156 compiled Pro runtime files match the reviewed Windows output byte-for-byte. Synthetic credentials and emulation do not establish real identity or native hosted-platform acceptance.

Evidence: `security-hardening-execution-20260909/pro-arm64-release-runtime-20260917/`, `pro-windows-release-runtime-20260917/`, and `cli-pro-installed-pair-20260917/`. The ARM64 attempt's host-emulation and old-PRoot filesystem failures are retained separately; application tests pass with pinned official PRoot 5.4.1. No application source, cloud configuration or UI changed during this acceptance check.

The 28 CLI tests include 22 real-client reset sequences and six real-process plugin-loader checks. Existing CLI source CI passes; it does not prove the private installed pair on every platform. Pair this change with ix-infrastructure/Ix-pro#127 and the kOS reset/status correction. Keep draft pending the complete supported release/platform matrix, reviewed publication and authenticated dev/prod workflow/rollback checks. See `docs/security/reset-outcome-reconciliation.md`.


Alot1z:
Review note on the new `docs/security/reset-outcome-reconciliation.md`:

The doc states the status ledger is "process-local and tenant-authorized" under the name **kOS**. Neither term appears anywhere in this repository — `kOS`: 0 hits, `tenant`: 0 hits across `ix-cli/src`, `core-ingestion/src`, and `docs` on current main (and the PR itself introduces them only in this doc). The client-side machinery the PR actually builds is concrete and checkable: the 202+UUID start ack, `/v1/reset/status/:opId`, the reconciliation paths, and `reset-outcome.test.ts` asserting the exact request sequence.

If `kOS` is the backend's internal name for its status store, one sentence saying so ("kOS — the backend service's process-local operation store, source not in this repository") makes the reference verifiable instead of mysterious. If it is not, grounding the wording to the artifact a reader can inspect ("the backend's process-local reset status store") would keep the security doc fully checkable against code — the same standard the rest of the PR holds itself to, and the one this repo's reference-parity gates enforce.

No objection to the reconciliation logic itself — the never-resubmit-on-unknown-outcome contract is the right safety property, and the test coverage listed matches it.

PR #676, opened 2026-09-17
feat(context): say where evidence is, and rank members by use

KageBinary:
> **Builds on #675.** This branch contains #675's commit so the two don't conflict in `context.ts`. Review from `d41fed8`; once #675 merges, the diff shrinks to this change alone.

## Problem

This is what `ix context config.ts` hands an agent today:

```
  [0] target — config.ts (file)
         resolved target — the bundle is centered on this entity
  [13] structural — member isPathInsideResolvedRoot
         direct structural relationship
  [14] structural — member createClient
         direct structural relationship
  ...
  [22] structural — member real
         direct structural relationship
  [40] provenance — history length 1, introduced rev 2651
         provenance unknown, extractor unknown
```

- **No locations.** The Ix repo has several `config.ts` files, and none of the evidence says where anything is. In agent benchmarks the Ix-mode run's first tool call was nearly always `find . -name <the file Ix just named>`. That re-search is most of why the bundle saved almost nothing.
- **Arbitrary members.** The backend returns members in no useful order, and ten are shown. `resolveWorkspaceRoot` (12 users in 8 files, the function the file exists for) was cut. `real`, a two-line helper, made the list.
- **Provenance is never read.** `/v1/provenance` returns `{entityId, chain: [{source: {uri, extractor, …}}]}`, but the bundle read `sourceUri`/`extractor` off the top level. Every bundle said "unknown".
- **Relationship titles are raw UUIDs** (`264cc04d-… --CONTAINS--> d1898d46-…`).

## Change

The same command on this branch:

```
Context: config.ts
  kind:          file
  path:          ix-cli/src/cli/config.ts
Evidence (highest relevance first)
  [0] target — config.ts (file)
         ix-cli/src/cli/config.ts — resolved target — the bundle is centered on this entity
  [10] structural — dependent stats.ts
         ix-cli/src/cli/commands/stats.ts — calls, imports or references the target
  [13] structural — member getEndpoint
         ix-cli/src/cli/config.ts:198-200 — defined in the target; used by 35 across 31 other files
  [14] structural — member resolveWorkspaceRoot
         ix-cli/src/cli/config.ts:322-342 — defined in the target; used by 12 across 8 other files
  [17] structural — member findWorkspaceForCwd
         ix-cli/src/cli/config.ts:285-287 — defined in the target; used by 6 across 4 other files
  ...
  [40] provenance — history length 1, introduced rev 2651
         provenance code, extractor tree-sitter/1.25
```

- **`location` on evidence** (`path`, `lineStart`, `lineEnd`) for the target, members, callers and dependents.
  - Text prints it on the second line. llm adds `path=`, `lines=` and `target_path=`. JSON adds `target.path` and entity lines.
  - `collectFacts` already had all of this from the expands it makes; it was reducing it to names.
  - A file's own line span is omitted, since it's just the whole file.
- **Members ranked by use**, in `ix context` only.
  - One bounded inbound expand per member: at most 40 members (picked by kind, then size), with `limit: 50` on each.
  - Members used from other files rank first, then by total use.
  - A member that can't be measured keeps its place below the measured ones.
  - `ix explain` keeps graph order and makes no extra requests.
- **Located entities come before** the backend's kind/name-ordered context nodes, so the 50-entity budget cuts what matters least. Only the ten leading members go first; the rest follow the context nodes. Putting all of them first let a large file (`ingest.ts`, 100+ members) fill the budget with its own members and push out the files it imports, which dropped file recall on an extensions question from 0.50 to 0.00 (fixed in `9884032`).
- **Relationship titles use names.** A name two entities share is qualified with its path. The ids stay in the evidence id and in the `relationship` records, so `--diff` matching is unchanged.
- **Provenance** is read from the newest chain entry. A flat record is still accepted.

**Compatibility.** Every new field is optional in `contextBundleSchema`, so saved investigations from before this still parse. The fields are listed rather than left to catch-all, because zod strips unlisted keys on save and read-back. `ix explain --format json` prints `facts` whole, so it gains `memberRefs` / `topCallerRefs` / `topDependentRefs` (additive). `docs/llm-format.md` is updated.

## Cost

- The usage probe adds one ~25ms request per member, six at a time. On `config.ts` (28 members) the command runs in 3.0–4.3s, against 2.5–2.7s with #675 alone and ~8s on main.
- On a kOS class target (`SkillBuiltins`) it went from 1.65s on main to 1.15s.

## Not in this PR

The backend's compact `/v1/context` response also loses paths and edges: node summaries carry `sourceUri: null`, and nodes spend the whole token budget so no edges survive. That's why the header says `relationships: 0`. That's fixed in ix-infrastructure/Ix-memory#210, which adds `path` to summaries and reserves budget for edges. This PR's last commit reads that `path`, and still falls back to `sourceUri` for older backends.

## Tests

- `context-evidence-locations.test.ts` (8 tests): locations on every evidence kind; name-only fallback when facts carry no locations; entity ordering under a tight budget; provenance from the chain; qualified relationship labels; text/llm rendering; schema round trip plus legacy bundles.
- `facts-scope.test.ts`, 4 new tests: use-ranking order and bounded probes; explain untouched; the 40-member cap; an unmeasurable member.
- Two llm expectations in `context-investigation.test.ts` now expect named relationship endpoints.

Full vitest run: 1876 passed. `typecheck`, `eslint` (no new warnings) and `knip` are clean.

## Measured on a coding-agent benchmark

Retrieval from natural entry points (not the answer's own file), same graphs, main build vs this branch:

| task | target | file recall |
|---|---|---|
| locate exit code | `resolveFileOrEntity` | 0.00 → **1.00** |
| ingest base-rev race | `commitPatchBulk` | 0.00 → 0.50 |
| reset blast radius | `clearIngestMtimeCache` | 0.00 → 0.50 |
| context budget flag | `parseBudgetOption` | 0.00 → 0.50 |

The gains all come from callers and dependents that now carry a path.


Alot1z:
**One merge-order note for the `docs/llm-format.md` wave**

Seven of the open PRs touch `docs/llm-format.md` (#676, #677, #678, #682, #688, #690, #696). I verified the overlaps against the actual patches (`gh api …/pulls/<n>` diff hunks), and they fall into two groups:

**1. Three PRs rewrite three different prose sections** — these do not conflict with each other, but each rewrites prose the other two also move:

| PR | `docs/llm-format.md` hunk begins at |
|---|---|
| #682 | line 11 (`… hand-written renderer emit compact records …`) |
| #690 | line 20 (`… so consumers can pass the flag unconditionally …`) |
| #688 | line 48 (`… region id=cli kind=subsystem label="Client" …`) |

**2. Three PRs patch the *same two regions*** — the one real collision:

| PR | hunks |
|---|---|
| #676 | line 94 (`item name=parseFile kind=method`) **and** line 114 (`budgets … chars=12000`) |
| #677 | line 94 and line 114 — *identical hunk context* |
| #678 | line 94 and line 114 — *identical hunk context* |

All three carry the same `@@ -94,16 +94,24 @@` and `@@ -114,7 +122,7 @@` hunks against the same base. They are a stack, so in-order merging rebases cleanly, but landing them out of order means three successive rewrites of the same two blocks. A one-line "land #676 → #677 → #678" from a maintainer removes that entirely.

**Stacking to be aware of:** #696 is based on #688's branch (`fix/honest-truncation`) and its `llm-format.md` hunks sit at 118/138 — i.e. it already assumes #688 landed.

**What I actually measured** (so the rest is separable from the above): the Context chain through #695, rebuilt on upstream `main` and run on this machine — **1898 passed / 22 skipped / 0 failed** at tip `2ce3c75`. That is the one number here I can vouch for from execution; everything above is hunk arithmetic from the API.

The designs compose — this is churn minimisation, not a change request.

PR #682, opened 2026-09-19
feat(cli): let IX_FORMAT and config.format choose the default format

KageBinary:
## What

`--format` falls back to `IX_FORMAT`, then `format` in `~/.ix/config.yaml`, then `text`. An explicit flag still wins.

## Why

`config.format` has been in the schema, loaded, saved, and printed by `ix config show` since the first release — and read by nothing. Every command hard-codes `"text"`. So the only way to move a caller off a renderer is to append `--format llm` to every call, which each of the six plugins has to remember on every command it wraps, and which an agent has to be told to do in prose.

The formats are not close in cost. On `ix context` the same answer comes out at roughly **7.6 : 0.8 : 1** for json : llm : text; the worst recorded `context --format json` was 40.8 KB. One env var is the difference between a plugin paying that on every call and not.

## What changed

- `default-format.ts` — `resolveDefaultFormat(env, readConfigFormat)`, returning the format and, when a configured value was not a format, which one was ignored. The config read is a thunk: a valid `IX_FORMAT` answers without touching `config.yaml`, and this runs during registration on every invocation, `ix --help` included.
- `register/oss.ts` — applies it to every OSS `--format` option. Two guards: the option must still be sitting on the built-in `text` (a command that declares a different default meant it), and the resolved format must be one that command accepts, so `IX_FORMAT=llm` leaves `query` — which renders `text` and `json` — where it is. `.option()` copies the default into the command's option values as it registers, so the stored value is moved alongside `Option.default()`; moving only the latter changes the help text and nothing else.
- `config.ts` (command) — `ix config set format yaml` now fails with the choices instead of saving a value that silently does nothing. `ix config show` marks a stored format that `IX_FORMAT` is overriding, or an `IX_FORMAT` being ignored.
- `docs/llm-format.md` — a "Choosing it once" section.

## Deliberately not done

**No switch on `process.stdout.isTTY`.** It is tempting — an agent is never a TTY — but a script that has always parsed `text` would start receiving records on its next run with nothing it could have done about it. An env var is the opt-in.

**An unrecognised value is ignored, not fatal.** `IX_FORMAT` is set once in a profile or a plugin launcher and read by every later command; making a typo fatal breaks all of them at once. It is reported on stderr, and only when stderr is a terminal — printing it into a captured stream is the cost this change exists to avoid.

## Tests

`__tests__/default-format.test.ts`, 15 cases: the resolution order, the lazy config read, padding/case, empty vs unrecognised values, which ignored value is reported — then five through a real `Command` with an isolated `IX_HOME`, because what matters is commander's precedence: flag beats env, `query` stays on text, an unrecognised value leaves the default alone.

Verified against the local backend: `ix stats` with `IX_FORMAT=llm` emits `nodes total=1425866 …` records; unset, it prints the text table.

Full suite green (106 files, 1871 passed / 2 skipped), `typecheck`, `lint`, `knip` clean.

## Follow-ups this unblocks

The plugins can drop their per-command `--format` handling and set one variable (spec tasks P-2/Q-1…Q-3), and the `ix mcp` shim can stop being the only thing that gets llm output right.


Alot1z:
Coordination note, not a review point: this PR's `docs/llm-format.md` hunk (line 11) sits alongside #690's (line 20) and #688's (line 48) — three different sections of one file, so all three compose. Flagging only so the file is not rewritten three times if they land in an order nobody planned.

Alot1z:
Runtime verification of this PR at its own head (`52675ef`), against a real backend rather than the suites. Hermetic stack (`arangodb:3.12.11` + memory-layer `1.0.28`, isolated ports, sandbox `IX_HOME`); CLI = this branch's own build; `config.yaml` in the sandbox driven deliberately and restored afterwards.

I probed the chain as a multiplicative contract — every adjacent pair, crossed with every way "unset" can be spelled — because a precedence bug at the bottom of the chain is invisible from the top.

## Every layer and every pair

| probe | observed |
|---|---|
| `format: llm` in config, no env, no flag | `nodes total=7002 function=7000 file=2` — the config value is read at last |
| env=llm, config=text | llm (env outranks config) |
| env=text, config=llm | text (and the pair above rules out "last writer wins") |
| env=llm + `--format json` | json (flag outranks env) |
| env=llm + `--format text` | text |
| `IX_FORMAT=` (empty) | config used — empty reads as unset, not as a value |
| `IX_FORMAT="  llm  "` | llm (padding) |
| `IX_FORMAT=LLM` | llm (case) |
| `IX_FORMAT=yaml`, stderr piped | config used, **and not one extra line on stderr** — the TTY-only rule holds |
| `IX_FORMAT=yaml` + `--format text` | flag wins over the ignored value |
| valid `IX_FORMAT=llm` + malformed `config.yaml` | llm, rc=0 — the config read is genuinely lazy |

## The observability half, which is the part that is easy to get wrong

The body notes that `.option()` copies the default into the command's option values, so moving only `Option.default()` would change the help text and nothing else. Both halves are right in this head:

```
$ ix stats --help                          # no env
  --format <fmt>  Output format (text|json|llm) (choices: "text", "json", "llm",
                  default: "text")
$ IX_FORMAT=llm ix stats --help
                  default: "llm")
$ IX_FORMAT=json ix stats --help
                  default: "json")
```

And sweeping **all 18 commands that offer `--format`**, every one follows the variable (18 moved, 0 stale) — while `query`, the guard case, holds `(choices: "text", "json", default: "text")` under `IX_FORMAT=llm`, which is the only correct answer since it cannot render llm at all.

The `config` command behaves as described, including the part that is easy to skip — not writing anything:

```
$ ix config set format yaml
Not a format: yaml
Choose one of: text, json, llm          # exit 1, config.yaml unchanged

$ ix config show
format      llm  (overridden by IX_FORMAT=text)
format      text  (IX_FORMAT=yaml ignored: not a format)
format      text                                  # agreeing: no marker
```

## Three things the suite leaves unpinned

All three are asserted in the body and observed above, but I could not find a case that would fail if they regressed:

1. **`--help`'s advertised default.** This is the exact coupling the description flags as the easy mistake, and `default-format.test.ts` drives five cases through a real `Command` without reading help text. One case that runs `--help` with `IX_FORMAT` set and asserts the default line would pin both halves of the option move.
2. **`config show`'s two markers** (override / ignored), which are the only feedback a user gets that their env var is not being used.
3. **`config set format <bad>` refusing** with exit 1 instead of saving a value that does nothing — the behaviour this PR fixes, and the one a later refactor could quietly undo.

Related, not a defect: the first guard ("the option must still be sitting on the built-in `text`") has no current occupant — all 18 commands default to text today. It reads as future-proofing, which is fine; worth knowing it is currently unexercised.

## One line for the new "Choosing it once" section

`IX_FORMAT` today governs the **success** path only. A failing command still answers in prose on stderr with empty stdout, even with the variable set:

```
$ IX_FORMAT=llm ix explain nosuchsymbol
No entity found matching "nosuchsymbol".        # stderr, exit 1, stdout empty
```

That is expected here — #689 ("answer a failure in the format the caller asked for") is not in this base — but the new section tells a reader to set the variable once, globally, and that reader will form an expectation about failures too. A sentence noting they follow once #689 lands would close the loop.

*(Aside so the numbers above aren't confusing: with the malformed config, the format chain still worked, but the destroyed `workspaces:` list meant cwd-based resolution picked a different workspace — 7050 nodes vs 7002. Pre-existing config robustness, unrelated to this diff.)*

Alot1z:
My earlier comment today covers the precedence contract itself — this is only an addition it did not measure: **the cross-PR dependency on `#684`**, now verified at runtime on both trees.

`#684`'s new `skills/ix/references/output-formats.md` documents exactly this contract (resolution order, the `config show` override notice, the `query` fallback), but on `#684`'s own head (`80a9ce1`) `IX_FORMAT` appears in **zero** files of the repo and is ignored at runtime — `IX_FORMAT=llm search` emits text on that build. Every one of the claims becomes true on this PR's build (`52675ef`), which I confirmed independently (env-selected llm output, the exact `format llm (overridden by IX_FORMAT=json)` notice, env-over-config precedence, the `query` text fallback).

So the skill reference and the implementation must land in the same release: if `#684` merges first, the shipped skill documents an env var no released CLI honors — the precise doc-vs-code drift the reference-parity gates exist to catch. Details are in my comment on `#684`.

PR #684, opened 2026-09-19
docs(skill): steer the skill to ranges, and stop shipping install notes in it

KageBinary:
## What

`skills/ix/SKILL.md` 8.0 KB → 3.8 KB, description 474 B → 192 B, the install-seam section moved to `docs/skill-harnesses.md`, and the guidance rewritten around what actually costs tokens.

## Why

A skill body is loaded before it can be used, so its size is paid on every invocation. 2.9 KB of this one described where `scripts/install-skill.sh` writes, how `TOOLSCAN_PATH` and `HARNESS_HOME` steer its probes, and what `--dry-run --json` emits — contributor documentation, in a file an agent loads to ask what calls `verify_token`. It is now `docs/skill-harnesses.md`, linked from the docs index, with nothing lost.

The rest is about where the tokens go:

- **Ranges, not files.** `ix read <symbol>` or `path:120-180` is 1–3k tokens; the file around it is commonly 30–60k, and every one of those is re-sent on every later step of the turn. In recorded benchmark runs, whole-file reads re-sent 10–80× were the largest sink Ix could have removed and didn't, because nothing told the agent to ask for the range. The body now opens on `ix context` and says to read what it points at.
- **Don't poll `ix status`.** Of nine `ix` calls an agent made across eleven recorded runs, **three were `ix status`**. A command that needs the backend says so itself.
- **`ix map` is not a per-edit step.** The old rule 4 ("Immediately after modifying code, run `ix map --silent`") is a tool call per edit for work the editor plugins already do on a debounce.
- **Format.** `references/commands.md` had **14 `--format json` examples against 12 `llm`**, and agents copy examples. On `ix context` the same answer runs ~7.6 : 0.8 : 1 for json : llm : text — json is about nine times llm. The examples now carry no format flag, with one line at the top saying to set it once and to use `--format json` on the call whose output is actually parsed.

## Correction

`references/output-formats.md` still said `explain`, `read`, `status` and `query` "do not implement `llm` and fall back to text". Three of those four do — `explain`, `read`, `status`, plus `doctor` and `savings` (Tier 5, per `docs/llm-format.md`). Only `query` (deprecated, `text|json`), `diff --content` and `ingest` fall back, which is what it says now. `explain` is the one that mattered: it is the first call most plugins make, and its records are ~55% smaller than the json people were using as a workaround for advice that was out of date.

## Depends on

The "set it once" lines reference `IX_FORMAT` / `ix config set format llm` from **#682**. If that lands after this, these two sentences are early rather than wrong — the flag they replace still works.


Alot1z:
Two-part verification of this PR at its own head (`80a9ce1`): the harness doc verifies cleanly; the new format-reference text carries a hidden dependency on `#682` that should be stated before either lands.

## Part 1 — `docs/skill-harnesses.md` verifies against code and runtime

- The claimed table matches the source exactly: claude → `~/.claude/skills`, agents → `~/.agents/skills`, codex → `~/.codex/skills`, cursor → **`~/.cursor/skills-cursor`** (`ix-cli/scripts/skill-harnesses.mjs`), with gemini/opencode/openclaw/vscode deliberately absent.
- `--probe` runtime, hermetic (`HARNESS_HOME=$(mktemp -d)`): `present=0` for agents/cursor, `1` for claude/codex — and `detectedVia` attribution works for real: on a machine where agents and cursor are installed but not on PATH, the probe reports them via `config-dir`, not `path`. The doc's claim that the attribution distinguishes the two is true on actual data.
- `install-skill.sh --dry-run --json` emits the documented per-host contract (`would-install`/`skip`, `dest`, `detectedVia`), rc 0.
- The SKILL.md de-duplication claim holds: zero references to the harness table remain in `skills/ix/SKILL.md` on this branch.

## Part 2 — the new format-reference documents a mechanism this PR does not contain

`references/output-formats.md` now documents: resolution order (`--format` → `IX_FORMAT` → `config.format` → text), `config show` reporting when `IX_FORMAT` overrides the stored value, and per-command fallbacks for `query`. Measured on **this PR's own build**:

- `IX_FORMAT=llm` is ignored — `search` emits text exactly as it does with no env var (`--format llm` on the same build works, so the renderer is fine; the env leg of the chain is absent).
- `IX_FORMAT` appears in **zero** files in the repo at this head (and GitHub code search agrees), on main included.
- `config show` prints only `format llm` under `IX_FORMAT=json` — no override notice.

Every one of those claims then verifies **true** on `#682`'s build (`52675ef`), which is where `IX_FORMAT` actually lives (33 patch-hits there): env-selected llm output, the exact override notice `format llm (overridden by IX_FORMAT=json)`, and env-over-config precedence all behave precisely as this PR's reference describes. The documentation is correct — against an implementation in another open PR.

## Suggested smallest action

Record the dependency in this PR's body (or stack it on `#682`): the `output-formats.md` hunks are only accurate once `#682` merges. If this PR lands first, the shipped skill documents an env var that does not exist in any released CLI — precisely the doc-vs-code drift this repo's reference-parity gates exist to catch. The same pair-ordering note applies to the `docs/llm-format.md` hunks in `#682`/`#688`/`#690`, which this PR's reference text cross-references implicitly.

No other issues found: the SKILL.md slimming (54/−92) removes install-machinery duplication without touching the workflow content agents actually consume.

PR #685, opened 2026-09-19
fix(cli): make an ambiguous target pickable from one call

KageBinary:
## What

`--format llm` now names each candidate for an ambiguous target, says where a `--path`-excluded name does live, and reports a miss once instead of twice.

## Why

This is the whole llm answer for an ambiguous symbol today:

```
error code=ambiguous_target message="Ambiguous symbol \"config\"." candidates=1:config.ts,2:config.ts,3:config.ts
```

Three candidates that differ in nothing, in the only field included. Nothing there can be acted on, so the caller spends another call to find out what it was offered — and `ix read` has emitted name, kind, path and id per candidate since its own renderer landed (`read.ts:120-148`). Everything else went through `ui.ts:189-193`, which did not.

After, against the live graph:

```
error code=ambiguous_target message="Ambiguous symbol \"config\"." count=8
candidate n=1 name=Config kind=class path=control-plane/rollback-runner/internal/rollback/rollback.go id=01bb9140
candidate n=2 name=Config kind=class path=data-plane/memory-layer/bedrock-backend/src/main/scala/ix/memory/db/bedrock/BedrockShadowCanary.scala id=02424d66
…
hint text="Use --pick <n> or --kind or --path to disambiguate."
```

`ix locate` carried its own copy of the old rendering (`locate.ts:221`); it now calls the shared reporter.

## Also

- **A miss under `--path` says where the name does live.** The candidates were fetched, filtered out and dropped; "`Widget` exists, in `src/ui/widget.ts`, just not under `server/`" is the answer that filter was asking for. Emitted as `suggestion` records in llm, a `suggestions` array in json, and a "Did you mean:" block for a person.
- **A miss is reported once.** The resolver wrote prose to stderr and the reporter wrote a record to stdout. An agent runs `ix … 2>&1`, so it read the same miss twice. The prose is now skipped when a structured format will carry it — that is what the new `format` on `ResolveOpts` is for.
- Ids in llm records are the 8-char prefix; json keeps the full uuid.

## Deliberately not done

A did-you-mean for a name that matched *nothing*. The idea was to offer the top search rows already in hand — but on that path there are none: the search returned zero rows, which is why it is a miss. I checked whether dropping `nameOnly` would widen it, on a 1.4M-node graph:

| term | `nameOnly: true` | `nameOnly: false` |
|---|---|---|
| `resolveWorkspaceRooot` | 0 | 0 |
| `conflictdetector` | 5 | 5 |
| `budgetPause` | 5 | 5 |

Same rows either way, so a second search would have been a round trip that could not help. (A partial name like `ingestFil` already resolves as *ambiguous*, not a miss, so it gets the candidate list above.) Real fuzzy matching would be a backend change, not a CLI one.

## Tests

`__tests__/ambiguity-and-suggestions.test.ts`, 8 cases: the record shape and ordering for candidates, full uuid retained in json, the hint dropping flags already supplied, the `--path` suggestions in llm/json/text, the miss printed once, and the empty case — which also asserts exactly one search call, so nobody adds the round trip back by accident.

Full suite green (106 files, 1870 passed / 2 skipped), typecheck, lint, knip clean.


Alot1z:
Independent verification on this PR's own build (head `94dd955`, hermetic backend 1.0.28, sandbox `IX_HOME`, corpus with two same-named exports `src/alpha.ts` + `src/beta.ts` — the exact case this PR is for).

**The one-call contract verifies:** `callers compute --format llm` → `error code=ambiguous_target … count=2` + per-candidate records `candidate n=1 name=compute kind=function path=src/alpha.ts id=32abb6b2` / `n=2 … path=src/beta.ts id=8bd9deaf` — distinguishable without re-querying. `--pick 2` lands on the right entity in the same call (`ref name=runAll … path=src/main.ts`); `--path src/alpha` disambiguates; exit 1; JSON carries full `candidates[]` (id/name/kind/path/score/rank). The flag-aware hint's "drop what was already given" behavior is suite-pinned.

**One rendering finding in this PR's own output path:** the llm payload now carries the same guidance twice, worded differently:

```text
hint text="Use --pick <n> or --kind or --path to disambiguate."
diagnostic code=ambiguous_resolution message="Use --pick <n> or --path to disambiguate."
```

The diagnostic is the pre-existing static one (main `resolve.ts:496`, forwarded on `AmbiguousResult`); the hint is this PR's flag-aware one (which is strictly better — it omits flags the caller already passed). On main, neither line reached the llm stream, so the duplication is introduced by this PR's rendering. Smallest fix in the llm branch of `reportAmbiguousTarget`: when the hint renders, suppress a diagnostic whose message duplicates it (or retire the static diagnostic at `resolve.ts` now that the flag-aware hint covers all rendered formats). Probe transcript available if useful.

PR #686, opened 2026-09-19
perf(cli): stop paying for JSON indentation nobody reads

KageBinary:
## What

`--format json` pretty-printed at every one of its 49 call sites. Two spaces
per nesting level is whitespace the only readers that matter — a pipe into
`jq`, an agent's tool result, the `ix mcp` child process — throw away.

Measured on this machine, `ix mcp doctor --format json`:

| | bytes |
|---|---|
| today (indented) | 1,935 |
| this branch, piped | 1,441 |
| saving | **25.5%** |

That applies to every JSON call the CLI makes, on top of whatever the payload
itself costs.

## How

One helper, `printJson`, in `format.ts`. It decides at print time, not import
time:

- stdout is a terminal → indent 2, exactly as today
- anything else → compact

`--pretty` forces the indented shape anywhere. It is declared once, in
`configureOssOptions`, for every command whose `--format` accepts `json` — 33
of them — rather than typed out 33 times, and the root `preAction` hook records
it before any renderer runs. No renderer signature changes.

## Not in scope

- the two `writeFileSync` state files in `context.ts` and the MCP host config
  in `mcp/install.ts`: people read those, and none of them is on a token budget
- `mcp/server.ts:688` — the `ix_context` double delivery is its own task (M-1
  in the token-usage plan)

## Verification

  (106 files, 1,869 passed, 2 skipped) · `node test/parser.smoke.mjs` — all green
- `node scripts/check-doc-parity.mjs` — green; the 33 new `--pretty` rows are in
  `skills/ix/references/flags.md`, and `output-formats.md` says JSON is compact
  off-TTY
- new `json-output-shape.test.ts` covers: compact off-TTY, indented on a TTY,
  `--pretty` overriding off-TTY, the shape being re-read per call, every
  json-capable command offering `--pretty`, commands without JSON not offering
  it, and the flag arriving through the `preAction` hook

Part of the CLI output-economy work (task C-7).


Alot1z:
Independent verification on this PR's own build (head `88f966a`, hermetic backend 1.0.28, sandbox `IX_HOME`), including the two edges most likely to be wrong — both check out.

**Measured savings** (compact vs re-indented same payload, so "pretty" is exact):

- `search probeToken --format json --limit 20` → **4,012 B compact vs 5,712 B pretty (29.8% saved)**, parses clean, `summary.count` intact.
- `context probeToken1 --format json` bundle → **2,095 B vs 2,743 B (23.6% saved)**.

**Contract verified:** off-TTY stdout is compact; `--pretty` restores indent on a pipe. `--pretty` is auto-added to every command whose `--format` includes `json` (`register/oss.ts:102`), recorded in the root preAction (`oss.ts:156`), and the suite walks the whole program asserting both presence and absence (`json-output-shape.test.ts`) — a new JSON command without it is a compile-time-discoverable test failure, which is the right shape for this invariant.

**Edges that could have been wrong, aren't:** the two remaining `JSON.stringify(…, null, 2)` sites in `context.ts` (379, 680) are on-disk saved-investigation writes — correct to keep pretty, since files are read by humans and diff tools. No `console.log(JSON.stringify)` bypass remains anywhere in `commands/`.

One compat note worth having on the record: the wire change is whitespace-only (same JSON semantics), so any consumer that *parses* is unaffected; only something regex-matching the indented shape would notice. Nothing in-repo does.

Probe transcript available on request.

PR #687, opened 2026-09-19
fix(search): prune the answer before the limit, and score it the right way up

KageBinary:
## What

Three kinds of row `ix search` prints that the caller cannot act on, plus a
score that pointed the wrong way.

### Chunk twins

The ingester emits one chunk per symbol, named after that symbol and carrying a
`DEFINES` edge to it (`core-ingestion/src/patch-builder.ts:264-300`). A name
match therefore matches both, and the pair arrives as two rows with the same
name and the same file — one of which no command can follow, because `explain`,
`callers` and `read` all want the symbol, not the chunk.

Folded away when the symbol is in the same answer. A chunk with no twin stays:
an unnamed `file_body:` chunk is the only row standing for that part of the
file.

### Bottom-tier fuzz

Tier 5 is the resolver's fallback — nothing in the name matched, a term merely
appeared somewhere in the node's attributes. Dropped as soon as anything
matched by name; kept when it is the whole answer.

Measured on this repo's graph, `ix search search`: 13 rows, every one a test
fixture, were competing for `--limit` slots against real hits.

Both passes run **before** the cut, so a pruned row makes room for a real one
rather than leaving a shorter answer.

### A score that ran backwards

The row carried the internal sort key, which is negative and lower-is-better:

```
node name=search  kind=method   score=-101
node name=search.ts kind=file   score=-48
```

It is now the tier as a 0–1 relevance, best first:

```
node name=search  kind=method   score=0.83
node name=search.ts kind=file   score=0.5
```

1.00 exact name + kind, 0.83, 0.67, 0.50, 0.33, 0.17 incidental. Per-tier and
not per-row deliberately: the sub-score separating two rows inside a tier is two
uncalibrated heuristics added together, and the order it encodes is already
carried by the row order and by `rank` in JSON. A number that looks more precise
than it is invites an agent to threshold on it.

### `unfiltered_search` out of the record stream

It fired on every search that did not pass `--kind` — most of them — and said
the same eighteen words each time about results the caller can already see. Out
of `--format llm`; kept in JSON, where a program may be keyed to it.

## Not fixed here, worth a separate issue

`ix search config` returns eight `name=config.js kind=module` rows whose `path`
is the *importing* file — one module node per import statement (1,211 module
nodes in a 4,614-node graph). That is an ingest-side identity question, not
something to paper over in the renderer.

## Verification

  (106 files, 1,871 passed, 2 skipped) — all green
- new `search-hygiene.test.ts`: 9 cases over `tierRelevance` (positive, bounded,
  monotone, clamped), `foldChunkTwins` (twin dropped, orphan chunk kept, same
  name in another file kept, chunk-only answer untouched) and
  `dropIncidentalMatches` (dropped beside a named hit, kept when alone, and the
  pruned row making room before the limit)
- output above is from a live run against the local backend, not a fixture

Part of the CLI output-economy work (task C-8).


Alot1z:
Independent verification on this PR's own build (head `d64497f`, hermetic backend release 1.0.28, sandbox `IX_HOME`, 7,002-node corpus).

**What verifies:** the score contract works end to end — `search probeToken --format json` reports positive `score=0.5` with `rank`/`tier`/`matchSource` per row; `--format llm` agrees (`search count=3 candidates=9` / `node … score=0.5`); and pruning demonstrably runs before the cut (limit 7 → `count: 7, totalCandidates: 21`; limit 20 → `count: 20, totalCandidates: 60`; limit 50 → `count: 50, totalCandidates: 60`; default → `count: 10, totalCandidates: 30`). The suite pins the ordering decision itself ("runs before the limit, so a dropped row makes room").

**Three measured honesty gaps in the file this PR owns** — the same class the bounded-list PR (#688) already fixed for `callers`:

1. **`totalCandidates` is a moving fetch window, not a total.** `search.ts:226` sets `fetchLimit = Math.min(limit * 3, 60)`, and `search.ts:358` reports `totalCandidates: rawNodes.length` — the size of what was *fetched*. A caller reading `totalCandidates: 21` believes 21 matches exist; raise `--limit` to 7 and the same query reports 63. Measured on one term whose true pool is 10: limit 3→9, 5→15, 7→21, 20→60, 50→60. `count` is honest (== `results.length`); `totalCandidates` is definitionally window-sized. Either rename it (`fetched`) or measure the pool properly — as written it is the only number on the row that changes meaning when the caller asks for more.

2. **#687's own new reductions are uncounted.** `foldChunkTwins` + `dropIncidentalMatches` can shrink the ranked set before `trimmed = ranked.slice(0, limit)` — but the summary carries no `folded`/`dropped` counters, so an agent cannot distinguish "pool exhausted" from "pruned, then limit-cut". The PR's docstring argues a chunk is "a retrieval unit the caller can do nothing with"; that cut is invisible in `--format json|llm`.

3. **No recovery hint.** #688's `callers --limit 2` prints `results_truncated` + `"30 callers; showing 2. Raise --limit to see the rest."`. `search` at any limit prints nothing when the window is fuller than the answer — `raise --limit` is exactly what a caller needs to discover more, and `search.ts` never says it.

Not asking this PR to own all of main's search honesty — but it already rewrites the summary block's neighborhood (`search.ts:340-375`), and its own new drops are the ones currently unreported. Smallest fix inside its scope: count what `foldChunkTwins`/`dropIncidentalMatches` removed and emit a `results_truncated` diagnostic when `totalCandidates > count`, in the format each reader gets (`llmLine("diagnostic", …)` / JSON `diagnostics[]` / a stderr dim line for text). Suite hooks exist: `search-hygiene.test.ts` already pins "runs before the limit"; one more case would pin "the cut is named".

Happy to provide the full probe transcript (commands + raw JSON per limit) if useful.

PR #688, opened 2026-09-19
fix(cli): report what a bounded list left behind, and rank before cutting it

KageBinary:
## The bug

```
$ ix callers verify_token --limit 50 --format llm
callers target=verify_token total=50 resolved=50
```

That symbol has 212 callers. Every one of these commands sliced first and then
reported the length of the slice as `total`, so the single number on the header
said the opposite of the truth, and nothing else on the record indicated the
list was partial. An agent reading it concludes the symbol has 50 callers and
stops looking.

Now:

```
callers target=verify_token shown=50 total=212 resolved=48 unresolved=2
diagnostic code=results_truncated message="212 callers; showing 50. Raise --limit to see the rest."
```

Fixed in `callers`, `callees`, `contains`, `imports`, `imported-by` — all
through `formatEdgeResults` — and in the `callers` text fallback, whose llm
header had the same shape. JSON's `summary` gains `shown` and `truncated`, and
its `total` now means the total.

## Rank before the cut

A cut applied to an unranked list keeps whichever rows the graph happened to
return first. `sliceEdgeResults` puts rows the caller can follow ahead of rows
it cannot:

1. a name **and** a path — directly readable
2. a name alone — costs a `locate` first
3. a dangling raw id — `explain`, `read` and `callers` can all do nothing with it

The sort is **stable**, so rows that rank equally keep the backend's order. This
is deliberately not a general re-sort: the backend order carries information
this layer cannot reconstruct, and the only claim made here is that an
actionable row should survive the cut before an unactionable one.

## Two commands that cannot know a total

Rather than print a number that looks like one:

- **`ix inventory`** asks the backend for `--limit + 1` and reports
  `shown=N truncated=true`. The backend applies the limit and has no count
  endpoint, so "there is at least one more" is the whole honest answer.
- **`ix text`** was never ranked at all, although `ix mcp` advertises it as
  returning "ranked hits" (`server.ts:238`). It was ripgrep's traversal order,
  which on this repo puts `docs/` and `README.md` ahead of any code because `d`
  and `R` sort early. It now scans a bounded window past `--limit` (5x, capped
  at 500), ranks **code > tests/fixtures > prose and lockfiles**, then cuts, and
  reports `shown=N scanned=M`. Ranking the rows it had already cut could not
  have fixed the order — the scan has to see more than it prints.

## Live output from this branch

```
$ ix callers relativePath --limit 3 --format llm
callers target=relativePath shown=3 total=4 resolved=3
diagnostic code=results_truncated message="4 callers; showing 3. Raise --limit to see the rest."
ref name=registerCallersCommand kind=function id=7f7f6640 path=ix-cli/src/cli/commands/callers.ts
...

$ ix text relativePath --limit 5 --format llm
text shown=5 scanned=25 truncated=true
diagnostic code=results_truncated message="25 matches scanned; showing the 5 highest-ranked. ..."
match path=ix-cli/src/cli/ts-module-resolution.ts line=327 ...
```

## Verification

  (106 files, 1,874 passed, 2 skipped) — all green
- new `honest-truncation.test.ts`, 12 cases: the count surviving the cut,
  stability, no mutation of the caller's array, followable rows winning the cut,
  the truncation diagnostic's wording, inventory's `shown`/`truncated`, the scan
  window, and the code/test/prose bands
- `docs/llm-format.md` gains a **Bounded lists** section stating the
  `shown`/`total` convention and why `inventory` and `text` differ

Part of the CLI output-economy work (task C-9).


Alot1z:
Runtime verification of this PR against a real backend, plus the one boundary I could not exercise (recorded so it is not mistaken for verified).

**Setup.** Hermetic stack: `ghcr.io/ix-infrastructure/ix-memory-layer:latest` (image `sha256:109e059a…`, release `1.0.28`, `schema_version: 3`) + `arangodb:3.12.11` on isolated ports (18090/18529), dedicated volume, `IX_HOME` pointed at a throwaway sandbox. CLI = this PR's build at head `28ebb8d3`. Fixture: one 31-line TypeScript file — `hub()` called by 30 `leafN()` functions — `ix map`ped with the same build. stdout and stderr captured separately.

**The motivating defect, measured on `main`** (same fixture, same backend, `5e611d1` build):

```
$ ix callers hub --limit 2 --format llm
callers target=hub total=2 resolved=2
ref name=leaf8 kind=function id=07ea3335 path=src/hub.ts
ref name=leaf15 kind=function id=1281412c path=src/hub.ts
```

`total=2` while the graph holds 30 callers, and no record says anything was left out — `ix callers hub` without the flag reports `total=30 resolved=30`.

**On this PR's build:**

```
$ ix callers hub --limit 2 --format llm
callers target=hub shown=2 total=30 resolved=2
diagnostic code=results_truncated message="30 callers; showing 2. Raise --limit to see the rest."
ref name=leaf8 kind=function id=07ea3335 path=src/hub.ts
ref name=leaf15 kind=function id=1281412c path=src/hub.ts
```

`total` is now the graph's count, `shown` is what the caller got, and the diagnostic states the remainder and the recovery. Sibling commands on the same build, same fixture:

```
$ ix contains src/hub.ts --limit 2 --format llm
contains target=hub.ts shown=2 total=31 resolved=2
diagnostic code=results_truncated message="31 contains; showing 2. Raise --limit to see the rest."

$ ix inventory --kind function --limit 3 --format llm
inventory kind=function shown=3 truncated=true
diagnostic code=results_truncated message="More than 3 function entities exist. Raise --limit, or narrow with --path."

$ ix search probeToken1 --format llm     # 4,001-symbol fixture
search count=10 candidates=30
```

`search` is honest through its own `count`/`candidates` pair, so the two reporting styles coexist without contradicting each other.

**Boundary I could not exercise.** `ix impact --limit` bounds "max top-impacted members" (`ix impact --help`), and this fixture has no members of the cut kind, so its bounded list stayed empty. For the record, because the numbers can look alarming: `impact hub --limit 2` prints all 30 caller rows, and so does `--limit 30` — that is correct here (the member list was empty, so nothing was withheld), but it means **the member cut itself is untested by this run**. A fixture with, say, 20 members of one file and `--limit 3` would settle whether that list reports its own truncation; I am not claiming either way.

**One reproduction trap worth knowing** (not a defect in this PR): a graph `map`ped by a different build of this stack can make every read return `count=0` with a healthy workspace — I hit it before re-mapping with the build under test. `ix doctor` shows the workspace and node counts, so it is diagnosable, but "the CLIs disagree about the stored shape" is a real way to waste a turn.

PR #689, opened 2026-09-19
feat(cli): answer a failure in the format the caller asked for

KageBinary:
## The gap

`--format llm` is the format the skill tells an agent to pass unconditionally,
and it was the one thing the CLI would not answer a failure in. With the backend
down:

```
$ ix callers verify_token --format llm
                                     # stdout: nothing
                                     # stderr: four lines of coloured prose
$ echo $?
1
```

A caller parsing records sees an empty stream and a non-zero exit with no reason
attached, and spends a turn finding out why — after waiting 1.5–4.7s for the
connect to fail.

## After

```
$ ix callers verify_token --format llm
error code=backend_unreachable message="Ix backend not reachable at http://localhost:9999." hint="Start it with `ix docker start`, then check `ix status`."
$ echo $?
1
```

One record, on **stdout** — stdout is where the answer goes, and "there is no
answer, because X" is an answer — and the prose is skipped rather than printed
alongside, so there is exactly one report. Exit code unchanged.

Four codes, all stable:

| code | when |
|---|---|
| `backend_unreachable` | ECONNREFUSED, ENOTFOUND, connect timeout, … |
| *(the backend's own)* | the response body carried `error` / `message` / `next` |
| `usage_error` / `resolution_failed` | `CliUsageError` / `CliResolutionError`, with their hint |
| `cli_error` | nothing above recognised it |

`cli_error` rather than a guess: inferring a class from message text would be a
label an agent could act on and be wrong about. `llmQuote` keeps a multi-line
message on a single line.

## Why argv, not the parsed command

`renderCliError` is installed as the handler for `unhandledRejection` and
`uncaughtException`, which can fire before commander has parsed anything. So
the format is read off argv, stopping at `--` so `ix text -- "--format llm"` is
not misread.

Only what was typed. `IX_FORMAT` and `config.format` choose the default in
**#682**; honouring them here before that merges would answer in records for a
run whose payload came out as text. Once #682 lands, its resolver should feed
`setErrorFormat` and this detector can go.

## `ix status` came along

It built its own "not reachable" sentence with no next step, and classified
*every* exception as unreachable — so a 500 from a backend that was up told
people to restart a container that was running fine. It now shares the one
description (`backendUnreachableError`) and rethrows anything that is not a
transport failure to the boundary, which names it.

## Already covered, not touched here

`unresolved_target` and `ambiguous_target` already emit a record with a
non-zero exit, and **#685** adds the candidates and the `hint` line to both, so
this PR deliberately leaves `ui.ts` alone.

## Verification

  (106 files, 1,872 passed, 2 skipped) · `check-doc-parity` — all green
- new `llm-errors.test.ts`, 10 cases: both flag spellings, the `--` stop, that
  `IX_FORMAT` is deliberately not read yet, the dead-backend record with nothing
  on stderr, the unchanged prose when no format was asked for, the backend's own
  code/message/next, the unclassified case, a usage error's hint, and a
  multi-line message staying on one line
- output above is from live runs against a dead endpoint

Part of the CLI output-economy work (task C-10).


Alot1z:
Verified at this PR's own head (`98c5c6f`) against a real backend, not the suites — plus two things in the failure surface this PR's own premise predicts but structured tests don't reach. Both have repros below; the second interacts with the PR's "exactly one report" guarantee.

**Setup:** CLI built from this head (0.10.5). Hermetic backend = `ix-memory-layer` 1.0.28 + `arangodb:3.12.11` on an isolated port; failure cases override `IX_ENDPOINT` only (`http://127.0.0.1:9999` = nothing listening). A purpose-built 30-line backend (`/v1/health` 200, everything else 500 with a structured body) provides the "up but erroring" case. Sandbox `IX_HOME`, real profile untouched.

## The claims, measured

| Claim | Measured |
|---|---|
| dead backend + `--format llm` → one record on **stdout**, exit 1, prose skipped | ✅ `error code=backend_unreachable message="Ix backend not reachable at http://127.0.0.1:9999." hint="Start it with `ix docker start`, then check `ix status`."` — 156 B stdout, exit 1 |
| *(the backend's own)* code when the body carries `error`/`message`/`next` | ✅ up-but-500 backend → `error code=internal message="boom from a backend that is definitely up" hint="retry once, then check the memory-layer logs"` — `next` becomes `hint` |
| `ix status` no longer calls a live-but-broken backend unreachable | ✅ server whose **health 500s** → `error code=internal message="health handler exploded" hint="restart the memory layer"`, exit 1. This is the exact case the PR body describes (restart advice for a running container) |
| exit code unchanged | ✅ 1 on every failure path above, 0 on success |
| success path untouched | ✅ `search … --format llm` still emits records (`search count=0 candidates=0` + diagnostic) |
| `IX_FORMAT` deliberately not read until #682 | ✅ confirmed at binary level: `IX_FORMAT=llm` + no flag → no record, prose only |

The llm-only scope is honoured as written: `--format json` / `text` / unset all give 0 B stdout + prose (the header comment's motivating complaint still applies to a JSON-parsing caller, but the body scopes this PR to `llm` explicitly, so I read that as a boundary rather than a defect).

## Finding 1 — commander's own errors never reach this boundary, so `--format llm` still yields an empty stream

The PR's stated gap is "a caller parsing records sees an empty stream and a non-zero exit with no reason attached". For the invocation-error class that is still the behaviour at this head:

```
$ ix stats --bogus-flag --format llm         # rc=1, stdout: 0 bytes
$ ix rank --format llm                       # rc=1, stdout: 0 bytes  (required option missing)
$ ix frobnicate --format llm                 # rc=1, stdout: 0 bytes
error: unknown option '--bogus-flag'         # commander's own prose, on stderr
```

**Mechanism:** production `main.ts` installs the boundary on `unhandledRejection`/`uncaughtException` and wraps `parseAsync`, but there is no `exitOverride()` / `configureOutput()` anywhere in `src/` (only in tests). Commander therefore writes its own message and exits before `renderCliError` runs — so `usage_error` in the table is reachable via `CliUsageError`, but the *more common* usage failure (a typo'd flag, a missing required option, an unknown command) bypasses the record path entirely.

This is the same class of failure the PR exists to fix, and `llm` is the format the skill tells agents to pass unconditionally. Minimal fix in the PR's own idiom: `exitOverride()` plus `configureOutput({ writeErr })`, or a `CommanderError` branch that routes through `renderCliError` as `usage_error` (commander's `err.code` distinguishes `commander.unknownOption` / `commander.missingMandatoryOptionValue` / `commander.unknownCommand`).

## Finding 2 — `resolution_failed` is currently unreachable

`CliResolutionError` is defined (`errors.ts:124`) and handled (`errors.ts:280`), but **never thrown anywhere in `src/`** — the only other occurrences are the two `instanceof` checks. So the documented code cannot occur from a real run today. The sites that look like its intended producers print prose and *return* instead:

```
src/cli/resolve.ts:254    stderr(`No entity found matching "${symbol}".`);
                          return { resolved: false, ambiguous: false };
src/cli/resolve.ts:311    (same shape)
```

Worth either wiring one of those to throw it, or dropping the row from the table so the code list stays exactly what a caller can observe.

**Coupling note for that fix (PRE-EXISTING, not introduced here):** on the resource-not-found path I measured *two* reports — a record on stdout (from `ui.ts:168`, `error code=unresolved_target …`) **and** prose on stderr (from `resolve.ts:254`). Both predate this PR. But if `resolve.ts` is changed to throw `CliResolutionError`, the boundary's "prose is then skipped rather than printed alongside" guarantee meets a site that has already printed — so that edit must also remove the `stderr(...)` call, or the double report becomes the boundary's.

## Not isolatable at binary level

The `--` stop is pinned by the unit test and by reading `detectRequestedFormat`, not by my probes: `ix search -- --format llm` is rejected by commander ("too many arguments") before the detector matters, so I can't demonstrate the detector's behaviour end-to-end through the CLI. Flagging it as unit-pinned only, not runtime-verified.

PR #690, opened 2026-09-19
perf(llm): carry eight characters of an id, not thirty-six

KageBinary:
## What

Most `--format llm` renderers already truncated ids to eight characters.
`explain`, `entity`, `map`, `subsystems`, `read`'s candidate rows and
`context --diff`'s entity/claim/relationship records did not — so the command an
agent runs first to orient itself spent more bytes on identifiers than on the
labels beside them:

```
region id=44a6a94c-8526-3ac5-828f-47fb2dfeaab7 kind=module label=".github Workflows" level=1 files=10 parent=aa593667-7451-354a-bc06-daac60fede25 ...
```

```
region id=44a6a94c kind=module label=".github Workflows" level=1 files=10 parent=aa593667 ...
```

Measured on this repo's graph, `ix subsystems --format llm`:

| | bytes |
|---|---|
| today | 4,981 |
| this branch | 3,917 |
| saving | **21%**, for 38 ids |

`ix map --format llm` carries the same 30 regions, and a UUID tokenizes at
roughly 1.8 characters per token, so the token saving is larger than the byte
saving.

## Eight characters is a value, not an abbreviation

`/v1/resolve-prefix` accepts an eight-character prefix everywhere the CLI takes
an id — `ix entity 02caa1e4` works — so the short form can be handed straight
back. `--format json` still carries the full id for a program that stores one.

## One helper instead of six hand-rolled slices

`llmShortId` replaces six `?.slice(0, 8)` sites and covers the ones that had
none. It shortens only values that *are* opaque blobs — hex ignoring dashes, at
least 16 characters — so these are left whole rather than corrupted:

- a slug region id: `root`, `cli-commands` (the shape `docs/llm-format.md`'s own
  examples use)
- a prefixed claim id: `c-8f31a2`
- a short hex-looking word: `cafe`, `decade`

It is idempotent, and it shortens **both sides of a reference**, so `parent=`
still matches another row's `id=` and `src=`/`dst=` still resolve to an `entity`
record in the same bundle — a tree is still re-buildable from the records.

## Verification

  (106 files, 1,871 passed, 2 skipped) · `check-doc-parity` — all green
- new `llm-short-ids.test.ts`, 9 cases: the shortening, the three shapes left
  alone, nullish handling, and a scan of `entity`, `node`, `ref`, `conflict` and
  `region` renderer output against a full-UUID regex
- `llm-tier4`'s entity assertion updated — that record was the one still
  carrying a full id
- `docs/llm-format.md` gains the rule under **Wire format**
- byte counts above are from live runs against the local backend

Part of the CLI output-economy work (task C-11).


Alot1z:
Same coordination note — your `docs/llm-format.md` hunk is at line 20; #682's is at 11 and #688's is at 48. Three sections, one file, no design conflict. A maintainer one-liner naming an order would close the only real hazard in this group. Note that #676/#677/#678 are a separate, tighter collision (same two hunks, 94/114).

Alot1z:
Runtime verification of this PR at its own head (`11d9ad0e`), against a **real backend** rather than the suites. Hermetic stack (`arangodb:3.12.11` + memory-layer `1.0.28`, image `sha256:109e059a0acf`, isolated ports, sandbox `IX_HOME`); CLI = this branch's own build.

## The load-bearing dependency is real on the released backend

The doc comment says eight characters is "what `/v1/resolve-prefix` already accepts everywhere the CLI takes an id". That endpoint is the whole PR, so I probed it directly, HTTP-level, on the pinned release (`1.0.28` — the one `install.ps1` ships today, not a newer unreleased backend):

| Request | Response |
|---|---|
| `GET /v1/resolve-prefix/eec702b1` (8) | `200 {"id":"eec702b1-cf06-a650-e339-a904bf33b33c"}` |
| `GET /v1/resolve-prefix/eec7` (4) | `200 {"id":"eec702b1-…"}` |
| `GET /v1/resolve-prefix/<full uuid>` | `200 {"id":"…"}` (the client short-circuits this anyway) |
| `GET /v1/resolve-prefix/zzzzzzzz` | `404 {"error":"No entity matches prefix: zzzzzzzz"}` |
| `GET /v1/resolve-prefix/EEC702B1` | `404` — the route is **case-sensitive** |

So there is no backend-version coupling: prefix resolution is pre-existing release behavior, and this PR only changes what the CLI *writes*, not what it must be able to *read*. Worth stating in the description, because a reviewer can't tell that from the diff.

## The round-trip, measured

```
$ ix map --format llm | grep id=
id=5dd3f7ac
$ curl -s localhost:18090/v1/resolve-prefix/5dd3f7ac
{"id":"5dd3f7ac-bd66-3107-9a8b-60552782b4fc"}
$ ix entity 5dd3f7ac
entity id=5dd3f7ac kind=region name=Region path=ix:map rev=3 edges=2
edge pred=IN_REGION dst=25043d64
```

Emitted id → passed straight back → same entity, and the edge rows carry short ids too, so a caller can walk the whole answer without ever seeing 36 characters. "A value the caller can pass straight back" is now measured rather than asserted.

## Two things the suite would not catch

**1. The length choice is not pinned to the acceptance gate.** `SHORT_ID_LENGTH = 8` here and `/^[0-9a-f]{8,31}$/i` in `src/cli/resolve.ts:606` are the two halves of one contract, owned in two files, tested in neither. The test named "cuts an opaque id to a prefix the CLI still accepts" (`llm-short-ids.test.ts:17`) asserts the cut but never the acceptance; `resolvePrefix` appears in tests only as a mocked method name (`explain-pipeline.test.ts:678,697`). A one-line invariant costs nothing and cannot rot:

```ts
expect(/^[0-9a-f]{8,31}$/i.test(llmShortId(UUID)!)).toBe(true);
```

That fails the moment either half moves, without needing a backend.

**2. An ambiguous prefix loses its diagnostic.** The backend *does* fail safe — `GET /v1/resolve-prefix/e` → `200 {"error":"ambiguous","matches":[…]}`, and `client.resolvePrefix` turns that into a loud `Ambiguous prefix "e" — matches: …`. But `resolveTarget` reaches it inside a bare `catch {}` (`src/cli/resolve.ts:605-621`), so the throw is discarded and control falls through to name resolution: the user is told no entity matches their name, and the candidate list the backend just sent is thrown away. Same swallow hides the case-sensitive 404 above, so an uppercased `EEC702B1` reads as "not found" rather than "wrong case".

How reachable is it? Rare — I enumerated ids through the endpoint and found **zero** 8-hex collisions (160 sampled ids; a 7k-node graph expects ~6e-4). But the failure is silent-by-construction once it happens, and the probability tracks graph size. Two cheap options: re-throw when the backend said `ambiguous` (an ambiguous 8–31-hex string that is also a real symbol name is not a case worth protecting), or keep the diagnostic as a warning before falling through.

Smaller notes, same area: the ambiguity payload caps at 10 matches with no total (`e` → 10, `e0` → 10, `e00` → 3), so the message can never say "34 candidates"; and because the gate is `{8,31}`, "eight characters" is the *floor* — a seven-character prefix is a name, not an id. One line in `docs/llm-format.md` saying the accepted range is 8–31 would stop a reader assuming any-length prefixes work.

PR #691, opened 2026-09-19
feat(ingest): index module-level constants, as constants (Ix#679)

### LINKED ISSUE #679: A top-level const is not indexed as an entity in TypeScript/JavaScript
## Summary

In TypeScript and JavaScript, ingestion only indexes a `const` when its value is
a function. A `const` bound to anything else — an object, an array, a `Set`, a
string, or a call result such as `z.object({...})` — produces no graph entity, so
it cannot be searched for, cannot be a context target, and never appears in a
bundle. The same gap exists in Python, Go, C, C++, Kotlin, PHP, Ruby, Elixir,
Zig, Haskell and Lua; Rust, Scala, Java, C#, Swift and CSS already capture
constants.

This is an extraction gap, not a retrieval one — no amount of ranking or budget
work in `ix context` can surface a node that ingestion never created.

## Where it comes from

`core-ingestion/src/queries.ts`, `TYPESCRIPT_QUERIES` (lines 27–48). The only
rules that match a `lexical_declaration` require a function value:

```
(lexical_declaration
  (variable_declarator
    name: (identifier) @name
    value: (arrow_function))) @definition.function

(lexical_declaration
  (variable_declarator
    name: (identifier) @name
    value: (function_expression))) @definition.function

(export_statement
  declaration: (lexical_declaration
    (variable_declarator
      name: (identifier) @name
      value: (arrow_function)))) @definition.function

(export_statement
  declaration: (lexical_declaration
    (variable_declarator
      name: (identifier) @name
      value: (function_expression)))) @definition.function
```

There is no `@definition.const` rule at all. `JAVASCRIPT_QUERIES` has the same
shape.

Definition kinds captured per language today:

| Language | `@definition.*` kinds |
|---|---|
| TypeScript | class, function, interface, method |
| JavaScript | class, function, method |
| Python | class, function |
| Go | function, interface, method, struct |
| C | enum, function, macro, struct, typedef, union |
| C++ | class, enum, function, macro, method, namespace, struct, template, typedef, union |
| Kotlin | class, enum, function, interface, property, type |

Rust (`const_item`), Scala (`val_definition`/`var_definition`), Java, C#, Swift
and CSS do emit `@definition.const`, so the behaviour is inconsistent across the
26 supported languages as well as incomplete.

## Effect, measured

Every `export const` in `ix-cli/src` is a non-function value — all 16 of them —
so none of Ix's own module-level constants are in Ix's own graph.

Probed against a full graph of this repo at `5e611d17` (backend 1.0.30, CLI
0.10.5), using the 24 ground-truth symbols from our retrieval benchmark:

```
4 of 24 ground-truth symbols absent: SUPPORTED_EXTENSIONS, BUDGETS,
                                     contextBundleSchema, BUNDLE_SCHEMA
```

Every one of the 20 function-valued symbols resolves. Every one of the 4 misses
is a `const` bound to a non-function value:

| Symbol | Declaration |
|---|---|
| `SUPPORTED_EXTENSIONS` | `ix-cli/src/cli/supported-extensions.ts:11` — `export const SUPPORTED_EXTENSIONS = new Set<string>([` |
| `BUNDLE_SCHEMA` | `ix-cli/src/cli/context-bundle-schema.ts:9` — `export const BUNDLE_SCHEMA = "ix-context-bundle/1";` |
| `contextBundleSchema` | `ix-cli/src/cli/context-bundle-schema.ts:37` — `export const contextBundleSchema = z.object({` |
| `BUDGETS` | `ix-cli/src/cli/commands/context.ts:51` — `const BUDGETS = [` |

`ix search SUPPORTED_EXTENSIONS` returns zero results against a graph that
contains `supported-extensions.ts`, and `ix context` on that file cannot name the
identifier the file exists to declare. `ix search BUDGETS` returns
`BudgetSnapshot`, `clampBudgets`, `formatBudgets`, `parseRequestedBudgets` —
neighbours of the thing asked for, not the thing.

In the benchmark this puts a hard ceiling on symbol recall for six tasks
regardless of how good retrieval gets:

| Task | Symbol-recall ceiling |
|---|---|
| `smoke-extension-list` | 0.00 |
| `nav-context-bundle-schema` | 0.00 |
| `impl-context-budget-flag` | 0.00 |
| `longhorizon-context-budget` | 0.00 |
| `impl-add-extension` | 0.50 |
| `xfile-extension-parity` | 0.50 |

## Reproducing

```bash
ix map --silent
ix search SUPPORTED_EXTENSIONS          # count=0
ix search parseBudgetOption             # hits — same file, function value
ix context ix-cli/src/cli/supported-extensions.ts --format llm
                                        # the file is there; the identifier is not
```

## The trade-off any fix has to make

The obvious rule — capture every `variable_declarator` — is too broad. In
`ix-cli/src` alone:

- 5,085 `const` declarations at any scope
- 259 at module scope
- 16 exported

A blanket rule multiplies the entity count of a single package by ~20x with
function-local bindings, which would hurt ranking and bundle budgets everywhere.
Restricting to **module scope** is the version that pays for itself: it picks up
all four missing symbols at a cost of 259 nodes, and it does not depend on the
declaration being exported — note that `BUDGETS` is module-scope but *not*
exported, so an export-only rule would still miss it.

Suggested shape, matching the existing style:

```
(program
  (lexical_declaration
    (variable_declarator name: (identifier) @name))) @definition.const

(program
  (export_statement
    declaration: (lexical_declaration
      (variable_declarator name: (identifier) @name)))) @definition.const
```

with the function-valued rules kept ahead of it so an exported arrow function
still classifies as `function`, not `const`.

Open questions for whoever picks this up:

1. Should `var`/`let` at module scope be included, or `const` only? `const` only
   is the smaller, higher-signal change.
2. Same treatment for Python module-level assignments (`UPPER_CASE = ...`), Go
   `const_spec`, and C++ `constexpr`? Those are the same gap in other languages
   and are probably worth separate follow-ups rather than one sweeping PR.
3. `DEFINITION_KIND_MAP` in `core-ingestion/src/index.ts:312` already maps
   `definition.const` — to `'function'`. So the plumbing exists and a new query
   rule needs no other change to land, but constants arrive in the graph as
   `kind: function`, which is wrong for `ix search --kind function`,
   `ix inventory --kind function` and `ix rank --kind function`. That mapping is
   already live for Rust, Scala, Java, C#, Swift and CSS constants today. Adding
   a real `const` node kind is a bigger change than this issue; worth deciding
   whether to do it here or accept `function` for now and split it out.

KageBinary:
Closes #679.

## The gap

In TypeScript and JavaScript a `const` was indexed only when its value was a
function. Bound to an object, an array, a `Set`, a string, or a call result such
as `z.object({...})` it produced no entity at all — not searchable, not a
context target, never in a bundle. Every `export const` in `ix-cli/src` is a
non-function value, so none of Ix's own module-level constants were in Ix's own
graph.

Measured against the 24 ground-truth symbols of the retrieval benchmark, the
four a full graph of this repo could not resolve were all of this shape. After
this change, parsing all 199 files of `ix-cli/src`:

```
SUPPORTED_EXTENSIONS: constant ix-cli/src/cli/supported-extensions.ts:11-39
BUDGETS:              constant ix-cli/src/cli/commands/context.ts:51-64
contextBundleSchema:  constant ix-cli/src/cli/context-bundle-schema.ts:37-109
BUNDLE_SCHEMA:        constant ix-cli/src/cli/context-bundle-schema.ts:9-9
```

## The cost, which is the whole design question

| rule | entities over `ix-cli/src` |
|---|---|
| today | 2,433 |
| **module scope only (this PR)** | **2,696 (+263, +10.8%)** |
| every `variable_declarator` | ~7,500 (5,085 `const`s at any scope) |

So the rule is anchored under `(program ...)`. Three further narrowings, each
verified against the parser:

- `"const"` as a literal keeps `let` out — mutable module state is not what an
  agent searches for by name
- `name: (identifier)` keeps `const { a, b } = obj` out — it binds no single
  name worth indexing
- export is **not** required: `BUDGETS` is module-scope and unexported, and an
  export-only rule would still have missed it

The declarator's value is captured so `index.ts` can drop the arrow-function and
function-expression cases the four function rules above it have already claimed.
Without that, an exported arrow function arrives twice, once per kind.

## `constant`, not `function` (decision D10)

This implements the **recommended but unconfirmed** decision D10: a new
`constant` entity kind rather than reusing `function`.

`definition.const` was already mapped to `'function'`, and already live for Rust
`const_item`, Scala `val`/`var_definition`, Java, C#, Swift and CSS keyframes —
so every one of those has been arriving as a function and inflating
`--kind function` on `search`, `inventory` and `rank` for as long as they have
been supported. One line fixes all of them.

(CSS keyframes coming out as `constant` is not obviously right either, but it is
closer than `function`; worth a separate look.)

## Two consequences handled, not left to be discovered

- **`findEnclosingFunction` keyed on `function`/`method`.** Giving constants
  their own kind would have re-coarsened every edge attributed to one — caught
  by `queries.scala.test.ts`, which asserts a `var` holds its own `REFERENCES`
  edge rather than its class (#557). A constant's initializer is a body: the
  `z.object(...)` in `const schema = z.object({...})` is a call made by
  `schema`. Renamed `findEnclosingDefinition` so the name says so.
- **The CLI resolver ranks by a kind list** that had no `constant` in it, which
  would have sorted a constant behind every config entry and heading sharing its
  name. Added there and to both `STRUCTURAL_KINDS` sets.

## Not in scope

Python module-level assignments, Go `const_spec` and C++ `constexpr` are the
same gap in other languages; the issue asks for those as separate follow-ups,
and this PR keeps to TS/JS.

Noted while here, **not** fixed: `export const handler = () => 1` already
matched two of the four function rules and produced two identical `function`
entities. Same name, same lines, so `patch-builder` folds them onto one node id
— wasted work rather than a wrong graph.

## Verification

- `core-ingestion`: `npx vitest run` — 38 files, 391 passed, 1 skipped
- `ix-cli`: `npx vitest run` (105 files, 1,862 passed, 2 skipped) ·
  `npm run lint` · `npm run typecheck` · `npm run knip` ·
  `node test/parser.smoke.mjs` · `check-doc-parity` — all green
- two new cases in `queries.typescript.test.ts`: the four benchmark symbols
  arriving as `constant`, function values staying `function` and not duplicating
  as `constant`, and `local` / `mutable` / destructured names staying out; plus
  a constant getting the line range of its own declaration rather than the file's
- entity counts above are from parsing the real tree, not a fixture

Part of the retrieval-quality work (task X-3).


Alot1z:
Runtime verification of this PR at its own head (`06d2e47`), hermetic backend 1.0.28, sandbox `IX_HOME`, small multi-file TS corpus (mixed functions + module-level constants), probed through `ingest → stats → search → explain → inventory`.

## What verifies

Constants capture works end to end for freshly parsed files:

| Probe | Result |
|---|---|
| `ix inventory --kind constant` | every module-level `const` in newly parsed files: `TOP_CONST`, `NESTED_CONST`, `MIXED_CONST`, plus fixture/generated ones — kind `constant`, not `function` |
| `search FIXTURE_ROWS` / `search GENERATED_TS` (raw `POST /v1/search`) | `kind: constant` rows returned |
| `stats` | `constant: N` counted as its own kind |
| The PR's own A/B claim | `export function` in the same file no longer produces a duplicate `constant` (the arrow-function/function-expression guard) |

## One measured gap: existing files never gain the new extraction — and `--force` doesn't recover them

The corpus predates this build: `src/util/math.ts` was first ingested on a pre-#691 build, so its `export const VERSION = '1.0.0'` was captured under the old rules. On the #691 build:

```
ix ingest .                    # math.ts mtime-clean → skipped
ix inventory --kind constant   # FIXTURE_ROWS, GENERATED_TS, LATE_FIXTURE — no VERSION
```

A byte-identical copy of the same file (`cp math.ts math2.ts`) ingested in the same run **does** yield `VERSION` as `constant` — so extraction itself is correct; the gap is only that an unchanged, already-ingested file is never re-extracted. Expected. But then:

```
ix ingest . --force            # full re-parse of all 10 files, 0 skipped
ix inventory --kind constant   # still no VERSION from math.ts
```

`--force` re-parses everything, yet the entity does not appear. Touching the file (`>> math.ts`, forcing an actual content change) finally produces it:

```
ix ingest .                    # processed: 2 files (math.ts + app.ts)
ix inventory --kind constant   # VERSION  src/util/math.ts   ← appears
```

So the constant-materialization gap survives a forced re-parse and is only closed by a content change. The likely owner is patch idempotency: `ingest` records `idempotentPatches` and the backend keeps an `idempotency_keys` collection — if identical parse output produces an identical patch, the backend dedupes it and the *new extraction rules never express themselves as a patch*, because the old parse result is what gets re-derived and compared. `--force` forces re-parsing, not re-diffing against the old rules.

Why this matters to this PR specifically: every existing Ix user has graphs ingested under the old mapping. Upgrading to this build does not add `constant` entities for any file that doesn't change — the feature silently applies only to new/edited files, and the remedy the command itself prints for other staleness problems (`ix ingest <root> --force`, visible in the stitch-refusal message) does not fix it. Same shape as the self-replacement concern: the released artifact must actually re-derive what the new code says.

## Smallest reproduction

1. Ingest any repo on a pre-#691 build (or `git stash` the mapping change).
2. Check out this PR, `ix ingest .` → no `constant` entities for unchanged files.
3. `ix ingest . --force` → still none.
4. Append a comment line to one file, `ix ingest .` → that file's constants appear.

## Suggested direction (maintainer's call)

If the diagnosis is right, the fix belongs where patch identity is computed: include the extractor/mapping version in the patch idempotency key (or invalidate on it), so a CLI whose extraction rules changed re-emits patches for unchanged files. A one-line `EXTRACTOR_VERSION` in the key would make every future extraction-rule change self-healing on the next ingest, `--force` or not.

PR #695, opened 2026-09-19
test(context): gate the bundle on what share of it names a real place

KageBinary:
> **Stacked on #694** (`feat/context-range-first`) → #693 → #692 → #678 → #677
> → #676 → #675, **and it merges #680**. That merge is deliberate and is the one
> thing worth reading first.

## Why #680 is in this branch

The gate asserts that no conflict row reaches the evidence list — which is
exactly what #680 delivers. Without it this branch's CI is red by construction.
#680 is independent of the context stack and merges cleanly to `main` on its
own; once it lands, the merge commit here becomes a no-op.

## What the gate measures

The bundle is the product and its quality was never measured, so the regression
that halved it shipped in silence: seventeen `conflict — <uuid> vs <uuid>` rows
took the evidence budget from the members and imports that answer the question,
and the useful-char share of a recorded bundle was **4%**.

Four target shapes, each asserted on the three ways a bundle goes wrong:

| shape | why it is here |
|---|---|
| a file whose members are the answer | the `config.ts` case from the benchmark |
| a symbol whose callers are the answer | the commonest `ix context` call |
| an entry point whose imports are the answer | 6 of 19 benchmark answers are a direct import |
| a target whose context is mostly conflict reports | **the one that broke** |

1. at least half its bytes name somewhere a caller can open
2. no conflict row reaches the evidence list
3. it stays inside the budget it reports, and ends with a runnable read

Fixtures rather than a live graph, so it runs in CI with no backend, and
`clampBudgets({})` rather than generous budgets, so it measures what a caller
actually gets.

## It fails without the fix — verified, not assumed

| fixture | this branch | with #680's `rankEvidence` change reverted |
|---|---|---|
| a file | 92.0% | 92.0% |
| a symbol | 90.7% | 90.7% |
| an entry point | 92.4% | 92.4% |
| **mostly conflict reports** | **89.2%** | **31.0% — 2 of its 4 assertions fail** |

That is the acceptance criterion for this task: the gate has to fail on the
behaviour it exists to catch.

## Verification

- `npx vitest run` — 1,918 passed, 2 skipped · `npm run lint` ·
- 16 new cases (4 fixtures × 4 assertions)

Part of the retrieval-quality work (task X-8).


Alot1z:
Runtime verification of the context stack (this PR + #694 → #693 → #692 → #678 → #677 → #676 → #675, tip `2ce3c75e73`), executed against a **real backend** rather than the test suites.

**Setup (hermetic):** `arangodb:3.12.11` + `ix-memory-layer:latest` (release 1.0.28) on isolated ports, dedicated volume, `IX_HOME` pointed at a fresh sandbox, CLI = this stack's build (0.10.5). Tiny 3-file TS corpus ingested via `ix map`, then the probes below. Stack and volume torn down afterwards.

**What was exercised at binary level, and what happened:**

- **Budget honesty (#692):** `--max-tokens 10` clamps and the JSON block reports the clamped `maxTokens: 500` — the effective budget is what's reported, not what was asked. `--max-tokens 500 --max-chars 3000` is refused: exit 1, stderr carries `mode_conflict`, stdout payload-free.
- **Truncation report (#693):** with a tight `--max-chars 1500`, the text closer names exactly what was cut and the recovery (`Note Truncated: … cut: 7 relationships, 4 neighbouring definitions … or raise --max-tokens`); with a huge budget the hint is absent and the JSON `truncation` block carries no `cut` array. Silence when nothing was cut, exactly as the suites claim.
- **Resume display (#694):** `--save probe-694` (10 entities / 4 relationships / 4 evidence), then `--list` shows the identical counts and truncated block, then `--resume` with the backend endpoint **dead** (`127.0.0.1:9`) renders the same banner + record without needing the backend. Banner/report/record agreement confirmed. `--diff` against the unchanged graph: coherent budget echo (`saved`/`requested`/`effective`), `-0 +0` everywhere.
- **Usefulness gate (#695):** evidence JSON (3120 B) stays inside the reported budget (`maxChars: 3210`); the text bundle closes with runnable `ix read` recommendations; zero conflict rows reach the evidence list (clean graph, so that invariant is vacuously exercised here — the conflict-shape coverage lives in the suite).
- **`--list` / `--resume` happy paths:** first binary-level exercise; both worked as documented.
- **`--format llm`:** the record stream agrees with the JSON honesty block — one summary record (counts + truncated counters), a `diagnostic code=bundle_truncated` line carrying the same cut summary, then scored evidence rows.

No divergence from the suite claims was found on any probed path; the UNVERIFIED-at-runtime flag on this stack's assessments resolves to **measured**. Everything ran in a sandbox home — no real profile state was touched.

PR #696, opened 2026-09-19
feat(cli): give every agent-facing row a place, and stop naming it three times

KageBinary:
> **Stacked on #688** (`fix/honest-truncation`), whose `sliceEdgeResults` these
> rows go through, **and it merges #676** for the located facts `ix explain`
> needs — `topCallerRefs`, `topDependentRefs`, `memberRefs` and the
> `EntityLocation` type. #676 is the lowest link of the context stack that
> provides them, so this pulls in #675 and #676 and nothing above. Both merge
> cleanly to `main` on their own and the merge disappears once they land.

## The cost of a bare name

An agent handed `caller registerCallersCommand` has to spend a turn finding it.
A turn is ~30k tokens. A path and a span are ~30 bytes, and the graph node they
come off was already in hand on every one of these rows.

```
ix callers   ref name=X kind=function id=7f7f6640 path=…/callers.ts lines=17-145
ix entity    entity id=… kind=function name=relativePath path=…/format.ts lines=15-27
ix impact    member name=X kind=function callers=3 path=…/read.ts lines=154-170
ix search    function  8ebf63f9  relativePath  ix-cli/src/cli/format.ts:15-27
ix explain   uses name=X rel=caller,dependent path=…/callers.ts lines=17-145
```

All live output. One helper, `rowLocation`, reads the path and span off a node;
a **file** gets no span, because its span is the file — the rule the context
bundle already uses. `--format json` gets `lineStart`/`lineEnd` as numbers; llm
and text get `17-145`.

## `ix explain` said the same thing three times

A caller is nearly always also a dependent, so on this repo `relativePath` gave:

```
caller name=registerCallersCommand
caller name=renderReadAmbiguityLlm
caller name=outputResult
dependent name=registerCallersCommand
dependent name=renderReadAmbiguityLlm
dependent name=outputResult
used_by text="Used by registerCallersCommand, renderReadAmbiguityLlm, outputResult, and 1 other."
```

Seven lines, three names, **no paths**. Now:

```
uses name=registerCallersCommand rel=caller,dependent path=ix-cli/src/cli/commands/callers.ts lines=17-145
uses name=renderReadAmbiguityLlm rel=caller,dependent path=ix-cli/src/cli/commands/read.ts lines=120-148
uses name=outputResult rel=caller,dependent path=ix-cli/src/cli/commands/read.ts lines=154-170
```

296 bytes → 306. **Ten bytes** for three complete coordinates, against the
locate turn each bare name used to cost.

The generated `used_by` sentence goes, because it is those names a third time
and `edges callers=4` already carries the "and 1 other". A hierarchy-derived
usage narrative says something the rows do not, so that one stays — `render.ts`
marks which is which (`usedByIsNameList`) rather than letting the renderer guess
from the text. `--format text` is unchanged: a person reading prose wants the
sentence.

## Verification

- `npx vitest run` — 109 files, 1,904 passed, 2 skipped · `npm run lint` ·
- new `row-locations.test.ts`, 14 cases: `rowLocation` on a symbol and on a
  file, a non-integer line number ignored, one-line collapse, edge rows in all
  three formats, a search text row, `ix entity`'s span, the caller/dependent
  merge, the dropped and the kept `used_by`, located member rows, and a row
  that still renders when the facts carry no location
- every output block above is a live run against the local backend

Part of the CLI output-economy work (task C-5).


Alot1z:
Independent verification on this PR's own build (head `f2c41b9`, base `5e611d17`, hermetic backend 1.0.28, sandbox `IX_HOME`, multi-file TS corpus with real call edges).

**The location contract verifies at runtime:** relation rows now answer "where" in the agent format — `callees runAll --format llm` → `ref name=compute kind=function id=8bd9deaf path=src/beta.ts lines=1-3`; `callers compute --pick 2` → `ref name=runAll … path=src/main.ts lines=2-4`. Line spans are the real spans, not file-level stubs, and `docs/llm-format.md` ships updated in the same PR so the contract is written where agents read it.

**One adjacent gap, marked as outside this PR's files** (`ui.ts` is not in the diff — flagging it here only because this PR's theme is exactly "every agent-facing row has a place"):

The ambiguous-target path is the one agent-facing surface still rendering candidates without a place. Measured on two same-named exports:

- `--format llm` → `error code=ambiguous_target message="Ambiguous symbol \"compute\"." candidates=1:compute,2:compute` — indistinguishable rows; an agent cannot pick without re-querying each by id.
- `--format json` carries `candidates[]` with ids (good), but no `path` either.
- The **text** renderer already prints `candidate.path` when present (`ui.ts:198` — ` in src/alpha.ts`). So the human format can disambiguate and the agent formats cannot — the inverse of every other surface this PR fixes.

Smallest fix where it properly belongs: include `candidate.path` (and keep the id) in the llm `candidates` value (`ui.ts:185-187`) and in the JSON `candidates[]` rows. Happy to file it separately if you'd rather keep this PR's diff as-is.

Probe transcript available on request.

PR #697, opened 2026-09-19
feat(read): stop at 400 lines, and say where the next 400 are

KageBinary:
> Implements **decision D4 as recommended, but unconfirmed**: a 400-line default
> cap on `ix read <file>`, with a cursor and an `--all` escape hatch. Behaviour
> change — for a minor release.

## The bug

`ix read <file>` had no bound at all, on the one target whose size a caller
cannot see before asking. Recorded reads ran **20,000 to 53,000 tokens**, and
one benchmark run read the same file in eleven slices because the first arrived
from the wrong anchor and cost too much to repeat.

## After

```
$ ix read ix-cli/src/cli/commands/context.ts --format llm
file path=ix-cli/src/cli/commands/context.ts line_start=1 line_end=400 target=file stale=true truncated=true total_lines=1734 next=ix-cli/src/cli/commands/context.ts:401-800
content lines=400
<400 lines>

$ ix read ix-cli/src/cli/commands/context.ts --all --format llm
file path=… line_start=1 line_end=1734 target=file stale=true
content lines=1734
```

1,734 lines → 400, with the next page one command away.

400 is roughly 4-6k tokens of TypeScript: enough to hold a whole small module,
small enough that a wrong guess is cheap to abandon.

**Never capped:** a range the caller typed (`ix read a.ts:1-900` — they said
what they wanted), a symbol target (it is its own span), and `--all`.

For a person, the source still goes to stdout and the note to stderr:
`400 of 1734 lines — next: ix read src/big.ts:401-800, or --all for the whole file`.

## What I did not do, and why

C-3 also asked for **symbol-before-filename on an extension-less target**
(`ix read config` meaning the symbol rather than the file). It is not in this
PR.

The only way to know a symbol exists is to ask the graph, and
`read-local-first.test.ts` asserts **zero round trips** for exactly that case —
`expect(searchCalls).toEqual([])` — with a docblock recording the 60-80s
incident that guarantee came from. I implemented the flip, saw that test fail,
and took the failure as the answer rather than editing it.

The cap removes the cost that motivated the flip: a whole-file read is now 400
lines with a cursor, not 20-53k tokens. What is left is precision, and it is
worth one explicit decision — the trade is one `nameOnly` search on
extension-less targets that match a local file — rather than a quiet reversal of
a measured one. The reasoning is recorded in `read.ts` where the order is
decided.

## Verification

- `npx vitest run` — 105 files, 1,862 passed, 2 skipped (plus 7 new) ·
  `npm run lint` · `npm run typecheck` · `npm run knip` · `check-doc-parity` ·
  `check-links` — all green
- new `read-line-cap.test.ts`, 7 cases: the cap and its cursor, `--all`, a typed
  range never capped, a file that fits left alone, the json fields, the stderr
  note with all 400 lines still on stdout, and no next page offered from the
  last one
- `flags.md` gains the `--all` row and a behaviour note
- output above is from live runs

Part of the CLI output-economy work (task C-3, milestone M3).


KageBinary:
Note on the MCP side: `ix_read` does not expose `--all`, and does not need to for this to be usable — the cap's own `next=ix-cli/src/cli/commands/context.ts:401-800` cursor is a target `ix_read` already accepts, so paging over MCP works with no new parameter. Adding `all` to the tool is a one-line follow-up if it turns out anyone wants the whole file in one call.

Alot1z:
Runtime verification on this PR's own build (head `0b25d1d`, sandbox `IX_HOME`, hermetic backend 1.0.28, real 4,002-line file).

**What verifies:** the cap and the cursor work end to end. Default read of a 4,002-line file → `lineEnd: 400, truncated: true, totalLines: 4002`, and `next` is runnable *as emitted* (absolute form round-trips through `ix read` — one call to page 2, confirmed). Explicit `1-1000` is never capped (1,000 content lines, no flag); the last page (3901-4002) is clean; `--format llm` carries `truncated=true total_lines=4002 next="…"`; the text footer lands on stderr with the same recovery. Suite covers the ordering (--all, typed ranges, cursor-in-json).

**Two boundary findings in the new cap code itself** — both from one line, `read.ts:107` (`raw.split("\n")` keeps the empty element after a trailing newline), and both at boundaries the suite misses because its "leaves a file that fits alone" fixture sits below the cap:

1. **Phantom line in every count.** A file with 401 real lines + trailing newline reports `totalLines: 402` (measured), and `next` offers `…:401-402` — a range whose second line is the empty tail. Every `totalLines`/`lineEnd`/`next` on a newline-terminated file is computed against N+1.

2. **A file that exactly fits is falsely flagged truncated, with a junk cursor.** A 400-line file ending in a newline → `truncated: true` + `next: …:401-401`. Nothing was cut (content ends at 400), but an agent reading the flags does one wasted fetch — the phantom page, which I confirmed returns `content: ""` — and one that trusts `truncated` believes more exists. The exact boundary case the cap exists to handle correctly.

**Smallest fix:** drop the trailing empty element before computing anything (or equivalently `totalLines = raw.endsWith("\n") ? lines.length - 1 : lines.length`) — one line clears both findings; plus one suite case: exactly-400-lines-with-trailing-newline must yield `truncated: false`/no `next`. Full probe transcript available if useful.

PR #698, opened 2026-09-19
feat(depends,trace): bound an unasked-for traversal, and report its end honestly

KageBinary:
> Behaviour change — for a minor release, with the rest of the M3 batch.

## The bug

```ts
const MAX_NODES = Infinity;
const DEFAULT_MAX_DEPTH = Infinity;
```

Both, in both commands. `ix depends` on a hub walks the whole upstream cone —
thousands of nodes and 5-50 KB of output for a question usually answered by the
first level or two, and the caller pays for all of it before seeing any of it.

Now depth **3** and **100** nodes. Depth 3 is far enough to show a path through
an intermediate; 100 nodes is more than fits on a screen. `--depth` and `--cap`
still take anything, including a larger number.

## The cut is ranked

A cap applied to an unranked level keeps whatever the graph happened to return
first. Each level is now ordered followable-first:

1. a name **and** a path
2. a name alone
3. a dangling raw id — nothing can follow it

Stable within a band, and the predicate order (CALLS before IMPLEMENTS) is
untouched.

## A depth stop is not a truncation

`if (depth > maxDepth) { truncated = true; ... }` was harmless while the default
was infinite. At depth 3 it fires on most leaf branches and claims nodes were
lost when the walk simply reached its bound — which is the dishonest half of
adding a bound, and the opposite of what this series is for.

| field | means |
|---|---|
| `truncated=true` | the node cap dropped nodes — the only case anything can know that |
| `depth_limited=true` | the walk stopped descending; there may or may not be more |

```
depends target=relativePath ... nodes=8 depth=3 depth_limited=true
diagnostic code=depth_limited message="Stopped descending at depth 3; there may be more below. Raise --depth to look further."
```

vs, when the cap is what cut:

```
diagnostic code=truncated message="Node cap of 100 reached; nodes were dropped. Raise --cap, or start from a narrower target."
```

Each names the one flag that lifts *that* bound, not both.

## About the numbers

**The bound rarely engages on this repo's graph** — 957 edges in total, and the
largest cone here is eight nodes, so a before/after byte count from this machine
would be noise. The 5-50 KB figure is the spec's, measured on a 1.16M-node
graph. So the bound's behaviour is covered by tests against a synthetic
2^depth fan-out rather than by a local number I cannot honestly produce.

## Verification

- `npx vitest run` — 1,870 passed, 2 skipped · `npm run lint` ·
- new `traversal-bounds.test.ts`, 8 cases: the depth default, the node default,
  a caller's own bounds still winning, a depth stop not reported as a
  truncation, the cap keeping a followable node over a dangling one, and both
  hint wordings reaching the record stream with the right code
- `depends-semantics.test.ts`'s source-grep assertion updated to the behaviour
  it is checking for
- `flags.md` gains the real defaults and a note on the two end states

Part of the CLI output-economy work (task C-4, milestone M3).


Alot1z:
Runtime verification of this PR at its own head (`2611483`), against a real backend. Hermetic stack (`arangodb:3.12.11` + memory-layer `1.0.28`, isolated ports, sandbox `IX_HOME`); each fixture mapped by this build. Two corpora: a hub with a wide cone (30 nodes for `depends`, 61 for `trace`) and a depth-5 call chain.

## `depends`: the contract holds, in every format

```
$ ix depends hub --cap 5
depends target=hub kind=function target_id=eb32f4df semantics=downstream_dependents nodes=5 depth=2 truncated=true
diagnostic code=truncated message="Node cap of 5 reached; nodes were dropped. Raise --cap, or start from a narrower target."

$ ix depends hub --depth 1
depends target=hub … nodes=30 depth=1 depth_limited=true
diagnostic code=depth_limited message="Stopped descending at depth 1; there may be more below. Raise --depth to look further."
```

Both halves of the distinction behave as described. `--cap 5` mints `truncated` plus the hint; `--depth 1` mints `depth_limited` instead and claims no loss — which is exactly the case that would have been a lie under the old `Infinity` defaults. A walk that simply finished says neither: `--depth 10` → `nodes=30`, no flags. `--cap 500 > --cap 100` shows the bound is still caller-overridable. Text shows the hint too (last line: `Node cap of 5 reached; nodes were dropped. …`).

## `trace`: the same honesty exists on one of its paths

| invocation | rows | what it reports |
|---|---|---|
| `trace hub` (default, both) `--cap 5` | 5 of 61 | **nothing** — llm, json and text |
| `trace hub --upstream --cap 5` | 5 | `truncated=true` + the same diagnostic (llm, json) |
| `trace hub --depth 1` | 31 of 61 | **nothing** |
| `trace hub --upstream --depth 1` | 31 of 61 | **nothing** |

The default form is the one that reads as "show me everything connected to this", and on a hub the cap cut 61 nodes to 5 with output that looks complete:

```
$ ix trace hub --cap 5 --format llm
trace mode=directional target=hub kind=mixed direction=both depth=3 target_id=eb32f4df up_nodes=5 up_depth=3 down_nodes=0 down_depth=1
   ← 56 rows gone, and this line is the only place the number appears; nothing says anything was dropped
$ ix trace hub --cap 5 --format json     → keys: mode,target,direction,kind,depth,upstream,downstream
$ ix trace hub --cap 5 --format text     → no hint
```

Three places drop it, and they are the same omission:

- `trace.ts:240` — the walker returns `{ tree, truncated, depthLimited, nodesVisited, maxDepthReached }`.
- `trace.ts:395-398` — `renderTraceBothLlm(…, up: { tree, nodesVisited, maxDepthReached }, down: { … })`. The two flags are not in the parameter, so the default path cannot report them on principle.
- `trace.ts:616-628` — the JSON literal for the same path emits `summary: { nodes_visited, max_depth }`, same omission. The text renderer has no trailer either.
- And where the flags *are* consulted (`trace.ts:424` and `710`), `depthLimited` is passed as a hardcoded `false`: `traversalHint(maxDepth, maxNodes, { truncated: true, depthLimited: false })`. So `depth_limited` can never appear for `trace` — although `trace.ts:193` computes it and `240` returns it.

Same contract, two commands: `depends --depth 1` says "there may be more below", `trace --depth 1` returns 31 of 61 nodes and says nothing, and `trace --cap 5` returns 5 of 61 and says nothing. The description frames the node cap as "the only case anything can know that" — for `trace` in its default form, that case is the one that stays quiet.

Suggested minimal fix, mirroring what `depends` already does at `depends.ts:219` / `224-227`: thread `truncated`/`depthLimited` into `renderTraceBothLlm`'s `up`/`down` params, add them to the directional JSON literal's `summary`, pass the real `depthLimited` at 424 and 710, and print `traversalHint` as a trailer in the text path. No new machinery — the helper and both call sites already exist.

*Scope of this comment: I exercised the bounds and the honesty reporting. I did not independently verify the followable-first ranking claim (`name+path` → `name` → dangling id, stable within bands) — my fixture's levels are homogeneous, so ordering would not have shown itself either way.*

PR #699, opened 2026-09-19
feat(ingest): let a repo say which of its files are not the point

KageBinary:
## What

A fixture-heavy repository puts its fixtures in the graph, the map and every
search result. Django's test tree and Vite's playground are most of what those
repositories contain by file count, and none of it is what someone asking about
the repository means.

```bash
ix ingest . --exclude playground --exclude '*.test.ts'
```

```
# .ixignore
# generated API surface
docs/generated/
**/__fixtures__/
```

## The half that actually matters

`IGNORE_DIRS` has only ever applied to the **filesystem walk**. `git ls-files`,
which is the discovery path in any git repository, never saw it — which is why
a git repo's `tests/` tree reaches the graph today. The new matcher runs on
both paths.

Directories are tested before the walk descends, so an excluded tree costs one
check rather than one per file inside it. The count is reported in the summary
(`excluded: 12 paths (--exclude / .ixignore)`, `filesExcluded` in json):
"that file is not in the graph" and "you excluded that file" are different
problems with different fixes.

## Matching: a deliberate subset of .gitignore

Supported: `#` comments, blank lines, `*` (stops at a separator), `?`, `**`
(crosses separators), a leading `/` to anchor at the ingest root, a trailing `/`
for directories only, and a bare name matching at any depth.

**Not** supported: `!` negation and character classes. A pattern starting with
`!` is dropped rather than half-honoured — both are answers to "I excluded too
much", and the honest fix for that is a narrower pattern. The doc comment says
so rather than letting the file pretend to be a `.gitignore`.

## No regex, on purpose

Patterns arrive from a file **inside the repository being ingested** — untrusted
input, on the one path in this CLI that parses arbitrary repositories.
`**a**a**a**b` compiled to a regex is the textbook catastrophic backtrack.

So it matches directly: the two-pointer wildcard walk (remember the last star,
resume one position further), at both the character and the segment level.
Worst case O(pattern × path), no recursion, no stack. Two tests drive
pathological patterns and assert the answer arrives in under 500 ms.

## Verification

- `npx vitest run` — 1,882 passed, 2 skipped · `npm run lint` ·
  all green
- new `ignore-globs.test.ts`, 14 cases: parsing (comments, negation dropped,
  overlong line, anchoring), matching (depth, anchor, `*` vs `**`, `?`,
  directory trees, a file pattern taking a tree, no-op, Windows separators) and
  the two backtracking cases
- new `ingest-exclude.test.ts`, 6 cases: the unfiltered walk, `--exclude`,
  `.ixignore`, one skip per excluded tree rather than per file, and
  `git ls-files` honouring it in a real `git init`ed fixture
- both extension lists untouched, so `extension-parity.test.ts` still holds

Part of the CLI output-economy work (task C-13, milestone M3).


Alot1z:
Runtime verification of this PR at its own head (`6118b7e`), hermetic backend 1.0.28, sandbox `IX_HOME`, fixture-heavy corpus checked into a real git repo so **both** discovery paths are exercised.

## What verifies

The feature works end to end, on both paths the doc comment promises:

| Case | Result |
|---|---|
| `git ls-files` path (committed repo) | `discovered: 2` — only `src/` files; `excluded: 4 paths` |
| fs-walk path (untracked file added under an excluded dir) | `discovered: 2`, `excluded: 5` — the new file is caught, not just the committed set |
| `--exclude` flags alone, no `.ixignore`, fresh `IX_HOME` | same exclusions; JSON summary carries `filesExcluded: 5` + full `skipReasons` |
| Graph contents | `fixtureCompute` (in an excluded fixture) → `No results found`; `compute` → 4 hits under `src/util/math.ts`; stats = 7 nodes / 2 files — exactly the non-excluded set |
| Pattern semantics | `fixtures/` (bare, unanchored), `/vendor/` (anchored), `docs/generated` (no trailing slash, excludes subtree), comment lines — all behave as the header documents |

The absence-in-graph proof matters: exclusion is not just a discovery-count trick, the excluded symbols never enter the searchable graph.

## One finding: a `!` negation line is a silent no-op at runtime

`parseIgnorePattern` drops `!`-prefixed lines deliberately (`ix-cli/src/cli/ignore-globs.ts:117-121`) — the header even explains why, and it's the right call. But at runtime the drop is invisible:

```
.ixignore:  !fixtures/          (user: "exclude everything except fixtures")
ingest:     discovered: 2, excluded: 5, exit 0, zero warnings
```

A user migrating lines from a `.gitignore` gets the **opposite** of what they wrote — over-broad exclusion instead of negated exclusion — with no signal anywhere: not in the text summary, not in the JSON, not on stderr. The same silent path handles character classes and the over-1024-char case (`parseIgnorePattern` returns null for all three, same call site).

The source comment is the only place this is acknowledged, and the source is not where a user looks when their graph is missing files they expected.

## Smallest fix, owned by this PR

`parseIgnorePattern` already knows when it refuses a line. Count the refusals at parse time and let the existing summary say so — the same honesty pattern this PR itself sets with `filesExcluded`:

```
  excluded:     5 paths (--exclude / .ixignore)
  unsupported:  1 pattern (! negation, char class, or >1024 chars)
```

~3 lines: a counter beside `patterns.push(...)` in `collectExcludePatterns`, a field in the summary object at `ingest.ts:3617`, one text row beside `ingest.ts:3667`. No behavior change, no parser change — it only makes a drop the parser already performs visible to the consumer, which is the same principle as `filesExcluded`: "an exclusion nobody is told about is indistinguishable from an ingest that missed something."

PR #700, opened 2026-09-19
perf(mcp): send the context bundle once, in the format its reader uses

KageBinary:
> Implements the `structured: true` half of **decision D7 as recommended, but
> unconfirmed**. Contract change for MCP clients — for a minor release, with the
> rest of the M3 batch. The `--tools=core` half is its own PR.

## The double delivery

`ix_context` sent every bundle **twice**: the JSON text, and the same object
again as `structuredContent`. A 20 KB bundle cost 40 KB.

Neither copy was the `llm` one a model actually reads — which is 46% smaller
and was not being sent at all.

Now the records go out once. `structured: true` opts back into the JSON object
for a caller that parses it, and gets both copies exactly as before.

## And the schema that came with it

An `outputSchema` is a promise that **every** result carries structured
content, so a tool whose structured copy is opt-in cannot declare one. Measured
against a live `tools/list`:

| | bytes |
|---|---|
| `ix_context` entry, today | 4,671 |
| `ix_context` entry, this branch | 1,034 |
| whole `tools/list`, today | 15,365 |
| whole `tools/list`, this branch | **11,694 (−24%)** |

That 4,671 was **30% of what every session pays to connect to this server**, for
a shape only a caller passing `structured: true` receives. And this is before
any of the toolset work (M-2).

## `ix_smells`

Same treatment, and the indent with it: the payload was serialized at two
spaces per level and then repeated as `structuredContent`.

## Verification

- `npx vitest run` — 1,863 passed, 2 skipped · `npm run lint` ·
- `mcp.test.ts`: the `ix_context` cases now assert `--format=llm` and one copy,
  with a new case for the `structured: true` path asking for `--format=json`
  and returning both, and one for a CLI error record arriving through the
  records path
- `mcp-structured-output.test.ts`: `ix_smells` asserted both ways, the
  `outputSchema` advertisement updated, and a new case that fails if
  `ix_context` ever costs more than 1,500 bytes of `tools/list` again
- byte counts above are from a live in-process `tools/list`

Part of the MCP work (task M-1, milestone M3).


Alot1z:
Runtime verification of this PR at its own head (`a33d68c`), speaking JSON-RPC to the real server over stdio (hermetic stack, memory-layer `1.0.28`, sandbox `IX_HOME`, corpus mapped by this build). Byte counts are UTF-8 lengths of the payload JSON, so they are directly comparable to the numbers in the description.

## The two headline numbers reproduce exactly

| | claimed | measured here |
|---|---|---|
| `ix_context` entry | 1,034 B | **1,034 B** (`description` 75 + `inputSchema` 708 + name) |
| whole `tools/list` payload | 11,694 B | **11,694 B** (raw frame 11,728 B with the JSON-RPC envelope) |

`ix_context` now carries no `outputSchema`, which is the right call by the argument in the description: a tool whose structured copy is opt-in cannot make the promise. `ix_smells` is 568 B.

## The delivery contract, both paths

```
ix_context {target:'probeToken1'}                     → text 584 B, llm records, structuredContent absent
                                                        (first line: context target=probeToken1 target_kind=function stale=false …)
ix_context {target:'probeToken1', structured:true}    → text 2,745 B JSON + structuredContent 2,096 B
                                                        (the text switches to --format=json as claimed)
ix_smells {}                                          → text 417 B, structuredContent absent, indented JSON: no
```

So: one copy, the llm one, for a reader that just reads; both copies, the JSON one, for a caller that parses. The indent removal on `ix_smells` is real too.

## One wrinkle in the failure path

With `structured: true`, a call that fails returns `isError: true` and the error as **text only** — no `structuredContent` — where the same flag on success returns both:

```
ix_context {target:'zzz-no-such-symbol-zzz'}                      → text: error code=unresolved_target message="No entity found matching \"zzz…\"."
ix_context {target:'zzz-no-such-symbol-zzz', structured:true}     → text: {"error":"No entity found matching \"zzz…\"","tool":"ix_context"}
                                                                    isError=true, structuredContent: absent
```

A caller that opted into `structured: true` because it parses, and reads `structuredContent` unconditionally, therefore gets `undefined` exactly when something went wrong — the case where a machine-readable error matters most. Nothing is broken: the payload is still machine-readable JSON in the text part, and the records path handles the non-structured case well (my default-path probe returned the same error as an `error code=…` record with `isError: true`, which is the case `mcp.test.ts` covers). But the flag's promise reads as "you get the parsed copy"; a sentence in the `structured` option's description, or emitting `structuredContent` on the error path too, would remove the one case where it isn't true.

## Two remaining empty schemas, for the record

`ix_map` and `ix_ingest` still declare `outputSchema`, both `properties: {}` (111 B each, 222 B of the 11,694 B catalog). They are the last instances of the shape this PR removed from `ix_context` — harmless, since an empty schema accepts anything, but they tell a client "structured output exists here" without saying anything about its shape. Their structured copy is unconditional, so this is not the same case as `ix_context`; noting it only so the remaining pair is a decision rather than an oversight.

PR #701, opened 2026-09-20
feat(mcp): make an ambiguous symbol resolvable, and bound what one tool returns

KageBinary:
> Covers **M-3 and M-4** together: they are the same change to the same tools —
> M-3 adds the parameters, M-4 adds the ones that are bounds. Splitting them
> would put two edits on each registration in two PRs.

## An ambiguous symbol was a dead end

Every symbol tool took a bare `symbol` and nothing else. A name with three
definitions answered "Ambiguous symbol" and **there was no second call that
could get past it** — the CLI's own `--pick`, `--kind` and `--path` were
unreachable over MCP, so the agent's only move was to shell out, which is the
thing this server exists to avoid.

`ix_callers`, `ix_callees`, `ix_imports`, `ix_imported_by`, `ix_explain` and
`ix_read` now take them:

```
["callers", "--kind=function", "--path=src/cli", "--pick=2", "--limit=20", "--format=llm", "--", "config"]
```

`--limit` goes to the five that have one. `ix explain` does not, and forwarding
a flag a command rejects turns a resolvable call into a usage error.

## `ix_search` did not exist

An agent over MCP could follow an edge from a symbol it already knew, and had no
way to find the first one. Now it can, with `kind`, `path`, `language` and a
default `limit` of 10.

## Two bounds

- **`ix_trace` gains `depth`, default 3.** `ix_depends` has had one since it was
  written; this had none, so a trace through a hub ran as deep as the graph goes.
- **A record stream is capped at 24 KB**, cut on a line boundary so no record is
  half-delivered, with a marker record:

  ```
  truncated tool=ix_text shown_bytes=24547 total_bytes=80000 hint="Narrow the call — every one of these tools takes a limit, a path or a depth."
  ```

  The only bound before this was the runner's 16 MiB heap guard, which exists to
  stop a long-lived server growing — not to stop one `ix text` filling a model's
  context. A host that truncates for you does it from the end with no marker, so
  the caller cannot tell a short answer from a chopped one.

**JSON results are deliberately not capped.** Cutting JSON produces something
that does not parse, which is worse than something large; their size is bounded
by the command's own budgets instead.

**`ix_read` takes no size bound either.** `ix read` owns its cap, and a second
number here would be one to keep in sync with it. A range goes in the target,
where the CLI already takes one: `read src/a.ts:40-80`.

## Verification

- `npx vitest run` — 1,872 passed, 2 skipped · `npm run lint` ·
  `npm run typecheck` · `npm run knip` — all green
- new `mcp-tool-params.test.ts`, 10 cases: the four flags forwarded, nothing
  sent for flags left out, `ix_explain` not being offered a `--limit`,
  `ix_read`'s flags, `ix_search` existing and forwarding its filters,
  `ix_trace`'s default depth, the cap on a line boundary with whole records
  before the marker, a fitting result untouched, and the cap firing through a
  real tool call
- the two catalog tests still hold: `ix_search` sits in
  `IX_MCP_OSS_TOOL_NAMES` in the position it is registered in

Part of the MCP work (tasks M-3 and M-4, milestone M3).


Alot1z:
Measured at runtime through a real MCP client (JSON-RPC over stdio: `initialize` → `notifications/initialized` → `tools/call ix_read`), on this PR's own build (`0.10.5`, hermetic ArangoDB on an alternate port, 2-file fixture repo). One unit mismatch in `capRecordStream`.

## The cap is named in bytes and measured in UTF-16 code units

`ix-cli/src/mcp/server.ts:729` and `:739-743`:

```ts
const MAX_TOOL_RESULT_BYTES = 24 * 1024;
...
if (text.length <= MAX_TOOL_RESULT_BYTES) return text;
const boundary = text.lastIndexOf("\n", MAX_TOOL_RESULT_BYTES);
const head = text.slice(0, boundary > 0 ? boundary : MAX_TOOL_RESULT_BYTES);
return `${head}\ntruncated tool=${tool} shown_bytes=${head.length} total_bytes=${text.length} hint="..."`;
```

`String.prototype.length` is UTF-16 code units. Two consequences, both measured:

**1. The cap leaks in real bytes.** Fixture: one TypeScript file, 3,000 lines, each carrying a Japanese comment (392,679 UTF-8 bytes / 278,679 code units), `ix read src/cjk.ts` through the MCP server:

```
marker:                              truncated tool=ix_read shown_bytes=24555 total_bytes=278806
ACTUAL utf8 bytes of returned head:  34891
cap constant MAX_TOOL_RESULT_BYTES:  24576
OVER THE CAP IN REAL BYTES:          true   (+42.6%)
```

Same call on an ASCII file (372,835 bytes) reports `shown_bytes=24493`, genuinely 24,493 bytes — so nothing in the current tests can see this.

**2. The marker misreports its own numbers.** Both fields are suffixed `_bytes` but are code-unit counts: `total_bytes=278806` against a 392,808-byte stream (the uncapped `ix read src/cjk.ts --format llm` reference, `wc -c`), i.e. understated by 29%. This is the number a caller uses to decide whether it can afford the full call, and it is the exact "truthful about what was omitted" contract the rest of this PR establishes.

The doc comment above the constant states the intent — how much a client "should have to hold", to stop one call "filling a model's context" — and a 1.4–3× overshoot is largest exactly for the non-ASCII sources the graph indexes on purpose (CJK comments/identifiers, em dashes). The best case here is 1.42×; all-CJK content approaches 3×.

## Smallest fix

`Buffer.byteLength` is already this codebase's idiom for byte accounting (`src/mcp/runner.ts:157`), so this is a two-line change inside one function, and it can stay byte-exact without splitting a character:

```ts
export function capRecordStream(text: string, tool: string): string {
  const buf = Buffer.from(text, "utf8");
  if (buf.length <= MAX_TOOL_RESULT_BYTES) return text;
  // 0x0A never appears inside a multi-byte UTF-8 sequence, so a byte-level
  // line search is safe and lands on a character boundary.
  const boundary = buf.lastIndexOf(0x0a, MAX_TOOL_RESULT_BYTES);
  const cut = boundary > 0 ? boundary : MAX_TOOL_RESULT_BYTES;
  const head = buf.subarray(0, cut);
  return `${head.toString("utf8")}\ntruncated tool=${tool} shown_bytes=${head.length} total_bytes=${buf.length} hint="..."`;
}
```

(If the intent is really code units, then the fix is the names — but then the constant's stated purpose, 24 KB of context, is no longer what it enforces.)

## Why the existing pins can't catch it

`src/cli/__tests__/mcp-tool-params.test.ts:110` pins the marker by shape (`/^truncated tool=ix_text shown_bytes=\d+ total_bytes=\d+ hint="/`), and the through-the-tool pin at `:120` builds its payload with `"node name=a kind=function path=src/a.ts\n".repeat(2000)` — pure ASCII, so `shown_bytes` and the true byte count agree and the pin passes either way. A payload with one non-ASCII character per record makes both assertions meaningful; asserting the *unit* rather than the shape is what closes this.

## Verified good on the same build

Independent of the above, this is confirmation rather than defect, from the same client session:

- **Bounds are real and fail closed.** Schema caps reject over-limit calls with the field named — `ix_text` `limit` > 100 → `-32602 … "Too big: expected number to be <=100 at limit"`; `ix_search` > 200 likewise. Defaults are advertised and applied: `search count=10 candidates=30`, `text limit=20` (`limit=10` / `20` in the schema), `trace depth=3`, `depends depth=2`, `rank top=10`, `inventory`/`smells limit=50`.
- **Catalog split holds.** Default `tools/list` = 10 tools, `--tools=all` = 25; `outputSchema` is declared only for the two write tools (`ix_map`, `ix_ingest`), and `ix_context`/`ix_smells` correctly declare none because their structured copy is opt-in.
- **The folded core tools are equivalent, not merely similar.** `ix_neighbors{relation:"callers"|"callees"|"imports"|"imported_by"}` in the default catalog is byte-identical to `ix_callers` / `ix_callees` / `ix_imports` / `ix_imported_by` in `--tools=all` on the same symbol (all four compared, `text === text`). The `--tools=all` fallback genuinely preserves what the core catalog drops.
- **Single-delivery context.** `ix_context{target}` returns one llm record stream (584 B) with no `structuredContent`; `ix_context{target, structured:true}` returns the JSON text plus a 2,096-byte `structuredContent` (`schema: "ix-context-bundle/1"`, budgets `maxEntities:100 maxEvidence:25 maxChars:12000`, and a zeroed `truncation` block) — no duplicate payload either way.

PR #702, opened 2026-09-20
feat(mcp): ten tools instead of twenty-six, and descriptions that say when to use them

KageBinary:
> Covers **M-2 and M-5**: both rewrite the same tool table, and splitting them
> would put two edits on every registration in two PRs.
>
> **Stacked on #701** (`feat/mcp-tool-params-and-bounds`), **and it merges
> #700** — the point of this PR is what `tools/list` costs a session, and #700
> takes 3.6 KB off it. Measuring without it would report a number nobody will
> ever see, since both land in the same minor release. Both merge cleanly to
> `main` on their own.

## The catalog

`tools/list` is paid on every session before a single call is made, and the
full catalog is 26 tools of which a recorded run used four. The cost is not only
bytes: a model choosing between 26 near-identical names picks worse than one
choosing between ten.

`--tools=core` is now the default:

```
ix_health  ix_locate  ix_search  ix_text  ix_impact
ix_overview  ix_read  ix_neighbors  ix_explain  ix_context
```

`--tools=all` advertises every tool, so nothing that works today stops working.
A tool outside the set is **not registered** rather than registered and hidden:
a surface a client cannot see but can still call is a surface with no
documentation.

`ix_callers`, `ix_callees`, `ix_imports` and `ix_imported_by` had byte-identical
schemas differing in one word. They are one `ix_neighbors{relation}` in the core
set, and all four stay under `--tools=all`.

**The Pro tools are exempt from the gate.** They are already conditional on Pro
being installed, and "Pro is installed, so its tools are offered" is not a
contract a toolset choice has any business silently reversing.

## Descriptions

Every one rewritten to ≤120 characters that says when to reach for the tool
instead of grep or Read, and that its rows carry places:

| tool | description |
|---|---|
| `ix_context` | Start here. Ranked evidence around a target, every row a path:lines, ending in reads to run |
| `ix_search` | Find a symbol by name. Ranked, each row carries path:lines. Use instead of grep for a definition |
| `ix_text` | Literal or regex search, code ranked above tests above prose. Use instead of grep for a string |
| `ix_read` | Source for a symbol, or path:start-end for a range. Prefer a range over a whole file |
| `ix_health` | Is the graph up to date. Call once at most; every other tool reports its own failure |

## The number, and where it lands

Measured in-process:

| | bytes | tools |
|---|---|---|
| today | 15,365 | 23 |
| `--tools=all` here | 15,269 | 25 |
| **`--tools=core`** | **6,437** | **10** |

**−58%.** The spec asked for ≤5 KB and **that is not reachable**: a tool with no
parameters at all is 412 bytes of name, description and annotations, so ten of
them are 4.1 KB before a single input is declared. Closing the last 1.4 KB would
mean giving back the disambiguation parameters from #701 that make an ambiguous
symbol resolvable at all — a worse trade than 1.4 KB is worth. Flagging it
rather than quietly missing the target.

## Verification

- `npx vitest run` — 107 files, 1,885 passed, 2 skipped · `npm run lint` ·
  all green
- new `mcp-toolset.test.ts`, 12 cases: the default being core, ten vs
  twenty-six, the size ceiling, `--tools=all` restoring the full catalog, Pro
  tools offered under both, an unadvertised tool not being callable,
  `ix_neighbors` replacing the four singles and routing each relation to its
  command (`imported_by` → `imported-by`), the flags carried through, every
  description ≤120 chars, and the `--tools` default
- the existing MCP tests now connect with `tools: "all"` where they assert
  against the whole catalog, which is what they were always about
- `flags.md` gains the `--tools` row and a note naming the ten

Part of the MCP work (tasks M-2 and M-5, milestone M3).


Alot1z:
Verified at this PR's own head (`ca6e9c3`) over a real MCP client, against a real backend — the catalog numbers reproduce, and I checked the one claim a listing can't show (a hidden tool must not be callable) plus CLI↔MCP agreement on the same operation.

**Setup:** CLI built from this head; hermetic `ix-memory-layer` 1.0.28 + `arangodb:3.12.11`; sandbox `IX_HOME`; `tools/list` measured from the raw newline-delimited JSON-RPC frames so the numbers are what a client pays.

## The catalog numbers

| | tools | `{"tools":[…]}` payload |
|---|---|---|
| default (`--tools=core`) | **10** | **6,447 B** |
| `--tools=all` | **25** | **15,279 B** |

Both land exactly **+10 bytes** off your table (6,437 / 15,269). The offset is identical in both rows, so it reads as framing rather than a disagreement about the catalog — worth a note only so the two figures are known to agree modulo counting. Core set is exactly the ten named in the body, and `--tools=all` still carries the four singles (`ix_callers`/`ix_callees`/`ix_imports`/`ix_imported_by`), so nothing that works today stops working.

**Descriptions:** 0 of 10 core and 0 of 25 total exceed 120 chars. No duplicate tool names.

## The claim a listing can't show

"**not registered** rather than registered and hidden: a surface a client cannot see but can still call is a surface with no documentation" — verified behaviorally, not by inspecting the list:

```
tools/call {"name":"ix_callers","arguments":{"symbol":"anything"}}
-> MCP error -32602: Tool ix_callers not found
```

Correctly a protocol-level not-found, not an empty result set — the distinction matters, since an agent can act on "no such tool" but will read an empty `callers` result as "there are none".

## Relation routing

`ix_neighbors` dispatches to the right underlying command, checked by the record the tool returns rather than by the schema:

| `relation` | emitted record |
|---|---|
| `callers` | `callers target=probeToken1 total=0 resolved=0` |
| `imported_by` | `imported-by target=probeToken1 total=0 resolved=0` |

## CLI ↔ MCP agreement (the same operation through both surfaces)

Same target, same corpus:

| | MCP | CLI |
|---|---|---|
| callers | `callers target=probeToken1 total=0 resolved=0 \| diagnostic code=no_edges message="No callers edges found."` | identical line, `--format llm` |
| context | `ix_context` → **584 B**, 5 lines | `context --format llm` → **585 B**, 5 lines |

The single-byte difference on `ix_context` is a trailing newline (CLI writes one, the MCP text part does not). After stripping one trailing newline from each, the two are **byte-identical** — so the surfaces agree on content, including the failure shape: MCP and CLI produce the same `no_edges` diagnostic rather than one degrading to prose. That is the part I'd most want preserved as this stack lands.

## Not verified here

The Pro-tools exemption ("Pro is installed, so its tools are offered"). No Pro install on this box, so I could not exercise `--tools=core` with Pro present — flagging it as unverified rather than assumed correct.

PR #703, opened 2026-09-20
feat(cli): let a caller ask for the answer without the scaffolding

KageBinary:
## What

```
$ ix search relativePath --format llm
search count=2 candidates=2
node name=relativePath kind=function id=8ebf63f9 path=ix-cli/src/cli/format.ts score=-101
node name=toMemberRelativePath kind=function id=a12e11f0 path=core-ingestion/src/patch-builder.ts score=-31
diagnostic code=unfiltered_search message="Results may be broad. Use --kind to filter and boost structural matches."

$ ix search relativePath --format llm --quiet --fields name,path
search count=2 candidates=2
node name=relativePath path=ix-cli/src/cli/format.ts
node name=toMemberRelativePath path=core-ingestion/src/patch-builder.ts
```

**343 bytes → 153**, a 55% cut, with both rows and both paths intact. Live run
against this repo's graph.

Both flags are declared once, for every command that has a `--format`, and
recorded by the root `preAction` hook — they are properties of a run rather than
of a payload, so no renderer signature grows two parameters it only forwards.

## Two lines drawn deliberately

**`--quiet` drops the scaffolding, not the bookkeeping.** Section titles,
`Resolved:` headers and advisory hints go. Warnings stay, error records stay,
and `shown=` / `total=` / `truncated=` stay.

A caller asking for less output is not asking to be misled about what was cut,
and a `--quiet` that hid `shown=50 total=212` would undo the one thing that
record exists for.

**`--fields` applies to rows, never to a header,** and keeps the **caller's**
order rather than the renderer's:

```
--fields path,name  →  node path=src/format.ts name=relativePath
```

`--fields path,name` is a request for a column layout; answering it in the
renderer's order ignores half of what was asked. A field no row carries is
dropped rather than emitted empty — `lines=` on a file entity would be the one
shape this cannot produce honestly.

## Overlap with the open stack

`format.ts`, `search.ts`, `inventory.ts` and `ui.ts` are also touched by #687,
#688, #690 and #696. The overlaps are textual (an added `projectRow(...)` wrap
and one import per file) and resolve in either merge order; this PR is based on
`main` so it can land independently of that stack.

## Verification

- `npx vitest run` — 106 files, 1,875 passed, 2 skipped · `npm run lint` ·
  all green
- new `quiet-and-fields.test.ts`, 13 cases: parsing, the caller's order, a
  missing field dropped, projection on search / node / edge / inventory rows,
  the header left alone, `--quiet` silencing titles and notes but **not**
  warnings, advisory diagnostics dropped from the record stream, and both flags
  present on every command that renders an answer and absent from ones that
  do not
- 66 new rows in `flags.md`, plus a note stating both lines above

Part of the CLI output-economy work (task C-12, milestone M3).


Alot1z:
Runtime verification of this PR at its own head (`f1b211b`), against a real backend — not the suites. Hermetic stack (`arangodb:3.12.11` + memory-layer `1.0.28`, isolated ports, sandbox `IX_HOME`); CLI = this branch's own build; corpus mapped with the build under test. Every string below is copied from this revision's output.

## The declared promises hold

| Promise (body) | Observed |
|---|---|
| `--quiet` drops the diagnostic, keeps the answer | `search probeToken1 --quiet --fields name,path` → header `search count=10 candidates=30` + rows, with the `hint:` block and the update banner gone |
| `--fields` follows the caller's order, not the row's | `--fields path,name` → `node path=src/big.ts name=probeToken1` (requested order, verbatim) |
| A field no row carries is dropped | `--fields name` on a mixed result set emits `name=` only where a row has one; a field absent from *all* rows is dropped, not emitted empty |
| `--quiet` never touches machine formats | `--format json --quiet` is byte-identical to `--format json` |
| A failure degrades to a record, not silence | `explain nosuchsymbol --format llm --quiet` → `error code=unresolved_target message="No entity found matching \"nosuchsymbol\"."`, exit 1 |

## One edge the body doesn't cover

When **every** requested field is absent — a typo, or a field valid elsewhere (`--fields bogus`, or `--fields label` against rows that carry no `label`) — the row does not degrade gracefully. It degrades to the bare kind token:

```
$ ix search probeToken1 --format llm --quiet --fields bogus
search count=10 candidates=30
node
node
node
... (11 lines: header + 10 rows)
$ echo $?   # 0
```

Ten lines of `node` with no field, no value, and exit 0 — indistinguishable from a result set whose projection legitimately produced nothing. The body's rule ("a field no row carries is dropped rather than emitted empty") describes the field correctly, but says nothing about the row that has no field left to render. A caller piping `--quiet --fields` into a script gets a confident, well-formed, useless answer.

Two low-cost options: reject a projection that matches nothing carried by any row in the result set (usage error, exit 2, naming the field), or keep the row's identity field (`name`/`path`) as a floor so a projected row is never contentless. Either way, "the projection matched nothing" should be sayable.

## Scope note on the truncation bookkeeping

The body also promises that truncation bookkeeping (`shown=` / `total=` / `truncated=`) survives `--quiet`. That promise is not exercisable on this PR's base: no command in this tree emits those records — `grep -rn "shown=\|total_bytes" src/` over `f1b211b` returns only the doctrine comments in `src/cli/output-shape.ts:27,56`. The marker is introduced by #688 (`sliceEdgeResults`); on that branch `callers hub --limit 2` returns `shown=2 total=30` plus a `results_truncated` record, and `--quiet` preserves both (verified there). So the invariant holds across the stack, but it is #688 that supplies the emitter — worth stating in this PR's description so a reviewer of #703 alone isn't looking for records that live one branch over.

